PR #1, opened 2015-09-08
Per template tuple generator code change (phase 3.1)

wofanli:
1) Support of random (I’ll add more distributions on phase 3.2)
 
2) Support Per template client/server generator.
aging/dual_port_mask/min_clients/clients_per_gb  can be taken from parent in default, but *can* be overridden.
For the new Yaml format, please refer to :
cap2/per_template_gen1.yaml
cap2/per_template_gen2.yaml
cap2/per_template_gen3.yaml
cap2/per_template_gen4.yaml
 
3) Dynamic calculation if to use the client-object with  64K bit port bit or the sparse one will be dynamic per the use of the template
 
4) Support two types of server pool
a) Stupid as today
b) With 64K bits ports for bundels – similar to client object (e.g. RTP) – need a knob for that .
Actually, I only implement it. But I have not used it in the plugin callbacks yet. Let’s wait for phase 3.2.
c) Have not done the check of verifying that there is no overlap with this pool addr


hhaim:
committed with some minor comments and small fix

PR #2, opened 2015-10-01
Tuple gne phase2 fixups

wofanli:
Code change according to code review comments after Tuple Generator phase 2 commit (trex-54).


wofanli:
Fixup of server number for Random distribution has not beed done yet.

wofanli:
@hhaim I added an api on generate_tuple to generate tuple by one lookup, which benefits performance.

PR #4, opened 2015-11-17
trex 54: new tuple-generator have a limitation with bundle traffic ( plugin template)

wofanli:


wofanli:
let me describe the issue:
When we allocate server for “bundle” flow (e.g. RTSP control) we should allocate the server port too (server start to talk to the client need to allocate port exacly like port allocate)
We need to make sure the port is distinct. 
The workaround today that each client talk always to the same server, and we use the client port (which is distinct) 

The  fix here is to maintain port usage of server allocators.

hhaim:
Yes. It is for all plugin that has a need for that..

Hanoh
Sent from my iPhone

On 17 ????? 2015, at 7:08, Wenxian Li <notifications@github.com<mailto:notifications@github.com>> wrote:


let me describe the issue:
When we allocate server for "bundle" flow (e.g. RTSP control) we should allocate the server port too (server start to talk to the client need to allocate port exacly like port allocate)
We need to make sure the port is distinct.
The workaround today that each client talk always to the same server, and we use the client port (which is distinct)

The fix here is to maintain port usage of server allocators.

-

src/tuple_gen.h
```diff
@@ -886,6 +928,16 @@ struct CTupleGenYamlInfo {
         }
         return 0;
     }
+    
+    void set_server_track_ports(uint8_t svr_pool_id, uint8_t plugin_id) {
+        if (plugin_id!=0) {
+            if (m_server_pool[svr_pool_id].m_is_bundling==false) {
```
wofanli: If user configures plugin_id and not configures track_ports, I'll force to enable track_ports. 
I'm thinking do we need such logic? Not every plugin needs to track server port, right? 

Or I can simply print out a warning without force to enable track_ports?

hhaim: Not all plugin need to track ports.
Maybe better to have an explicit YAML section.
And you should verify that in that case user choose server with tracking ports

Hanoh
Sent from my iPhone

On 17 ????? 2015, at 8:24, Wenxian Li <notifications@github.com<mailto:notifications@github.com>> wrote:


In src/tuple_gen.h<https://github.com/cisco-system-traffic-generator/trex-core/pull/4#discussion_r45026017>:

> @@ -886,6 +928,16 @@ struct CTupleGenYamlInfo {
>          }
>          return 0;
>      }
> +
> +    void set_server_track_ports(uint8_t svr_pool_id, uint8_t plugin_id) {
> +        if (plugin_id!=0) {
> +            if (m_server_pool[svr_pool_id].m_is_bundling==false) {


If user configures plugin_id and not configures track_ports, I'll force to enable track_ports.
I'm thinking do we need such logic? Not every plugin needs to track server port, right?

Or I can simply print out a warning without force to enable track_ports?

-

hhaim:
Hi Wenxian,
looks good a few comments, 

1. You should add a to the  YAML a new KEY 
     - name: avl/delay_10_rtp_250k_full.pcap
       cps : 1.982
       ipg : 10000
       rtt : 10000
       w   : 1
      server 
       server_pool: "s1"	# this should be tracked server pool 
       one_app_server : false
       track_server_port :true # should be false in default ( should come only in case you have plugin)
       plugin_id : 1

2. Could you  fix the  SFR profiles ( add tracked server pool) ? 
3. you should remove the restriction in the code for server 

bool CFlowsYamlInfo::verify_correctness(uint32_t num_threads) {


  /* patch defect trex-54 */
    if ( is_any_plugin_configured() ){
         /*Plugin is configured. in that case due to a limitation ( defect trex-54 )
          the number of servers should be bigger than number of clients   */

        int i;
        for (i=0; i<(int)m_vec.size(); i++) {
            CFlowYamlInfo * lp=&m_vec[i];
            if ( lp->m_plugin_id ){
                uint8_t c_idx = lp->m_client_pool_idx;
                uint8_t s_idx = lp->m_server_pool_idx;
                uint32_t total_clients = m_tuple_gen.m_client_pool[c_idx].getTotalIps();
                uint32_t total_servers = m_tuple_gen.m_server_pool[s_idx].getTotalIps();
                if ( total_servers < total_clients ){
                    printf(" Plugin is configured. in that case due to a limitation ( defect trex-54 ) \n");
                    printf(" the number of servers should be bigger than number of clients  \n");
                    printf(" client_pool_name : %s \n", lp->m_client_pool_name.c_str());
                    printf(" server_pool_name : %s \n", lp->m_server_pool_name.c_str());
                    return (false);
                }
            uint32_t mul = total_servers / total_clients;
            uint32_t new_server_num = mul * total_clients;
            if ( new_server_num != total_servers ) {
                printf(" Plugin is configured. in that case due to a limitation ( defect trex-54 ) \n");
                printf(" the number of servers should be exact multiplication of the number of clients  \n");
                printf(" client_pool_name : %s  clients %d \n", lp->m_client_pool_name.c_str(),total_clients);
                printf(" server_pool_name : %s  servers %d should be %d \n", lp->m_server_pool_name.c_str(),total_servers,new_server_num);
                return (false);
            }
          }
        }
    }

    return(true);
}

       ipg : 10000
       rtt : 10000
       w   : 1
       one_app_server : false
       track_server_port
       plugin_id : 1

wofanli:
@hhaim Updated the pull request. Could you help review it? thx.

PR #8, opened 2016-02-28
m_max_ports -> 12

RaminNietzsche:
change m_max_ports for 10 and 12 port. (Tested :D)


hhaim:
Thanks. fixed using your diff.

PR #11, opened 2016-04-09
Toggle-Filter support on trex tui

danklein10:
This request is for allowing user input to toggle port view on and off by hitting the port number when on trex tui display.

It uses the ToggleFilter general class (found at trex_control_plane/common) directory and tested using unit tests found at regression/functional_tests/filters_test.py

Functionality: when user hit the port number, the port display goes off if displayed, or on if not.
When more than one port shown, the 'total' column is also showed.

This change has been tested only using two-ports VM, so further lookup would be great.

Finally, we should consider how to merge it correctly with "Port" panel.

### DIFF
--- a/scripts/automation/trex_control_plane/stl/console/trex_tui.py
+++ b/scripts/automation/trex_control_plane/stl/console/trex_tui.py
@@ -13,6 +13,8 @@ else:
 from trex_stl_lib.utils.text_opts import *
 from trex_stl_lib.utils import text_tables
 from trex_stl_lib import trex_stl_stats
+import trex_root_path
+from common.filters import ToggleFilter
 
 # for STL exceptions
 from trex_stl_lib.api import *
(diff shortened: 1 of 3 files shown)


scripts/automation/trex_control_plane/stl/console/trex_tui.py
```diff
@@ -13,6 +13,8 @@
 from trex_stl_lib.utils.text_opts import *
 from trex_stl_lib.utils import text_tables
 from trex_stl_lib import trex_stl_stats
+import trex_root_path
```
imarom: We changed the model to a self contained model.
files from stl/* cannot point outside.

i will move your file filter.py to under trex_stl_lib.utils

imarom:
Dan,
filter does not work with Python 3 - pressing 0 removes all ports, pressing1 removes all >1 ports.

can you check it out ?

danklein10:
check out the latest commit ([3bafb03](https://github.com/cisco-system-traffic-generator/trex-core/pull/11/commits/3bafb0394c07ef2abb4ce34c7fb4ec01eb09f2df)), which should solve the python 3 issues.

imarom:
code was merged successfuly to master :+1:

PR #12, opened 2016-04-25
fix trex-204, add a sanity check for per template client/server yaml …

wofanli:
…configuration.


hhaim:
Hi Wenxian, 
looks good. 
Could you add a unit test ( with a wrong YAML)  to check this scenario?
I would convert the assert to exit as user can provide a wrong YAML

PR #13, opened 2016-04-25
Fix bug on mac replacement.

wofanli:
We only need to replace mac for client side. So add a check here.


hhaim:
looks good. I've integrated this.
Could you add a unit-test for this scenario?

PR #17, opened 2016-06-28
fix spelling of Ubuntu

rwhitworth:


imarom:
@rwhitworth 
Thanks.
I will merge the fix and it will be in the main branch in our next update.

Itay

PR #19, opened 2016-07-11
Added MAC learning in NAT mode

shmuelzon:
When NAT learning is enabled, TRex will also learn the source MAC address used
when sending packets to the server side. This MAC address will be used as the
destination MAC when sending packets from server to client. This will ignore the
destination MAC address defined in /etc/trex_cfg.yaml.


hhaim:
need to do it in a different way

PR #20, opened 2016-07-11
Send gratuitous ARPs

shmuelzon:
Since TRex can (will) not respond to ARP requests, it seems to be simpler to
just send gratuitous ARPs on all ends (clients, servers and GRE tunnels, when
used). This is done every 2 seconds.


hhaim:
already in v2.12

PR #22, opened 2016-07-15
add option to use DPDK tx checksum offloading to generate packets wit…

martin-weiser-allegro-packets:
…h correct IP, TCP and UDP checksums

This pull request is intended as a proposal resulting from an issue that we faced with trex.
Without this commit the packets that are generated always contain 0 for the UDP and TCP checksums. While this is acceptable for UDP which allows 0 to signal that no checksum has been calculated it is not allowed for TCP.
With DPDK many NICs can perform the checksum generation in hardware without creating much overhead in trex.

I probably did not choose the proper places to add this code but it is a large codebase and this is the best I could do so far. I'll be happy to improve this patch if you can point me in the right direction.

### DIFF
--- a/linux_dpdk/ws_main.py
+++ b/linux_dpdk/ws_main.py
@@ -492,6 +492,7 @@ includes_path =''' ../src/pal/linux_dpdk/
 ../src/dpdk22/lib/librte_mempool/
 ../src/dpdk22/lib/librte_pipeline/
 ../src/dpdk22/lib/librte_ring/
+../src/dpdk22/lib/librte_net/
 ../src/dpdk22/
               ''';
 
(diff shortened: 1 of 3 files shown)


hhaim:
Martin, 
looks good, thank you. will integrate that yo main branch in the office
There is a way to do this by software too without too much cycles beacuse we need only to calculate the diff. 
Are you using stateless mode?

hhaim:
actualy there is a need to add the update to another function to support plugins, will fix that

martin-weiser-allegro-packets:
Hi Hanoh,
thanks for the quick reply. Actually I have not used stateless mode so far.

PR #33, opened 2016-12-27
Let non-sudo users run scapy_daemon_server script.

hedjuo:
Remove sudo privilege restriction from scapy_daemon_server.


hhaim:
already fixed in a different way

PR #34, opened 2016-12-29
[STL_GUI-72] Add predefined packet templates

hedjuo:
Added possibility to add predefined templates to scapy server which will be available in the stateless GUI.

Original issue in YouTrack: [STL_GUI-72](https://trex-tgn.cisco.com/youtrack/issue/STL_GUI-72)


hhaim:
rebased

PR #49, opened 2017-06-30
Fix typos, bash formatting

HarrisonATT:
There were some minor typos that I found.
Also in the web version of the documentation, any word that starts with the '$' symbol is highlighted as a variable. Many of the code examples used the '$' as the PS1 prompt, which caused the command to be highlighted as if it was a variable. Adding spaces after all of the PS1 prompts should fix this problem.


hhaim:
thank you. it become a bigger commit 
see this 
https://github.com/cisco-system-traffic-generator/trex-core/commit/ca220418ecea64d3429039c4026db39064286b81

PR #57, opened 2017-09-05
Napatech updates - Documentation and regression test additions

mlilja01:
I have fixed two small bugs preventing the regression to run all tests. I have updated the documentation and added ntacc to the regression test suite


hhaim:
done

PR #58, opened 2017-09-12
Add support for encapsulating flows with L2 GRE

lxu4net:
It's been tested with OVS L2 GRE tunnel.

There are example configuration /etc/trex_cfg.yaml :
- port_limit: 2
  version: 2
  interfaces: ['06:00.0', '06:00.1']
  port_bandwidth_gb : 10 
  port_info:
      - dest_mac   : '6c:92:BF:27:40:02'
        src_mac    : '6C:92:BF:27:3F:E4'
        tunnels    :
            type   : nvgre
            dl_dst : 'FA:FF:FF:FF:FF:FF'
            dl_src : '6C:92:BF:27:3F:E4'
            tun_id : 100
            tun_dst: '192.168.1.10'
            tun_src: [ 192.168.1.20, 192.168.1.21 ]

      - dest_mac   : '6c:92:BF:27:40:02'
        src_mac    : '6C:92:BF:27:3F:E5'
        tunnels    :
            type   : 'nvgre'
            dl_dst : 'FA:FF:FF:FF:FF:FF'
            dl_src : '6C:92:BF:27:3F:E5'
            tun_id : 200
            tun_dst: '192.168.1.10'
            tun_src: [ 192.168.1.30, 192.168.1.31 ]


hhaim:
Thanks for the contribution 

A few comments/questions

1.  Shouldn't the Tunnel configuration be located in the traffic profile and not in the TRex hardware configuration file?

2. I would add a simulation test in our test to verify that the feature continues to work 
There are a few levels of unit test

2.1 to test the module itself
2.2 add system test 
Something like this
https://github.com/cisco-system-traffic-generator/trex-core/blob/master/src/bp_gtest.cpp#L434

3. Consider adding this support to the new TCP feature 

Hanoh


The GRE tunnel configuration should be set

lxu4net:
1. Shouldn't the Tunnel configuration be located in the traffic profile and not in the TRex hardware configuration file?

The Tunnel configuration just like the VLAN configuration. They are encapsulation format. 
And the tunnel can be seen as a virtual port, just like 'ip link' and OVS.

2. I would add a simulation test in our test to verify that the feature continues to work
There are a few levels of unit test

I will add some tests later.

3. Consider adding this support to the new TCP feature.

I will check it.

hhaim:
understood. 
I think the tunnel configuration is similar to the dot1q 

There are a two ways to add it

1. trex_cfg file (per port) 
2. cluster file - per pool of client, this way you can simulate a big network on GRE

I think you should support both options

hhaim:
We need to have a dynamic plugin mechanism to load tunnel support dynamically
I've traced this here 
https://trex-tgn.cisco.com/youtrack/issue/trex-482

once we will have it we would integrate this

PR #59, opened 2017-10-09
VLAN and ARP related bug fixes

driesdewinter:
While using trex on a network with VLANs, I came across two bugs:

- ARP replies from the router were ignored because of an incorrect interpretation of the VLAN header in CPretest::handle_rx(): the complete VLAN tag was matched against the VLAN ID of the port instead of only the VLAN ID in the packet, so it didn't work if the priority field was non-zero. Using the right utility method of VLANHeader to get the VLAN ID revealed that the endianness conversions in VLANHeader were not consistent so I cleaned that up too.

- Assertion failure when sending out a gratuitous ARP during a test, but the assertion was on a member variable that was never used elsewhere so I removed the variable.

### DIFF
--- a/src/common/Network/Packet/VLANHeader.h
+++ b/src/common/Network/Packet/VLANHeader.h
@@ -47,7 +47,7 @@ class VLANHeader
 ////////////////////////////////////////////////////////////////////////////////////////
 public:
 
-    // Sets the complete tag field without sub fields manipulation
+    // Sets the complete tag field without sub fields manipulation, host byte order
     void    setVlanTag          (uint16_t data);
     uint16_t  getVlanTag          ();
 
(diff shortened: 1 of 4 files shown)


hhaim:
Now I get it,
Did you remove the m->l2_len due to the assert?
If yes, it needs to be fixed in a different way ...
L2_len is used by the driver, need to set it in all the cases. This is the reason I’ve put the assert there

driesdewinter:
My previous fix for the assertion failure was indeed a mistake. I guess I followed my eclipse blindly too quickly: I never realized that there were multiple definitions of rte_mbuf and that is probably why eclipse only showed some of the usages. Anyway, I changed the fix now so that l2_len is initialized with a decent value (I think) for my particular case.

hhaim:
Done

PR #70, opened 2017-11-29
Add hltapi interface_config procedure support

sumboid:
For now, it's possible to configure port source and destination addresses, VLAN, and also user can force arp resolving process.
I didn't find hltapi commands for changing l2 parameters like destination mac (at least in ixia hltapi documentation I found related commands only for some *static_endpoint* mode, not sure TRex supports such mode).

If there is something that can be implemented without client changing please let me know, I'll extend implementation within this PR.

Output sample:
```python
{  
  'config':{  
    'layer_config':{  
      0:{  
        'ipv4':{  
          'dst':'1.2.3.4',
          'state':'resolved',
          'src':'4.3.2.1'
        },
        'ether':{  
          'dst':'08:00:27:23:21:dc',
          'state':'configured',
          'src':'08:00:27:d5:ac:98'
        }
      },
      1:{  
        'ipv4':{  
          'dst':'4.3.2.1',
          'state':'resolved',
          'src':'1.2.3.4'
        },
        'ether':{  
          'dst':'08:00:27:d5:ac:98',
          'state':'configured',
          'src':'08:00:27:23:21:dc'
        }
      }
    },
    'vlan_config':{  
      0:[  
        1,
        2
      ],
      1:[  
        1,
        2
      ]
    }
  },
  'status':1,
  'log':None
}
```

### DIFF
--- a/scripts/automation/trex_control_plane/stl/trex_stl_lib/trex_stl_hltapi.py
+++ b/scripts/automation/trex_control_plane/stl/trex_stl_lib/trex_stl_hltapi.py
@@ -28,6 +28,7 @@ interface_config_kwargs = {
 
     'gateway': [],
     'intf_ip_addr': [],
+    'dst_mac_addr': [],
 
     'vlan': [],
     'vlan_id': [],
@@ -225,6 +226,20 @@ def merge_kwargs(default_kwargs, user_kwargs):
             print("Warning: provided parameter '%s' is not supported" % key)
     return kwargs
 
+# change MAC from formats 01-23-45-67-89-10 or 0123.4567.8910 or {01 23 45 67 89 10} to Scapy format 01:23:45:67:89:10
+def correct_mac(mac):
+    if type(mac) is not str: raise STLError('Argument %s should be str' % mac)
+    updated_mac = mac.replace('{', '').replace('}', '').strip().replace('-', ' ').replace(':', ' ').replace('.',
+                                                                                                                ' ')
+    if updated_mac[4] == ' ' and updated_mac[9] == ' ':
+        updated_mac = ' '.join([updated_mac[0:2], updated_mac[2:7], updated_mac[7:12], updated_mac[12:14]])
+    updated_mac = ':'.join(updated_mac.split())
+    try:
+        mac2str(updated_mac)  # verify we are ok
+        return updated_mac
+    except:
+        return None
+
 # change MACs from formats 01-23-45-67-89-10 or 0123.4567.8910 or {01 23 45 67 89 10} to Scapy format 01:23:45:67:89:10
 def correct_macs(kwargs):
     list_of_mac_args = ['mac_src', 'mac_dst', 'mac_src2', 'mac_dst2']
@@ -234,15 +249,11 @@ def correct_macs(kwargs):
             mac_value = kwargs[mac_arg]
             if is_integer(mac_value) and mac_arg in list_of_mac_steps: # step can be number
                 continue
-            if type(mac_value) is not str: raise STLError('Argument %s should be str' % mac_arg)
-            mac_value = mac_value.replace('{', '').replace('}', '').strip().replace('-', ' ').replace(':', ' ').replace('.', ' ')
-            if mac_value[4] == ' ' and mac_value[9] == ' ':
-                mac_value = ' '.join([mac_value[0:2], mac_value[2:7], mac_value[7:12], mac_value[12:14]])
-            mac_value = ':'.join(mac_value.split())
-            try:
-                mac2str(mac_value)                                # verify we are ok
-                kwargs[mac_arg] = mac_value
-            except:
+
+            corrected_mac = correct_mac(mac_value)
+            if corrected_mac is not None:
+                kwargs[mac_arg] = corrected_mac
+            else:
                 raise STLError('Incorrect MAC %s=%s, please use 01:23:45:67:89:10 or 01-23-45-67-89-10 or 0123.4567.8910 or {01 23 45 67 89 10}' % (mac_arg, kwargs[mac_arg]))
 
 def is_true(input):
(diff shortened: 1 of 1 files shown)


scripts/automation/trex_control_plane/stl/trex_stl_lib/trex_stl_hltapi.py
```diff
@@ -424,15 +441,84 @@ def cleanup_session(self, **user_kwargs):
         self.trex_client = None
         return HLT_OK()
 
-    def interface_config(self, port_handle, mode='config'):
+    def interface_config(self, **user_kwargs):
         if not self.trex_client:
             return HLT_ERR('Connect first')
+
+        kwargs = merge_kwargs(interface_config_kwargs, user_kwargs)
+        is_array_based = not isinstance(kwargs['port_handle'], int)
```
ybrustin: Please reconsider this check, port handle can be for example string "1" or "1 2".
Look at function _parse_port_list

scripts/automation/trex_control_plane/stl/trex_stl_lib/trex_stl_hltapi.py
```diff
@@ -424,15 +441,84 @@ def cleanup_session(self, **user_kwargs):
...
         if mode not in ALLOWED_MODES:
             return HLT_ERR('Mode must be one of the following values: %s' % ALLOWED_MODES)
-        # pass this function for now...
-        return HLT_ERR('interface_config not implemented yet')
 
+        if is_array_based and not kwargs['port_handle']:
+            return HLT_ERR('Invalid port_handle parameter')
+
+        if not is_array_based:
+            kwargs['port_handle'] = [kwargs['port_handle']]
```
ybrustin: Please use _parse_port_list function to parse this argument

scripts/automation/trex_control_plane/stl/trex_stl_lib/trex_stl_hltapi.py
```diff
@@ -424,15 +441,84 @@ def cleanup_session(self, **user_kwargs):
...
+                if kwargs['intf_ip_addr'] or kwargs['gateway']:
+                    config = port.get_layer_cfg()
+                    src = (kwargs['intf_ip_addr'] and kwargs['intf_ip_addr'][i]) or config['ipv4']['src']
+                    dst = (kwargs['gateway'] and kwargs['gateway'][i]) or config['ipv4']['dst']
+
+                    port.set_l3_mode(src, dst)
+
+            if (is_true(kwargs['arp']) and is_true(kwargs['arp_send_req'])) or kwargs['intf_ip_addr'] or kwargs['gateway']:
+                for i, port_id in enumerate(kwargs['port_handle']):
+                    print(port_id)
```
ybrustin: Please remove the prints

scripts/automation/trex_control_plane/stl/trex_stl_lib/trex_stl_hltapi.py
```diff
@@ -424,15 +441,84 @@ def cleanup_session(self, **user_kwargs):
...
+            kwargs['vlan_id_list'] = kwargs['vlan_id_list'] and [kwargs['vlan_id_list']]
+            kwargs['vlan_id_inner'] = kwargs['vlan_id_inner'] and [kwargs['vlan_id_inner']]
+
+        try:
+            self.trex_client.set_service_mode(kwargs['port_handle'], True)
+        except Exception as e:
+            return HLT_ERR('Error enabling service mode: %s' % e)
+
+        try:
+            if mode == 'destroy':
```
ybrustin: Few notes regarding 'mode' variable.

1. AFAIK, mode 'destroy' should not implicitly continue to 'create' or 'modify'.
2. By spec, it should be list if port_handle is a list (with possible different values)
3. 'modify' and 'create' are same here

sumboid: @ybrustin, I have one fundamental question regarding "interfaces" in ixia, is there something corresponding that TRex has?

Here is the quote from ixia hltapi docs (mode parameter):
```
Action to be taken on the interface selected. This option takes a list of values
when -port_handle is a list of port handles. This option is valid for the old and
the new APIs. Valid options are:
   * config
   * destroy - clear all interfaces before configuring new ones.
   * modify - Valid only when using IxTclNetwork. When ::ixia::interface_config is
provided with -port_handle parameter, -mode modify and other supported parameters,
except -interface_handle, the modification is supported for L1 parameters only. When
::ixia::interface_config is provided with -port_handle parameter, -mode modify and
other supported parameters, including -interface_handle, the modification is supported
for L2-L3 parameters also, but only for protocol interfaces (-l23_config_type
protocol_interface). 
```
So, regarding list:
1. As I understand destroy means clearing all configs and create one from the scratch. But at TRex point, there is nothing to clear (or setting some default values), except vlan options.
2. Yep, my bad, I'll fix this.
3. After the first reading, I thought config and modify is almost the same for TRex, but for now, I'm not sure TRex can support `modify` at all. So maybe it would better to keep `config` and `destroy` and make it returns `HLT_ERR` on `modify` mode?

ybrustin: > As I understand destroy means clearing all configs and create one from the scratch. But at TRex point, there is nothing to clear (or setting some default values), except vlan options.

I hoped you did some research regarding what should be the behavior.
Let's ask (internally) some experts that might have answers.
Did you check that it works with existing HLT tests?

> After the first reading, I thought config and modify is almost the same for TRex, but for now, I'm not sure TRex can support modify at all. So maybe it would better to keep config and destroy and make it returns HLT_ERR on modify mode?

That was I would do, better do declare something is not supported, than declare it's supported and behavior will be incorrect...

sumboid:
@ybrustin,
I've fixed minor issues and also I keep only `config` mode just to not confuse users.
One more L2 parameter was added: `dst_mac_addr`. And I've removed auto arp resolution since this operation overwrites mac address implicitly.

hhaim:
not going to be committed

PR #71, opened 2017-12-06
General Exception was changed on more precise

0xck:
General Exception was changed on more precise built-in and trex related (from trex_exception.py) exceptions in order to make them more distinct. Because it's tricky to have deal with general Exception object and it is not common Python practice. I tried to find suited exceptions among built-in and defined in trex_exception.py for each place the Exception was used. I decided do not create the new in trex_exception.py because I know nothing about codes number there.


0xck:
Just in case. The `TimeoutError` is used in one place there it works only on Python 3, if there is some needs to save compatibility with Python 2 one can be changed.

hhaim:
thanks. rebased.

PR #80, opened 2018-02-01
More flexible support of Napatech NICs

mlilja01:
This patch enable a more flexible support of Napatech driver/firmware.


hhaim:
Hi Michael, 
it seems there are conflict between your old commit   9e28adb2974e05ef3cd52664b0d0db0fb46a393f and current commit that include the same.
Another issue is that "Napatech related documentation update" commit change *all* the doc "\n" due to your editor auto-replace. can you fix that?

mlilja01:
I'll fix it. Thanks


/Michael - Sent from my phone

PR #81, opened 2018-02-04
Accept drivers and firmware newer than 3.7 and 9.8

mlilja01:
Driver and firmware newer than 3.7/9.8 are automatically accepted. 

Signed-off-by: Michael Lilja <ml@napatech.com>


hhaim:
rebased

PR #84, opened 2018-02-19
Dockerfile using 16.04 LTS

michaelhans:
Adding a new Dockerfile based on Ubuntu 16.04 LTS, the included 16.10
is EOL and was throwing errors when building through docker
build (can't find Yakkety Yak base layer).

Confirmed this gets you a docker image capable of building trex master and simulator.


hhaim:
rebased

PR #90, opened 2018-03-29
Fix Python syntax in http_https.py

markabey:
Cryptic errors otherwise. This one is in the manual as the 2nd ASTF example.

### DIFF
--- a/scripts/astf/http_https.py
+++ b/scripts/astf/http_https.py
@@ -14,8 +14,10 @@ class Prof1():
                            dist_server=ip_gen_s)
 
         return ASTFProfile(default_ip_gen=ip_gen,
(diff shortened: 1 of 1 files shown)


ybrustin:
Thanks, merged.
I've also added prints of errors:
```
> sudo ./t-rex-64 -f astf/http_https.py --astf
converting astf profile astf/http_https.py to json /tmp/astf.json
./astf-sim -f astf/http_https.py  --json > /tmp/astf.json
ERROR could not convert astf profile to JSON try to debug it using the command above.
Produced output:
    Failed importing /auto/srg-sce-swinfra-usr/emb/users/ybrustin/trex/trex-core/scripts/astf/http_https.py
    invalid syntax (http_https.py, line 18)
ERROR encountered while configuring trex system
```

hhaim:
merged

PR #91, opened 2018-04-06
Update proper softlinks when '-o OUT' option used

jsmoon:
When I specified '-o OUT' option during configuration by ./b configure in linux/ and linux_dpdk/, built targets were located at given 'OUT' directory.
But I found that softlinks of the built targets were not generated properly in the scripts/. I investigated the reason and found that post_build routine in ws_main.py uses dedicated location to search built targets. It should refer to the specified 'OUT' directory from configuration.

This change will update the softlinks properly. So pkg and release can collect recent built targets also.

(cd linux && ./b configure -o ../../build/linux && ./b build)
(cd linux_dpdk && ./b configure -o ../../build/linux_dpdk && ./b build)
ls -l scripts/bp-sim*
ls -l scripts/_t-rex*

### DIFF
--- a/linux/ws_main.py
+++ b/linux/ws_main.py
@@ -640,10 +640,12 @@ def build_info(bld):
     pass;
       
 def install_single_system (bld, exec_p, build_obj):
-    o='build/linux/';
+    o=bld.out_dir+'/linux/';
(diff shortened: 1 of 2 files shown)


hhaim:
rebased

Issue #107, opened 2018-05-11
bpf filter for IPv6 is not enabled


egwakim:
Hi,
IPv6 BPF filter did not work when capturing packet.
I checked current TRex libbpf, "INET6" is not defined and it's not built by default.
I added it in build flag(-DINET6) and it worked.
I'm not sure why it's not enabled by default, is there any special reason for it?
If no special reason, shall I enable this build flag?

TRex/linux_dpdk/ws_main.py
...
291     # build the BPF as a shared library
1292     bld.shlib(features = 'c',
1293               includes = bpf_includes_path,
1294               cflags   = cflags + ['-DSLJIT_CONFIG_AUTO=1','-DINET6'],
1295               source   = bpf.file_list(top),
1296               target   = build_obj.get_bpf_target())
1297

hhaim:
Rebased

PR #110, opened 2018-05-28
BPF Filter : IPv6 BPF filter did not work before. Now it works with I…

JunsubHan4rmEricsson:
https://github.com/cisco-system-traffic-generator/trex-core/issues/107

Previously, IPv6 BPF filter did not work when capturing packet. 
This is because "INET6" was not defined in the TRex libbpf.
Now build flag, "-DINET6" has been added to solve this problem. 

Please check and provide feedback.


hhaim:
rebased.

PR #111, opened 2018-05-29
Add value_list in flow variable of vm instruction

yonghwan007:
Hi,
As I discussed with you, I merged my code again with your latest big commit.

**1. Summary**
The current flow variable in field engine only support 3 types: inc, dec and random with init_value, min_value, max_value.
Frequently user want to set a list values for specified field, like udp port [4001, 4005. 5003], the values are defined by use case without regular rules.
So, I have implemented 'value_list' of flow variable which controls array of values.

| value_list  | array of uint64_t as string | fixed list of values instead of init_value, min_value, max_value

**2. Example**
ipaddr_list = ["16.0.0.1","16.0.0.4","16.0.0.5","16.0.0.2","16.0.0.3","16.0.0.6"]

"vm": {
    "instructions": [
        {   
            "name": "src",                                                                                    
            "op": "dec",
            "size": 4,
            "step": 1,
            "type": "flow_var",
            "value_list": [
                268435457,
                268435460,
                268435461,
                268435458,
                268435459,
                268435462
            ]   
        }, 

**3. multi core test result**
List = [1,4,5,2,3,6]
Cores = 3
Output = 1,4,5,2,3,6,1,4,,,,,

BR,
Yonghwan

### DIFF
--- a/doc/trex_rpc_server_spec.asciidoc
+++ b/doc/trex_rpc_server_spec.asciidoc
@@ -797,6 +797,7 @@ Fix TCP/UDP and IPv4 headers using hardware assit engine
 | init_value  | uint64_t as string          | init value for the field
 | min_value   | uint64_t as string          | minimum value for the field
 | max_value   | uint64_t as string          | maximum value for the field
+| value_list  | array of uint64_t as string | fixed list of values instead of init_value, min_value, max_value
 | step        | uint64_t as string          | step, how much to inc or dec. 1 is the default (in case of 'random' this field is not used) 
 |=================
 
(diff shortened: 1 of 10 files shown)


hhaim:
Hi Yonghwan, 
it was merged. a minor things to consider: 

1. check that the number of element  is less than maximum (64K)
==> here in the Python
            self.value_list = convert_val_list(value_list)
            valid_fv_var_sizes(self.size, self.value_list)


and the same in the server check the same in the server. 

2. field engine.json should be updated, I've sent a question to the GUI guys

kisel:
Hi Yonghwan, Hanoch,

We will check and update field_engine.json accordingly.
I think it would be enough to make min/max values optional(remove required:true) and add a *value_list* field with an "EXPRESSION" type, so python expressions including arrays could be specified from a GUI

Thanks,
Anton

yonghwan007:
Okay, I will check it.

hhaim:
Anton, Will this work?
https://github.com/cisco-system-traffic-generator/trex-core/commit/81b4c1235f6127004f1bc1e85a9a7c257b8d3c91

PR #120, opened 2018-07-11
Add decompilation of vm raw to scapy server

EgorBlagov:
### Added API to decompile low level VM instructions to High Level
this method could be used to import native YAML profiles (generated from *.py files) into GUI. It takes low level vm instructions, parses it and returns it in Scapy Server field engine format
So, briefly, it turns this:
```
{
	"max_value": 12350,
	"min_value": 1235,
	"name": "ip_src",
	"op": "inc",
	"size": 4,
	"step": 1,
	"type": "flow_var"
},
{
	"pkt_offset": 14,
	"type": "fix_checksum_ipv4"
}
```
into this
```
{
  "id": "STLVmFlowVar",
  "parameters": {
    "op": "inc",
    "min_value": "1235",
    "size": "4",
    "name": "ip_src",
    "step": "1",
    "max_value": "12350"
  }
},
{
  "id": "STLVmFixIpv4",
  "parameters" : {
    "offset" : 26,
  }
}
```

### Added support of Expression in FieldEngine parameters
Since value_list is an expression (`"[1,2,3,4]"`), it's necessary to support expression format in FieldEngine parameters also (`{"vtype": "EXPRESSION", "expr": "[1,2,3,4]"}`)

### Fixed Scapy server imports
After Python API refactoring, it's seems to not be possible to start Scapy server as separate file (`python scapy_zmq_server.py`), with this fix Scapy server could be started as module `python -m trex.scapy_server.scapy_zmq_server` from 'interactive' directory, or as separate file  `python scapy_zmq_server.py` from 'scapy_server' directory, for both python2 and python3.

### Fixes and enhancements of Scapy server unit tests
Tests now pass either if `/etc/ethertypes` file exists or not (it was depend on EtherType human readable name (e.g. "LOOP" or "IPv4").
Tests could be started from `unit_tests` directory, from `scapy_server` directory and higher. As earlier tests are started by `nosetests` utility.
Fixed test of vm instructions (IP address is 4 bytes long, not 1)


hhaim:
rebased

PR #121, opened 2018-07-12
Scapy server imports fixes (repeat)

EgorBlagov:
There was an issue [this PR](https://github.com/cisco-system-traffic-generator/trex-core/pull/120) merge 

this commit https://github.com/cisco-system-traffic-generator/trex-core/commit/9e70667dc663ff4c5e3081c03a2fa2f834da7ece
overwrote this Scapy imports fix https://github.com/cisco-system-traffic-generator/trex-core/commit/1d08e750dbc967f7d1f278fd3c128a8f401a99c0

Since there is `scapy_service_dir` variable used in `_get_templates` method, it would be nice to merge this import fix again
Also this fix helps to start scapy_server by different ways, and it fixes `unittests` start

### DIFF
--- a/scripts/automation/trex_control_plane/interactive/trex/scapy_server/scapy_service.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/scapy_server/scapy_service.py
@@ -2,6 +2,11 @@
 import os
 import sys
 
+scapy_service_dir = os.path.abspath(os.path.dirname(__file__))
+trex_dir = os.path.join(scapy_service_dir, os.pardir, os.pardir)
+
+sys.path.append(scapy_service_dir)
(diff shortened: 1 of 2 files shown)


imarom:
Hi Egor,

We have decided against mixing paths inside source code even if it may allow you to run individual files with less typing.

Having the code executed in a package mode allows us better handling as \__init\__.py chain is being executed and we can rely upon it.

So we support only this mode:

**python -m trex.scapy_server.scapy_zmq_server**

EgorBlagov:
Okay, got it.

So I'll provide another fix to Scapy server, because ther variable scapy_server_dir, doesn't exist now and _get_templates returns empty array.

Thanks & regards,
Egor.

PR #128, opened 2018-07-18
DOC: clarify dummy ports single interface support matrix

mcallaghan-sandvine:
As per #125, we identified confusion and clarified whether or not single interfaces are viable for anything other than stateless. This change adds the necessary clarity to distinguish "dummy ports" in general vs. "single interface" support across each TRex mode and single vs. dual generator scenarios.

----
Signed-off-by: Matt Callaghan (mcallaghan@sandvine.com)


hhaim:
rebased

mcallaghan-sandvine:
reviewed - changes are delivered into main

PR #131, opened 2018-07-30
Fixed empty templates list issue

EgorBlagov:
Providing another fix, as I mentioned in https://github.com/cisco-system-traffic-generator/trex-core/pull/121

- Added scapy_service_dir property to ScapyService, because this path is used in several places (protocols.json, field_engine.json)
- Removed os.chdir, changed to relative paths
- Added test to check if get_templates returns same tempaltes list in several calls

All tests pass


hhaim:
rebased

PR #141, opened 2018-08-22
Add 'push_trex_config' and 'get_devices_info' commands for TRex daemon server

Rhymmor:
Added two commands for being able to change TRex configuration file (`/etc/trex_cfg.yaml`) without having access to the server, just with a TRex daemon running on it.

Command `get_devices_info` is to get information about network interfaces on the server for making TRex config with it.

Command `push_trex_config` is to push a new configuration file to the server.


ybrustin:
> Could you suggest something here? Maybe move this code to external_libs?

I would not add this function, do you really have a need in it?
A common case is that TRex is configured locally via "dpdk_setup_ports.py -i" or "dpdk_setup_ports.py -c" etc.
One could create several config files this way.

Thanks,
Yaroslav.

Rhymmor:
Thing is that in our case a user may not have access to configure it locally (just having an open port for TRex daemon) and/or want to do it from GUI. So a user has to know about NICs to configure it.

kisel:
Hi Yaroslav,

We've been asked by multiple teams to let them configure TRex through the GUI and avoid calling dpdk_setup_ports.py manually, that's the reason why do we want to let modifying *default* config. So basically GUI displays a list of interfaces and allows to pick them, as well to define additional parameters or modify raw yaml.
I think modifying default config is very useful along with such features as automatic start of TRex daemon itself - this will be used on servers running shared trex setups. Most users will benefit from this. I.e trex boots automatically with the default cfg. If new NICs added/replaced GUI can be used to reconfigure and reboot TRex.

As for having more than a one config - basically that sounds like allowing to list, write and read arbitrary files(not only configs, but maybe pcap files for ASTF modes) or passing a configuration straight to TRex 'start' API call as a b64 encoded yaml. This is also good and very likely will be needed in future, but was just out of scope of our current needs and full implementation will definitely raise more questions and security concerns.

To get info on NIC cards lshw looked like a best tool as it produces nice detailed xml suitable for GUI and doesn't require parsing, which could break in future, but we can use /proc fs or lspci for now to avoid that extra dependency and simply export less details to the GUI now.
However, If lshw is not installed trex will still work, just won't be able to display interface details.

You mentioned that you'might consider using lshw in future, so maybe we could start using it? Or lshw can also be compiled statically and embedded into trex, eliminating the dependency on lspci ( btw, regarding lspci we had few tickets from users who tried to run trex without lspci installed ) 

Thanks,
Anton

kisel:
Hi @ybrustin, @hhaim .
Anatoly has made changes to write configs to trex local dir. GUI for the daemon will get ability to list NIC devices and instead of writing  `/etc/trex_cfg.yaml` it will generate configs inside trex_dir.

Also there is a PR which provides GUI with the medatada in order to be able to edit TRex configuration(similar to packet editor way) - https://github.com/cisco-system-traffic-generator/trex-core/pull/151

hhaim:
looks good. I'm running regression to see that everything works.

hhaim:
rebased

PR #144, opened 2018-09-19
prevent zero pkts request from single_burst and multi_burst mode

jsmoon:
When single_burst or multi_burst mode used, total_pkts and pkts_per_burst can be set by zero.
This causes continuous sending pkts unexpectedly.
This change prevents zero value request from single_burst and multi_burst mode.

Following messages are part of rpc respond when zero value is used.

"error": {
    "code": -32000,
    "message": "Failed To Execute Method",
    "specific_err": "single_burst: 'total_pkts' cannot be zero."
},

"error": {
    "code": -32000,
    "message": "Failed To Execute Method",
    "specific_err": "multi_burst: 'pkts_per_burst' cannot be zero."
},


ybrustin:
Thanks, will be rebased to master.

PR #155, opened 2018-10-03
Replaced utf apostrophes with ascii

EgorBlagov:
- After testing on different VMs and Docker, found that in metadata's description present non ascii characters, this commit replaces it to ascii characters.


hhaim:
rebased

Issue #160, opened 2018-10-23
-k option is ignored


teknoraver:
TRex says:
```
-k  <num>                  : Run 'warm up' traffic for num seconds before starting the test.
```
But no warm up traffic is sent, no matter the value I specify

ybrustin:
-l (small L) option is required for this to have effect

PR #161, opened 2018-10-24
doc: clarify -k option behaviour

teknoraver:
warm up traffic (-k option) only works in conjunction with the
latency test (-l option), but the documentation doesn't say it.

Issue #160

Signed-off-by: Matteo Croce <mcroce@redhat.com>


hhaim:
rebased

PR #170, opened 2018-11-05
Added Napatech support for Intel PAC A10 GX

sohnt:
Added Napatech support for Intel PAC A10 GX


hhaim:
Could you replace this code with is_napatech()? 

```
  if ((0x8086 == self.m_devices[key]['Vendor'] and 0x09c4 == self.m_devices[key]['Device']) and \
            (os.popen("/opt/napatech3/bin/adapterinfo | grep -c " + key).readline().rstrip() == '1')):
                # These adapters doesn't need binding
                Napatech_cnt += 1
                continue 
```

PR #172, opened 2018-11-07
DPDK: change references of ixgbe to device

tbarbette:
I guess TREX started with ixgbe support. But some of the code path referring to ixgbe are actually for all devices.

That lead me to think some code path were not used. A cheap patch but I thought it would be useful for other people trying to poke around main_dpdk.cc


hhaim:
You are correct. the only driver was ixgbe in the first version

hhaim:
rebased.

PR #174, opened 2018-11-12
Fix dummy rx function used when AVX512F CPU feature available.

jsmoon:
This change will solve the ARP issue in #148.


hhaim:
rebased

hhaim:
The performance was reduced by 20% in STL mode. I've change the CL a bit to fix this

hhaim:
Could you test the latest code. I've change the code to fix a regression performance issue

jsmoon:
I've tested with latest code. ARP issue was gone and regarding performance
issue was not shown.
Actually I didn't met the performance issue you mentioned.
So, I cannot understand why your code can fix the performance issue.
Could you explain the performance issue in detail?

hhaim:
We have 24/7 regressions setup. the X710 setup with 16 ports shows 30% degradation in STL tests (in the ELK database).  I don't understand why it happen but changing the code as I did fix the issue.

PR #180, opened 2018-12-07
Fixed Scapy unreachable on python3

EgorBlagov:
- When Scapy server is started on python3, recv_string method of ZMQ
  socket returns byte sequence instead of decoded string, fixed this for both
  scapy server and scapy client implementation
- Edited Scapy client, to be able to work with python3
- Tested with python 2 and 3
- Replaced `send_string` by supported `send` method with `encode` call in scapy_zmq_client


hhaim:
rebased

PR #185, opened 2019-01-07
fix for vlan fmt option in services

andreasbourges:
Conflicts:
	scripts/automation/trex_control_plane/interactive/trex/common/services/trex_service_arp.py

Re-imported the changes for class VLAN() that got lost during api consolidation.


hhaim:
rebased

PR #189, opened 2019-02-01
TCP UDP checksum by SW.

jsmoon:
Hi,

I have implemented checksum support by SW where HW does not support TX checksum offload.
Since HW may not support TX checksum offloads (like as virtio in Cloud), checksum cannot be filled when STLVmFixChecksumHw specified in STL stream VM.

Although HW does not support TX checksum offloads, I think users expect to send packets with normal checksum when they set to update checksum.

And I found that there is a bug in DPDK code. I fixed it also as you can see.

Please review my Pull Request as soon as possible.


hhaim:
Thank you. It was asked here

hhaim:
reabsed with a few changes. please review

jsmoon:
I think your change is good for me also.

PR #199, opened 2019-03-06
dp_core supports per profile operation

jsmoon:
I've implemented code for per profile operation supporting in DP side.

Change summary:
- CP to DP messages should carry profile_id. (default value is 0 which means all, the same as legacy operation)
- TrexStreamsCompiledObj contains profile_id instead of TrexStatelessDpStart.
- added m_paused_streams to make TrexStatelessDpPerPort state reasonable during per profile operation. (one of streams is active, the state would be TrexStatelessDpPerPort::ppSTATE_TRANSMITTING)
- added compatible constructors which have 0 value profile_id to reduce code changes in CP side. (After CP side changes to use profile_id, it can be removed)

I've tested simple unit tests (bp-sim-64 --ut and run_regression --func) and passed with this changes.

### DIFF
--- a/src/bp_sim.cpp
+++ b/src/bp_sim.cpp
@@ -3065,6 +3065,7 @@ inline bool CNodeGenerator::handle_stl_node(CGenNode * node,
             /* count before handle - node might be destroyed */
             #ifdef TREX_SIM
             uint8_t port_id = node_sl->get_port_id();
+            uint32_t profile_id = node_sl->get_profile_id();
             update_stl_stats(node_sl);
             #endif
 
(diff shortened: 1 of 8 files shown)


hhaim:
One minor comment

m_active_nodes should be converted to map<profile-id,vector<streams>> 
Might be better to add it to an new object (PortPerProfile) moving the vector into it.

This way all the operations will iterate only on relevant streams (and not on all)

State is global for all streams, e.g. should be transmitted as long there is *one* stream in one of the profiles that is active

jsmoon:
I agree with that your suggestion would be better in performance.
But would you mind if I implement your suggestion after per streams operations are removed?
I think that your suggestion should consider them at this point and I am warring about it makes my code more complicated.

jsmoon:
My team made a decision to use one branch internally for stl dynamic profile feature.
So, I'll close this PR and this change will be merged to the internal branch.
Gwangmoon will make a notice about it to you.
I'm sorry for your inconvenience.

PR #208, opened 2019-03-20
Enable Scapy server for ASTF

Rhymmor:
Hi guys!

In our app, we use Scapy server for parsing/editing PCAP files from GUI. It would be really helpful if TRex could start in ASTF mode with Scapy server.

Thank you!

### DIFF
--- a/scripts/dpdk_setup_ports.py
+++ b/scripts/dpdk_setup_ports.py
@@ -710,7 +710,7 @@ Other network devices
 
 
     def run_scapy_server(self):
-        if not pa() or pa().no_scapy_server or not pa().interactive or pa().astf:
+        if not pa() or pa().no_scapy_server or not pa().interactive or (not pa().scapy_server and pa().astf):
             return
         try:
             master_core = self.m_cfg_dict[0]['platform']['master_thread_id']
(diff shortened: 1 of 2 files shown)


ybrustin:
Hi,

Is it going to be open source GUI like for stateless?

Thanks,
Yaroslav.

Rhymmor:
Hi @ybrustin,

We'd like to be open source, but it's not only our decision to make. It is still under discussion.

hhaim:
I suggest to have an explicit option to enable the Scapy server for ASTF -- why to pay the time to load it, if the GUI is not ready

Rhymmor:
@hhaim, I was going to add an option, but there is already `--no-scapy-server` flag now, should I add `--scapy-server` flag in this case?

Also, we can't publish GUI, since it won't be ready without this functionality.

hhaim:
```--scapy-server``` would be good

ybrustin:
Hi,

You could run the Scapy server independently from TRex, for development...

Thanks,
Yaroslav

Rhymmor:
Added `--scapy-server` flag

hhaim:
rebased

Issue #213, opened 2019-03-21
./b configure fails on RHEL 7.4 Regex does not match output


ballle98:
There is an issue introduced with a0d773a

the regex is incorrect for CentOS 7.4 because there is no equal sign in SEARCH_DIR like there is with CentOS 7.6

```
ld --version
GNU ld version 2.25.1-32.base.el7_4.2 

```

```
ld --verbose | grep 'SEARCH_DIR'
SEARCH_DIR("/usr/x86_64-redhat-linux/lib64"); SEARCH_DIR("/usr/lib64"); SEARCH_DIR("/usr/local/lib64"); SEARCH_DIR("/lib64"); SEARCH_DIR("/usr/x86_64-redhat-linux/lib"); SEARCH_DIR("/usr/local/lib"); SEARCH_DIR("/lib"); SEARCH_DIR("/usr/lib");

```

compared to 7.6 output:

```
ld --version
GNU ld version 2.27-34.base.el7

ld --verbose | grep SEARCH
SEARCH_DIR("=/usr/x86_64-redhat-linux/lib64"); SEARCH_DIR("=/usr/lib64"); SEARCH_DIR("=/usr/local/lib64"); SEARCH_DIR("=/lib64"); SEARCH_DIR("=/usr/x86_64-redhat-linux/lib"); SEARCH_DIR("=/usr/local/lib"); SEARCH_DIR("=/lib"); SEARCH_DIR("=/usr/lib");

```

PR #214, opened 2019-03-21
fixes cisco-system-traffic-generator/trex-core#213 configure fails on…

ballle98:
… RHEL 7.4 Regex does not match output


hhaim:
rebased

PR #198, opened 2019-02-28
[Client Implementation][New feature]Dynamic streams add/delete at run-time #181 

JunsubHan4rmEricsson:
https://github.com/cisco-system-traffic-generator/trex-core/issues/181

Hello. Cisco team
@Cyborn-Minjae-Lee @egwakim @ejangle @JaeyongYu @jsmoon @JunsubHan4rmEricsson @taewoonyun @yonghwan007 @youngjun1
I have implemented the client side of trex-core regarding issue #181 

Here is the description of the files modified

trex_types.py
	- STL_PORT_INFO class has been added to track logical port information "port_id"."profile_id" (ex, 0.logical_0, 0.logical_1, 1.logical_2)
	
parsing_opts.py
	- port parameter can be either int or STL_PORT_INFO instance

trex_client.py
	- _for_each_port has been modified to be used for each logical port instead of physical (ex, 0.logical_0, 0.logical_1, 1.logical_2)

trex_stl_client.py
	- Console APIs	**Ones that allow logical port functionality
		- push_line
		- start_line
		- stop_line
		- update_line
		- pause_line
		- resume_line
	- Client APIs	**Ones that allow logical port functionality
		- remove_all_streams
		- add_streams
		- start
		- stop
		- udpate
		- update_streams
		- pause
		- pause_streams
		- resume
		- resume_streams
		- push_remote
		- push_pcap
	
trex_stl_port.py
	- STL Port APIs	**Ones that allow logical port functionality
		- paused_streams
		- resume_streams
		- update_streams
		- add_streams
		- remove_all_streams
		- start
		- stop
		- pause
		- resume
		- update
		- push_remote
		- sync (get_port_status)
		- sync_streams (get_all_streams)
		
trex_stl_profile_manager.py
	- STLProfileManager
		- Used in STLPort to manage logical port state and streams

trex_psv.py
	- port validation of IDLE/TX/PAUSED has been modified for logical port as well
	
common.py
	**Commonly used functions in new imeplementations
	- parse_phyiscal_port_ids
		- Return physical ports ids from the list of any port list
	- parse_logical_port_ids
		- Return logical ports names in string from the list of any port list
	- parse_a_dual_port
		- Return a dual port from any port
	- parse_dual_ports
		- Return a list of dual port  from any port list

I'm still working on code refinement and testing.
Therefore, please feel free to provide feedback and let me know which part could be improved.

PR #211, opened 2019-03-21
Stl dynamic profile

Cyborn-Minjae-Lee:
Hi Cisco team

@Cyborn-Minjae-Lee @egwakim @ejangle @JaeyongYu @jsmoon @JunsubHan4rmEricsson @taewoonyun @yonghwan007 @youngjun1

Let's pull request this code.
It is related to an issue for #181 .
This only for the part of Server side including DP.
Server side was tested with the next Client.
If you want to consider for Client, please refer to https://github.com/bidtt4/trex-core/pull/4/files .
Client, Test and Documents will be pull-requested soon.

Please condsider and feel free to feed back.

BR,
Minjae Lee

PR #218, opened 2019-04-02
[Client Implementation] Dynamic streams add/delete at run-time #181 #198

JunsubHan4rmEricsson:
Hello.
@Cyborn-Minjae-Lee @egwakim @ejangle @JaeyongYu @jsmoon @JunsubHan4rmEricsson @taewoonyun @yonghwan007 @youngjun1

I have implemented the client side of trex-core regarding issue #181
Here are short descriptions of what changes have been made

**1) trex_client.py**
- _for_each_port has been modified to be used for each profile instead of port

**2) trex_psv.py**
- modified to validate for each profile instead of port

**3) trex_types.py**
- PortProfileID class has been added

**4) trex_stl_client.py**
- Console APIs and Client APIs modified to accept profile instead of port
- New STL command (profiles_line) has been added to provide profile information (let me know what you think)

**5) trex_stl_port.py**
- Modified to send JSON RPC for each profile instead of port

**6) common.py**
- Simple function (parse_ports_from_profiles) added to filter port information from profiles

**7) parsing_opts.py**
- PortProfileID are used for console APIs (start_line, stop_line ...)

Please take a look at it and let me your feedback

### DIFF
--- a/scripts/automation/trex_control_plane/interactive/trex/common/trex_client.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/common/trex_client.py
@@ -22,6 +22,7 @@ from .trex_conn import Connection
 from .trex_ns import NSCmds,NSCmd,NSCmdResult
 from .trex_logger import ScreenLogger
 from .trex_types import *
+from .trex_types import PortProfileID
 from .trex_exceptions import *
 from .trex_psv import *
 from .trex_vlan import VLAN
@@ -115,6 +116,9 @@ class TRexClient(object):
 
         # port state checker
         self.psv = PortStateValidator(self)
+
+        # server version check for dynamic port addition
+        self.is_dynamic = False
 
 
     def get_mode (self):
@@ -359,9 +363,8 @@ class TRexClient(object):
     def _transmit(self, method_name, params = None):
         return self.conn.rpc.transmit(method_name, params)
 
-
     # execute 'method' for a port list
-    def _for_each_port (self, method, port_id_list = None, *args, **kwargs):
+    def _for_each_port (self, method, port_list = None, *args, **kwargs):
 
         # specific port params
         pargs = {}
@@ -372,14 +375,21 @@ class TRexClient(object):
 
 
         # handle single port case
-        port_id_list = listify_if_int(port_id_list)
+        if port_list is not None:
+            port_list = listify(port_list)
 
         # none means all
-        port_id_list = port_id_list if port_id_list is not None else self.get_all_ports()
+        port_list = port_list if port_list is not None else self.get_all_ports()
         
         rc = RC()
 
-        for port_id in port_id_list:
+        for port in port_list:
+
+            port_id = int(port)
+            profile_id = None
+            if isinstance(port, PortProfileID):
+                profile_id = port.profile_id
+
             # get port specific
             pkwargs = pargs.get(port_id, {})
             if pkwargs:
@@ -387,8 +397,12 @@ class TRexClient(object):
             else:
                 pkwargs = kwargs
 
+            if profile_id:
+                pkwargs.update({'profile_id': profile_id})
+
             func = getattr(self.ports[port_id], method)
             rc.add(func(*args, **pkwargs))
+
 
         return rc
 
@@ -422,6 +436,9 @@ class TRexClient(object):
         rc = self._transmit("get_supported_cmds")
         if not rc:
             return rc
+
+        # server version check for dynamic port addition
+        self.is_dynamic = 'get_profile_list' in rc.data()
 
         self.supported_cmds = sorted(rc.data())
 
--- a/scripts/automation/trex_control_plane/interactive/trex/common/trex_psv.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/common/trex_psv.py
@@ -5,11 +5,10 @@ Author:
   Itay Marom 
 
 """
-
+from functools import wraps
 from .trex_exceptions import TRexError, TRexTypeError
-from .trex_types import listify_if_int
+from .trex_types import listify, PortProfileID
 from ..utils.common import *
-
 
 # type of states to be enforced
 # types starting with _ are internally used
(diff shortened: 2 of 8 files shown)


JunsubHan4rmEricsson:
Hi, I also have a feedback about the test_basic_multi_burst test mentioned in **Stl dynamic profile #211**
This takes place because profile state is not updated asynchronously from the server while port state is. 

In other words, port state in client does not match the overall profile state in client 
(because profile state in client does not get updated at all asynchronously)

Do you have any feedback about how to fix this problem?

hhaim:
Yes, I mentioned that the issue was with the events. Your fix is good and simple.  The client should only know the phy port status for the async profiles duration. 
 BTW with the alternative virtual port design (per profile) there is a need for this async notification  ( start a profile with duration -- there was a need to sync the state of the virtual port)
However, with the current API -- only phy port should know the global state. The per profile state could be pool.

Did you run regression with the client change?

hhaim:
It seems this commit break the regression again. 
Please run regression before PR 
Probably the events issue 
 
```
test_basic_cont (stateless_tests.stl_client_test.STLClient_Test) ... 
FAILED
test_basic_multi_burst (stateless_tests.stl_client_test.STLClient_Test) ... 
FAILED
test_basic_single_burst (stateless_tests.stl_client_test.STLClient_Test) ... 
FAILED
test_bpf (stateless_tests.stl_client_test.STLClient_Test) ... 
```

hhaim:
Playing with the new feature I found the following:

1. In general, it  works, which is very cool  
2. Let's look into this simple scenario from Console 

```

start -f stl/imix.py -m 10kpps --port 0            (total 10kpps)
start -f stl/imix.py -m 10kpps --port 0.imix1      (total 20kpps)
start -f stl/imix.py -m 10kpps --port 0.imix2      (total 30kpps)
start -f stl/imix.py -m 10kpps --port 0.imix3 -d 5 (total 40kpps fpr 5 sec and then back to 30kpps)

# wait 5 sec 
profiles
```

I would expect that profiles will show that imix3 is in stop (STREMAS) state (pooling) but it is stuck in TX
```
stop --port 0.imix3 
```
Just move it to STREAM mode (but is in STOP already) but does not **remove** it

There is a need to support explicit remove of a profile. I think that from the Console any none default profile should be remove when asked to stop

so this 
```stop --port 0.imix3```

will remove it 

In case of one profile the next start remove the profile and there was no problem with it but in this case the server need to check the status of of the profile and if it is not running to remove it

3. There is no way to remove all the old profiles and clean it

```stop --port 0.*```

just move all the profiles to state STREAMS but does not remove them. next start does not help too. we are stuck  

Again the suggestion is that  
```
stop --port 0.* 
```
will stop and **remove** all profile and make it clean 

stop --port 0 will just stop the old profile like it was. 

BTW only 'reset' command works 

4. there should be a way to look into one profile streams from Console 


```
streams --port 0.imix3
```

or 

```
streams --port 0 --id imix3
```

5. Another bug like 3 but with one stream 

```
reset
start -f stl/imix.py -m 10kpps --port 0.imix5 -d 5
#wait 5 sec
start -f stl/imix.py -m 10kpps --port 0.imix5 -d 5
  ==> show an error that it already started, but it really stopped 
  
```




I would try to fix those state bugs before looking into the regression

**Comment by hhaim** [comment]
Talking to the team here, the ``--remove`` should be explicit because the user might want to see statistics on the streams after the traffic was stopped.

**Comment by JunsubHan4rmEricsson** [comment]
Hi, I have read through all the issues you have mentioned. 
Please take a look at the following suggestions and provide feedbacks.

1) Profile state update
	[Issue.0] Regression test (test_basic_cont, test_basic_multi_burst, test_basic_single_burst)
	[Issue.2] Console examples with duration (Profile state stuck in TX)
	[Issue.5] Another bug like 3 but with one stream

	#Cause
		Profile state does not get udpated asynchronously while port state does.

	#Suggestion.1
		Implement "aynchroous profile state update" 
		For example, "_on_profile_stopped" and "async_event_profile_stopped"
		This is the cleanest solution from client perspective
	
	#Suggestion.2
		Everytime client API is used, the sync the profile state.
		This coudl be heavy in case many profiles exist

2) Explicit "Remove" of a profile
	[Issue.3] There is no way to remove all the old profiles and clean it
	
	#Suggestion
		Right now, "reset" command can be used for each phy port. 
		If you use "reset -p 0", this removes all the profiles located in the phy port. 
		Would you be okay with it?

3) "Streams" for each profile
	[Issue.4] There should be a way to look into one profile streams from Console

	#Suggestion
		I have created a "profiles" command that shows state and streams for each profile. 
		Could you please take a look at it?

**Comment by hhaim** [comment]
@JunsubHan4rmEricsson very nice summary.

1) **Profile state update**

**[Issue.0]**  It was fixed in the server side and regression only testing the **port** status, so I would expect the regression to pass. Do you understand why it fails with the new client code?
This is different than per profile status issue and more fundamental.

**[Issue.2-3]**  More elegant to solve this using async events mechanism (this was the main reason we developed it). In general, we try to make an old client compatible with a new server(e.g. Old client should not crash with a **new** event from profile stop)

 2) **Explicit "Remove" of a profile**

``reset`` is a big hammer and it is doing many other things. Dynamic  remove of **one** profile should be supported as it is the opposite to the function dynamic add. I think it is not a big deal to support it as it is  **supported** from API perspective, I'm just suggesting to add this utility to the Console using ```--remove``` and have the ability to remove the **default** profile too (before your feature only new start remove the old one implicitly)

**[Issue.4]**  Will look.

**Comment by hhaim** [comment]
Regarding this **[Issue.4]**  the new ``profiles`` command does **not** show the streams just meta data like the number of streams and state. There should be a way to show the full streams list (like command **streams** output)  but only for specific profile-id. 
One option is to add a  filter to ``streams`` command or to add a switch to ``profiles`` command

**Comment by Cyborn-Minjae-Lee** [comment]
Hi Hanoh,

Junsub is very busy for another job. So, I reply instead of him.

Regarding "Profile state update". Client is not updated for profile state by Async.
For example, start traffic with duration which is stopped by asyncronous stop event in sever side.
To solve this issue, I would like to suggest two solution.

The first is syncronizing the profile state between client and server using the synchronous command "state_profile" in command "get_port_status" to get profile state which was already implemented.

The second is to implement for asynchronous publish the event for profile state changing in Server side. And then, Client will know the profile state.

Please feed back your opinion?

BR,
Minjae Lee

**Comment by hhaim** [comment]
I’ve already answered this async event would be more elegant solution. This is how the port works

**Comment by Cyborn-Minjae-Lee** [comment]
Hi Hanoh,

I commented Async event for not only port but also Profile work, additionally.

Best Regards,
Minjae

**Comment by hhaim** [comment]
I understand your comment and question.
Again I suggest to send async event for profile change in *addition* to the port change event that is sent with current server (the change you did)

**Comment by JunsubHan4rmEricsson** [comment]
Hi Hanoh.

Here are the suggested solution for the request. 
Let me know if I understood it correctly

**1) Profile state update**
	: async event for profile state needs to be implemented both in server and client side (in addition to the exisitng async port state event)
**2) Explicit "Remove" of a profile**
	: --remove function for "stop command" needs to be supported (instead of recycling the reset command)
**3) "Streams" for each profile**
	: streams command need to show profile information as well

Thanks.

**Comment by hhaim** [comment]
Correct. 1 involve changing the server.
 2,3 are only Console utility to make this feature more usable

**Comment by JunsubHan4rmEricsson** [comment]
Right. I'll get to it ASAP
thanks for the quick feedback

**Comment by hhaim** [comment]
Hi,
Most of the events are relevant for readonly mode (START is not needed as the client started the profile itself), only STOPED_TX is relevant for the main client-- so better test it with readonly client too.

I would try to implement this type of test that work with events. 

Every random time add a profile and check for new events for end of TX to remove the old 
```

while (total_time < duration){
    
   add_profile(rand(name))
   time.sleep_ms(rand(100,300))
   event=get_event();
   if event=profile_finish();
      remove_profile(event.profile_name)
   
 }

```

JunsubHan4rmEricsson:
Alright sounds good. 
I will request a new pull-request with this code-refinement  
**in addition to the Server side implementation" which was pulled to the master branch a few days ago**

hhaim:
replaced with a new branch

PR #219, opened 2019-04-05
Add --unbind-unused-ports option to dpdk_setup_ports.py

ahothan:
This option if enabled will automatically unbind all ports that are not used
by TRex (i40e driver only).

Signed-off-by: ahothan <ahothan@cisco.com>


hhaim:
rebased

PR #225, opened 2019-04-25
doesn't work on python3.7 since async is a keyword

nickhuber:
```
  File "trex/common/trex_client.py", line 1188
    self.conn.async.set_timeout_sec(timeout_sec)
                  ^
SyntaxError: invalid syntax

```

The self.async and any uses for it need to be renamed to work with recent version of python


ybrustin:
Rebased, thanks!

PR #226, opened 2019-04-25
Astf set var 32 bit to 64 bit

ejaecki:
The current set_var() API of ASTF is not supporting the 64bit parameter.
it requires complex calculation from user side.

It means that if a user inputs the parameter value great than 2^32 into the set_var() API then the user will get the error message as below.

i.e. prog_s.set_var("var2",8x1024x1024x1024); # set var 0 to 2^33
------ error message ------------------------------
terminate called after throwing an instance of 'Json::LogicError'
what(): LargestUInt out of UInt range
./t-rex-64: line 80: 10805 Aborted (core dumped) ./_$(basename $0) $INPUT_ARGS $EXTRA_INPUT_ARGS

After patch : It'll run without errors
---------------------------------------------------

The reason why user wants to input the parameter value great than 2^32 is that he wants to test to download a xxGB file with a fixed xxMB buffer size.
So, I added the scripts/astf/ftp_multiple_flows_sim.py file to help your understanding.

### DIFF
--- a/scripts/astf/ftp_multiple_flows_sim.py
+++ b/scripts/astf/ftp_multiple_flows_sim.py
@@ -1,0 +1,86 @@
+# Example for multiple FTP flows simulation
+# Options :
+# fsize: file size in MB(Mega Bytes) to download by each flow
+# nflows : number of flows to download by FTP
+# tinc : time to increase flow, 0 to start all flows immediately, n to inclease number of flows by random(n) seconds
+#
+# Example of command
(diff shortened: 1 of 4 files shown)


ejaecki:
Including
@hhaim @Cyborn-Minjae-Lee @egwakim @ejangle @JaeyongYu @jsmoon @JunsubHan4rmEricsson@taewoonyun @yonghwan007 @youngjun1 @Kangphil

ejaecki:
Fixed typo.
i.e. prog_s.set_var("var2",810241024*1024); # set var 0 to 2^33
-->  prog_s.set_var("var2",8*1024*1024*1024); # set var 0 to 2^33

hhaim:
Nice! thanks for the ftp profile. rebased.

PR #228, opened 2019-04-29
enable FlowLatencyStats with value_list

Kangphil:
When using STLFlowLatencyStats with value_list, below error message is occur.

Attaching 1 streams to port(s) [0]: [FAILED]
start - Port 0 : *** You specified field engine with minimum packet size of 0 bytes. For latency together with random packet size, minimum packet size should be at least 76

It's caused because m_max_pkt_len and m_min_pkt_len in TrexStreamPktLenData class are not properly initialized. And, this commit is to fix this issue.


hhaim:
Thanks. looks good

hhaim:
rebased. thanks

PR #232, opened 2019-05-13
Fix mount /mnt/huge fail on read-only directory permission case

egwakim:
To fix the issue when some server have already have mounted hugetlbfs, but have read-only permission on /mnt/ directory. TRex will exit with exception in spite of available mounted hugetlbfs in that case because mkdir /mnt/huge will fail with exception.
This change will fix the issue on that case and I did not added exception handling intentionally to be able to exit TRex when real issue happening when creating directory on /mnt/huge.

### DIFF
--- a/scripts/dpdk_setup_ports.py
+++ b/scripts/dpdk_setup_ports.py
@@ -680,13 +680,12 @@ Other network devices
 
 
     def config_hugepages(self, wanted_count = None):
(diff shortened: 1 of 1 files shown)


egwakim:
Hi,
Can you review it for merge?

hhaim:
rebased. thanks

PR #233, opened 2019-05-15
replace tab to space for regression test

egwakim:
Rework for regression test : replace tab to space


hhaim:
Was fixed already. By mistake was committed again.

PR #246, opened 2019-05-23
ASTF Dynamic Multiple Profile in CP side

yonghwan007:
I added CP code for ASTF dynamic multiple profile with DP code which already pull requested.
Major changes is as below.

1.  Classes
- class TrexAstfPerProfile -> Handling state, buffer, hash, etc per each profile
- class TrexAstfProfile -> Management for related profiles
- class TrexAstf -> Others except profiles

2. Handling commands requested from Client
- Added m_cmd_wait_list to handle mass start or stop commands at the same time.
- This waiting list is used to avoid DP overload due to mass messages between CP and DP.

3. New event through asynchronous port.
- EVENT_ASTF_PROFILE_STATE_CHG = 60, -> event for profile state change
- EVENT_ASTF_PROFILE_CLEARED   = 61,  -> event for profile cleared (profile DB remove)

4. Profile clear for remove database
- Added TrexAstfDeleteDB message between CP and DP to remove .
- It will delete CAstfDB in DP and TrexAstfPerProfile instance in CP by profile_clear RPC command.

### DIFF
--- a/scripts/automation/regression/data/astf_dns2.json
+++ b/scripts/automation/regression/data/astf_dns2.json
@@ -1,0 +1,76 @@
+{
+    "templates": [
+        {
+            "client_template": {
+                "program_index": 0,
+                "cluster": {},
+                "cps": 1,
+                "port": 54,
+                "ip_gen": {
+                    "dist_client": {
+                        "index": 0
+                    },
+                    "dist_server": {
+                        "index": 1
+                    }
+                }
+            },
+            "server_template": {
+                "program_index": 1,
+                "assoc": [
+                    {
+                        "port": 54
+                    }
+                ]
+            }
+        }
+    ],
+    "ip_gen_dist_list": [
+        {
+            "distribution": "seq",
+            "ip_end": "32.0.0.255",
+            "ip_start": "32.0.0.0",
+            "dir": "c",
+            "ip_offset": "1.0.0.0"
+        },
+        {
+            "distribution": "seq",
+            "ip_end": "64.0.255.255",
+            "ip_start": "64.0.0.0",
+            "dir": "s",
+            "ip_offset": "1.0.0.0"
+        }
+    ],
+    "program_list": [
+        {
+            "commands": [
+                {
+                    "name": "tx_msg",
+                    "buf_index": 0
+                },
+                {
+                    "name": "rx_msg",
+                    "min_pkts": 1
+                }
+            ],
+            "stream": false
+        },
+        {
+            "commands": [
+                {
+                    "name": "rx_msg",
+                    "min_pkts": 1
+                },
+                {
+                    "name": "tx_msg",
+                    "buf_index": 1
+                }
+            ],
+            "stream": false
+        }
+    ],
+    "buf_list": [
+        "ACoAAAABAAAAAAAAA3d3dwVjaXNjbwNjb20AAAEAAQ==",
+        "ACqEgAABAAEAAAAAA3d3dwVjaXNjbwNjb20AAAEAAcAMAAEAAQAAAAAABGRkZGQ="
+    ]
+}
--- a/src/44bsd/flow_table.cpp
+++ b/src/44bsd/flow_table.cpp
@@ -224,21 +224,34 @@ void CFlowTable::parse_packet(struct rte_mbuf * mbuf,
 static void on_flow_free_cb(void *userdata,void  *obh){
     CFlowTable * ft = (CFlowTable *)userdata;
     CFlowBase * flow = CFlowBase::cast_from_hash_obj((flow_hash_ent_t *)obh);
-    ft->terminate_flow(flow->m_ctx,flow);
+    ft->terminate_flow(flow->m_ctx->m_tcp_ctx, flow, false);
 }
 
 void CFlowTable::terminate_flow(CTcpPerThreadCtx * ctx,
-                                CFlowBase * flow){
+                                CFlowBase * flow,
+                                bool remove_from_ft){
     uint16_t tg_id = flow->m_tg_id;
     if ( !flow->is_udp() ){
-        INC_STAT(ctx, tg_id, tcps_testdrops);
-        INC_STAT(ctx, tg_id, tcps_closed); 
+        INC_STAT(flow->m_ctx, tg_id, tcps_testdrops);
+        INC_STAT(flow->m_ctx, tg_id, tcps_closed);
     }
-    handle_close(ctx,flow,false);
+    handle_close(ctx, flow, remove_from_ft);
 }
 
 void CFlowTable::terminate_all_flows(){
     m_ft.detach_all(this,on_flow_free_cb);
+}
+
+static bool flow_predicate_cb(void *ctx, void *obh) {
+    CFlowBase * flow = CFlowBase::cast_from_hash_obj((flow_hash_ent_t *)obh);
+    return (flow->m_ctx == (CPerProfileCtx*)ctx);
+}
+
+void CFlowTable::terminate_profile_flows(CPerProfileCtx* ctx) {
+    if (ctx->m_flow_cnt > 0) {
+        m_ft.detach_if(this, on_flow_free_cb, ctx, flow_predicate_cb);
+        assert(ctx->m_flow_cnt == 0);
+    }
 }
 
 void CFlowTable::handle_close(CTcpPerThreadCtx * ctx,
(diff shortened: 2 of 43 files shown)


hhaim:
I reviewed  ```trex_astf.cpp```, the idea is good but could be better 


1.	PAUSE state

I had a discussion with Yaroslav about the CP and he suggested a partial workaround for the long build/parse in DP. 
He suggested to add a state (CP) STATE_PAUSE between STATE_BUILD -> STATE_TX. 
Today start will do PARSE/BUILD/TX (slow). Instead, we could load all profile to a PAUSE state and then without delay start/stop any profile (without remove)
Another change from current design is the implicit remove that should be done only for default profile.
from the Console we should have pause/resume new command to support that. This is not related to the PR 

2. In high level there is a need to explictly remove a new profile (--remove) in stop. 


3.	

```
void TrexAstf::build(string profile_id) {
..
    if (is_another_profile_busy(profile_id)) {
        add_cmd_in_wait_list(profile_id, CMD_BUILD);
        return;
    } else {
        del_cmd_in_wait_list(profile_id);
    }

```

Why do we need to wait list? the event from DP PARSE/BUILD have profile_id inside the message from DP->CP.

The Python API have 

```
load  : STATE_LOADED
start :   STATE_PARSE->STATE_BUILD->STATE_TX

so if I do this 

load(pid=6)
load(pid=7)
start(pid=7)
start(pid=6)

```
it should work. parse will send event to DP and event back from DP will progress the state 


4. add this ```typedef std::string cp_profile_id_t``


5. ```TrexAstfPerProfile`` should not be public class. put the vars in private senction 

6.  most of the function with cp_profile_id_t arg should be moved to ```TrexAstfPerProfile``

   
```
class TrexAstfProfile : public AstfProfileState {


    uint32_t            get_profile_index_by_id(std::string profile_id);
    state_e             get_profile_state_by_id(std::string profile_id);
    std::string         get_profile_id_by_index(uint32_t profile_index);
    CSTTCp*             get_sttcp_by_id(std::string profile_id);

    trex_astf_hash_e profile_cmp_hash(std::string profile_id, const std::string &hash);
    void profile_init(std::string profile_id);
    void profile_append(std::string profile_id, const std::string &fragment);
    void profile_set_loaded(std::string profile_id);

    /**
     * Handle profile status
     */
    void profile_change_state(std::string profile_id, state_e new_state);
    void profile_check_whitelist_states(std::string profile_id, const states_t &whitelist);
    bool is_profile_state_build(std::string profile_id);

    bool profile_needs_parsing(std::string profile_id);
    bool is_error(std::string profile_id) {
        return get_profile_by_id(profile_id)->m_error.size() > 0;
    }
```
let's take this snippets 

```
void TrexAstf::profile_clear(string profile_id){
    profile_check_whitelist_states(profile_id, {STATE_IDLE, STATE_LOADED});
    if (profile_id == DEFAULT_ASTF_PROFILE_ID) {
        profile_init(profile_id);
        return;
    }

    profile_change_state(profile_id, STATE_DELETE);
```

will be converted to, no lookup for each function simpler

```
void TrexAstf::profile_clear(cp_profile_id_t cp_profile_id){
    TrexAstfPerProfile * pid= get_profile(cp_profile_id);
    profile_check_whitelist_states(pid, {STATE_IDLE, STATE_LOADED});
    if (profile_id == DEFAULT_ASTF_PROFILE_ID) {
        pid->profile_init();
        return;
    }

    pid->profile_change_state(STATE_DELETE);
    ..
```


7.  ```void cleanup(std::string profile_id);```

should be only for default profile (backward compatible and regression reasons) 


8. Nice!  there is no meaning for global state anymore. however, this is good for backward compatibility and it will keep one profile the same

```
void TrexAstf::update_astf_state() {
    int temp_state = 0;
    vector<state_e> states = get_profile_state_list();

    for (auto it : states) {
        temp_state |= (0x01 << it);
    }

    if (temp_state & (0x01 << STATE_TX)) {
        change_state(STATE_TX);
    } else if (temp_state & (0x01 << STATE_BUILD)) {
        change_state(STATE_BUILD);
    } else if (temp_state & (0x01 << STATE_PARSE)) {
        change_state(STATE_PARSE);
    } else if (temp_state & (0x01 << STATE_CLEANUP)) {
        change_state(STATE_CLEANUP);
    } else if (temp_state & (0x01 << STATE_LOADED)) {
        change_state(STATE_LOADED);
    } else if (temp_state & (0x01 << STATE_IDLE)) {
        change_state(STATE_IDLE);
    }
}
```


9. This functions could be moved to be per profile object ``` TrexAstfPerProfile ``` with back pointer 

```
void TrexAstf::parse(string profile_id) {
    ..

    send_message_to_dp(0, msg); // give back pointer to TrexAstf for sending he messages 
}

```

build(string profile_id);
transmit(string profile_id)
..


10 . the same for this function. move this logic to TrexAstfPerProfile

```
void TrexAstf::dp_core_finished(int thread_id, uint32_t profile_index) {
    TrexAstfPerProfile* mprof = get_profile_by_id(get_profile_id_by_index(profile_index));

    mprof->m_active_cores--;
    if ( mprof->m_active_cores == 0 ) {
        all_dp_cores_finished(profile_index);
    } else {
        assert(mprof->m_active_cores>0);
    }
}

```


and this too 
```
void TrexAstf::all_dp_cores_finished(uint32_t profile_index) {
```

in short try to move  TrexAstf  to TrexAstfPerProfile and provide a back pointer to TrexAstf

yonghwan007:
First I will update code as your opinion, but I need to get more information for 1, 3, 7 review.

1. PAUSE state
You suggested to add a state (CP) STATE_PAUSE between STATE_BUILD -> STATE_TX.
For this state, I understood current start command will work PARSE/BUILD/TX but new command will work PARSE/BUILD/PAUSE before loaded all profiles, and then start command will do TX. Is it correct?
Could you explain more detailed state machine?
Also, do you mean that  new STATE_PAUSE and new pause/resume commands are not related to this PR?

3. wait list
I think this idea comes from the similar long parse/build issues of 1.PAUSE state as you mentioned above .
If we start many profiles at the same time in console, it may cause DP overload.
In this reason, I serialized commands in waiting list, so it could solve many issues.

For example, we can execute many commands at the same time in console.
start -f astf/http_simple.py --pid h0
start -f astf/template_groups.py --pid t0
start -f astf/udp_mix.py --pid u0
start -f astf/http_simple2.py --pid h1
stop --pid *

7. void cleanup(std::string profile_id);
Now I have no issue for backward compatible and regression test.
Could you explain more detail?

hhaim:
1. I didn't ask you to impl it, is just an idea to keep in mind for the next phase.  I think we should first commit with "regression" works when the engines are ready for multi-profile and then add features. 
3. No need for a wait list. The messages will be in the queue (CP->DP) and each CP profile will handle its (DP->CP) events. 
Just remove it and try 

```
start -f astf/http_simple.py --pid h0
start -f astf/template_groups.py --pid t0

```
If you run this from the console, you won't get to an issue because each command is blocking for all the states ```(  STATE_LOADED,
        STATE_PARSE, // by DP core 0, not cancelable
        STATE_BUILD, // by DP cores, not cancelable
        STATE_TX,)```
. To get thing more interesting you will need to split the API to do specifically state.


Try to refactor the code and I will review it again. 
I think for (3)  you will need in the next phase a test to verity it

yonghwan007:
Thanks for your comment.
I will try to refactor my code first, then let's discuss it one by one.

hhaim:
> Thanks for your comment.
> I will try to refactor my code first, then let's discuss it one by one.

Yep, try to move all the logic to the TrexAstfPerProfile object and give it back pointer to the messaging 
In the DP it is the same. all the logic  is in per Profile object.

yonghwan007:
I finished refactoring as your opinion.

hhaim:
The code structure is much better now .
DP code to be committed first.

yonghwan007:
I fixed the bug of wrong back pointer of TrexAstf for default profile, and committed my fix again.
But Travis is not running. Can you check it?

hhaim:
You need to add the keyword to enable it.  see here [wiki](https://github.com/cisco-system-traffic-generator/trex-core/wiki/Travis-CI)

yonghwan007:
I added new RPC command to support the TUI counter windows in client side. It merges statistics of each profile.
`TREX_RPC_CMD(TrexRpcCmdAstfTotalCountersValues, "get_total_counter_values");`

I will do regression test with travis after DP code passed.

hhaim:
@yonghwan007 
Could you verify this PR with latest code?

yonghwan007:
I have done regression test with new PR(#283) including latest code.
I will close this PR after your confirmation.

hhaim:
Sure, you can close this PR.
We are building the lab in the DMZ, please run full regression on your setups for now

PR #224, opened 2019-04-23
STL profile regression tests refinement

JunsubHan4rmEricsson:
Hi 
@hhaim @Cyborn-Minjae-Lee @egwakim @ejangle @JaeyongYu @jsmoon @JunsubHan4rmEricsson @taewoonyun @yonghwan007 @youngjun1

A few problems occured with #220 regression test. 
I think we should use this thread for STL profile regression test  refinement. 

For the first commit, I've added "test_pause_resume_update_dynamic_profile" which basically tests pause/resume/update with 100 profiles. 

It runs well on our system but please take a look at it and run it on yours as well.
Let me know how it goes.

PR #262, opened 2019-05-27
From STL_Profile_Regression_Test_Refinement for travis regressiont test

JunsubHan4rmEricsson:
Merged from #224 for travis regression test

### DIFF
--- a/scripts/automation/regression/stateless_tests/stl_client_test.py
+++ b/scripts/automation/regression/stateless_tests/stl_client_test.py
@@ -827,6 +827,7 @@ class STLClient_Test(CStlGeneral_Test):
             num_profiles = 100
             tx_profile_list = []
             tx_all_profile = str(self.tx_port) + str(".*")
+            pg_id_list = []
 
             while profile_id <= num_profiles:
                 tx_profile_name = str(self.tx_port) + str(".profile_") + str(profile_id)
@@ -834,18 +835,20 @@ class STLClient_Test(CStlGeneral_Test):
                 profile_id = profile_id + 1
 
             for index, tx_profile in enumerate(tx_profile_list):
+
+                index_pg_id = random.randint(1,num_profiles) + num_profiles * index
+                pg_id_list.append(index_pg_id)
+
                 stream = STLStream(name = 'latency',
                                packet = self.pkt,
                                mode = STLTXCont(percentage = self.percentage),
-#                               flow_stats = STLFlowLatencyStats(pg_id = index))
-                               flow_stats = STLFlowLatencyStats(pg_id = random.randint(10,100) + num_profiles * index))
+                               flow_stats = STLFlowLatencyStats(pg_id = index_pg_id))
                 self.c.add_streams([stream], ports = tx_profile)
-                self.c.start(ports = tx_profile_list)
+                self.c.start(ports = tx_profile)
 
-            for i in range(100):
-                for tx_profile in tx_profile_list:
-                    self.c.pause(tx_profile)
-                    self.c.resume(tx_profile)
+            for tx_profile in tx_profile_list:
+                self.c.pause(tx_profile)
+                self.c.resume(tx_profile)
 
             for i in range(100):
                 self.c.pause(tx_all_profile)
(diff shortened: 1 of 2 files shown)


hhaim:
see results here
https://github.com/cisco-system-traffic-generator/trex-core/pull/264

JunsubHan4rmEricsson:
I think the problem was due to such high PPS for 100 profiles. 
I've decreased the PPS for test cases, 
**1) test_pause_resume_update_dynamic_profile
2) test_pause_resume_update_streams**

JunsubHan4rmEricsson:
2d26abc should be disregarded by the way.
I've made more commits afterwards

JunsubHan4rmEricsson:
Hi Hanoh, 
About the regression failure on 76e2842 on trex08, 
You have previously mentioned that 

**trex-08 does not work (XL710) while trex-09 (X710) works.
it might be related to the fact that trex-08 is connected like this

0->3
1->2
2->1
3->0

Instead of (trex-09)

0->1
1->0
2->3
2->2

This has more race conditions than the normal when one core handles both 0,1 2,3**

What do you suggest from our side now?
It seems that every other error has been fixed.

hhaim:
> Hi Hanoh,
> About the regression failure on [76e2842](https://github.com/cisco-system-traffic-generator/trex-core/commit/76e2842535af0d9d638ac8d5b8a9405abe70de89) on trex08,
> You have previously mentioned that
> 
> > trex-08 does not work (XL710) while trex-09 (X710) works.
> > it might be related to the fact that trex-08 is connected like this
> > 0->3
> > 1->2
> > 2->1
> > 3->0
> > Instead of (trex-09)
> > 0->1
> > 1->0
> > 2->3
> > 2->2
> > This has more race conditions than the normal when one core handles both 0,1 2,3
> 
> What do you suggest from our side now?
> It seems that every other error has been fixed.

Could you elaborate? I don't understand the question.
I've suggested more than on time to build setup like trex-08 in your place (add one card) and find the issue and fix it locally. our regression  setup is only to find issues that related to esoteric setups not for the main stream, simple setup

JunsubHan4rmEricsson:
First, **test_pause_resume_update_dynamic_profile** test case has been fixed.
This is the one that occurred in most of regression setups

Two errors that seem to take place continuously **alternatively** in travis and travis only, not ours.
**1) test_latency_pause_resume_dynamic_profile (takes place only in trex08 setup)**
AssertionError: Port 0 : *** For payload rules: Can't have two streams with same pg_id, or same stream on more than one port

**2) test_pause_resume_update_streams (takes place only in trex19 setup)**
AssertionError: expected TX ~200pps, got: 92.4

Neither of these errors take place in our environment (which is set up exactly like trex 08 of yours)
Here's the setup I'm using for regression. 
**08:00.0 Ethernet controller: Intel Corporation Ethernet Controller XL710 for 40GbE QSFP+ (rev 02)
08:00.1 Ethernet controller: Intel Corporation Ethernet Controller XL710 for 40GbE QSFP+ (rev 02)**

As mentioned, the two test cases do not occur in my environment. 

Now, 
For **1) test_latency_pause_resume_dynamic_profile**, you have mentioned the following.
This is what you said before. 
> 
> 0->3
> 1->2
> 2->1
> 3->0
> 
> Instead of (trex-09)
> 
> 0->1
> 1->0
> 2->3
> 2->2
> 
> This has more race conditions than the normal when one core handles both 0,1 2,3
I'm not exactly sure what you mean with this. 
But you seem to have a clue about why this fails. 
So, my question is. 
What do you suggest that I do?
and what exactly do you mean by what yous said before? 
0->3, 1->2...what do you mean and how do you think it affects the pg_id error?

For **2) test_pause_resume_update_streams**, I did not implement this test case myself.
There seems to be a problem with a performance issue for the following reasons.
1) It only takes place in trex19 only
2) The problem is with the shortage of TX (ex, expected TX ~1200pps, got: 1006.8)
Does this test case pass all the time?
Or do you think that this test case (which I didn't write) is failing because of our recent implementation?

Does it make sense?

hhaim:
1) When the core that Tx the traffic redirect the traffic to different core there are usually new race condition in the code. This is why we loopback trex-08 in this way.  I suggest to connect the loppback like we do and reconstruct and understand the issue
2) The regression without this PR works fine on all the setups . I would look for the reason for 1 first maybe it is related issues.

JunsubHan4rmEricsson:
Hanoh, what's etc/trex_cfg.yaml for the system (trex08)?

PR #264, opened 2019-05-27
From STL_Profile_Regression_Test_Refinement for travis regressiont test 

elados93:
This is a follow up from: [pr 262](https://github.com/cisco-system-traffic-generator/trex-core/pull/262)

### DIFF
--- a/scripts/automation/regression/stateless_tests/stl_client_test.py
+++ b/scripts/automation/regression/stateless_tests/stl_client_test.py
@@ -705,50 +705,101 @@ class STLClient_Test(CStlGeneral_Test):
             assert False , '{0}'.format(e)
 
 
-    def check_error_in_percentage_dynamic_profile(self, golden):
-        self.c.clear_stats()
-        time.sleep(10)
-        stats = self.c.get_stats()
-        check_params = (
-            stats[self.tx_port]['tx_pps'],
-            stats[self.rx_port]['tx_pps'],
-            stats[self.tx_port]['rx_pps'],
-            stats[self.rx_port]['rx_pps'],
-            )
-        for param in check_params: # 5% error is relaxed enough
-            assert get_error_in_percentage(golden, param) < 0.05, 'golden: %s, got: %s' % (golden, param)
+    def test_pause_resume_update_dynamic_profile(self):
+        try:
+            self.c.reset()
+            stream = STLStream(mode = STLTXCont(pps = 400))
 
-    def __basic_cont_dynamic_profile (self):
+            port_id = 0
+            profile_id = 1
+            num_profiles = 100
+            profile_list = []
+            while profile_id <= num_profiles:
+                profile_name = str(port_id) + str(".profile_") + str(profile_id)
+                profile_list.append(profile_name)
+                profile_id = profile_id + 1
 
-        pps = 500
-        port_list = [self.tx_port, self.rx_port]
-        profile_list = ["p1", "p2"]
-        golden = pps * len(profile_list) 
+            self.c.add_streams(stream, ports = profile_list)
+            self.c.start(ports = profile_list)
+
+            for profile in profile_list:
+                assert self.c.ports[port_id].is_transmitting()
+                self.c.pause(ports = profile)
+
+            assert self.c.ports[port_id].is_paused()
+
+            for profile in profile_list:
+                self.c.resume(ports = profile)
+                assert self.c.ports[port_id].is_transmitting()
+
+            self.pause_resume_update_streams_iteration(delay = 5, expected_pps = 40000)
+
+            for index, profile in enumerate(profile_list):
+                if index % 2 == 0:
+                    self.c.pause(ports = profile)
+            self.pause_resume_update_streams_iteration(delay = 5, expected_pps = 20000) # paused stream not transmitting
+            for index, profile in enumerate(profile_list):
+                if index % 2 == 0:
+                    self.c.resume(ports = profile)
+            self.pause_resume_update_streams_iteration(delay = 5, expected_pps = 40000) # resume the paused
+
+            self.c.update(ports = profile_list, mult = '10kpps')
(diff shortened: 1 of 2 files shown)


hhaim:
trex-08 does not work (XL710) while trex-09 (X710) works.
it might be related to the fact that trex-08 is connected like this

0->3
1->2
2->1
3->0

Instead of (trex-09)

0->1
1->0
2->3
2->2

This has more race conditions than the normal when one core handles both 0,1 2,3

hhaim:
the other type of error 
```~1000000pps, got: 785791.6``` 
Could be fixed by adding a specific test tunable per setup 

see example here 

you should add an object to this file 
https://github.com/cisco-system-traffic-generator/trex-core/blob/master/scripts/automation/regression/setups/trex08/benchmark.yaml#L234

and use it in the test like this 

it  take the per test tunable from the benchmark.yaml 

```
        dp_cores     = self.stl_trex.system_info.get('dp_core_count', 0)
```

JunsubHan4rmEricsson:
I think the problem was due to such high PPS for 100 profiles. 
I've decreased the PPS for test cases, 
**1) test_pause_resume_update_dynamic_profile
2) test_pause_resume_update_streams**

JunsubHan4rmEricsson:
> the other type of error
> `~1000000pps, got: 785791.6`
> Could be fixed by adding a specific test tunable per setup
> 
> see example here
> 
> you should add an object to this file
> https://github.com/cisco-system-traffic-generator/trex-core/blob/master/scripts/automation/regression/setups/trex08/benchmark.yaml#L234
> 
> and use it in the test like this
> 
> it take the per test tunable from the benchmark.yaml
> 
> ```
>         dp_cores     = self.stl_trex.system_info.get('dp_core_count', 0)
> ```

I think the problem was due to such high PPS for 100 profiles. 
I've decreased the PPS for test cases, 
**1) test_pause_resume_update_dynamic_profile
2) test_pause_resume_update_streams**

JunsubHan4rmEricsson:
> 
> 0->3
> 1->2
> 2->1
> 3->0
> 
> Instead of (trex-09)
> 
> 0->1
> 1->0
> 2->3
> 2->2
> 
> This has more race conditions than the normal when one core handles both 0,1 2,3

What do you suggest I do for this?
It seems to do with your test setup

hhaim:
> > the other type of error
> > `~1000000pps, got: 785791.6`
> > Could be fixed by adding a specific test tunable per setup
> > see example here
> > you should add an object to this file
> > https://github.com/cisco-system-traffic-generator/trex-core/blob/master/scripts/automation/regression/setups/trex08/benchmark.yaml#L234
> > and use it in the test like this
> > it take the per test tunable from the benchmark.yaml
> > ```
> >         dp_cores     = self.stl_trex.system_info.get('dp_core_count', 0)
> > ```
> 
> I think the problem was due to such high PPS for 100 profiles.
> I've decreased the PPS for test cases,
> **1) test_pause_resume_update_dynamic_profile 2) test_pause_resume_update_streams**

This is not the right way to solve this problem. There are setups that are very slow (10PPS), does it mean that 100mpps setup will test with the same rate? 
Have a look how I suggested to solve this.

hhaim:
> > trex-08 does not work (XL710) while trex-09 (X710) works.
> > it might be related to the fact that trex-08 is connected like this
> > 0->3
> > 1->2
> > 2->1
> > 3->0
> > Instead of (trex-09)
> > 0->1
> > 1->0
> > 2->3
> > 2->2
> > This has more race conditions than the normal when one core handles both 0,1 2,3
> 
> What do you suggest I do for this?
> It seems to do with your test setup

This type of setups are more prone to software race conditions. The best way to solve it is to reconstruct with the same setup in your place and find/solve the rate

JunsubHan4rmEricsson:
> > > the other type of error
> > > `~1000000pps, got: 785791.6`
> > > Could be fixed by adding a specific test tunable per setup
> > > see example here
> > > you should add an object to this file
> > > https://github.com/cisco-system-traffic-generator/trex-core/blob/master/scripts/automation/regression/setups/trex08/benchmark.yaml#L234
> > > and use it in the test like this
> > > it take the per test tunable from the benchmark.yaml
> > > ```
> > >         dp_cores     = self.stl_trex.system_info.get('dp_core_count', 0)
> > > ```
> > 
> > 
> > I think the problem was due to such high PPS for 100 profiles.
> > I've decreased the PPS for test cases,
> > **1) test_pause_resume_update_dynamic_profile 2) test_pause_resume_update_streams**
> 
> This is not the right way to solve this problem. There are setups that are very slow (10PPS), does it mean that 100mpps setup will test with the same rate?
> Have a look how I suggested to solve this.

I understand but the original PPS for the first test was just too high. 
Let's look at **test_pause_resume_update_dynamic_profile** for example (the one with the most errors)
Line 747 and line 752 continue to show errors in this test case. 
expected_pps was 1000,000 and 2,000,000 which is just too high. 

These test cases have been passing under 40G setup.
However, under 10G setup (yours) test case continues to show the shortage of PPS
**AssertionError: expected TX ~1000000pps, got: 785791.6**

That is why me and my team members decided to lower the PPS transferred

JunsubHan4rmEricsson:
> > > trex-08 does not work (XL710) while trex-09 (X710) works.
> > > it might be related to the fact that trex-08 is connected like this
> > > 0->3
> > > 1->2
> > > 2->1
> > > 3->0
> > > Instead of (trex-09)
> > > 0->1
> > > 1->0
> > > 2->3
> > > 2->2
> > > This has more race conditions than the normal when one core handles both 0,1 2,3
> > 
> > 
> > What do you suggest I do for this?
> > It seems to do with your test setup
> 
> This type of setups are more prone to software race conditions. The best way to solve it is to reconstruct with the same setup in your place and find/solve the rate

What do you mean rate?
aren't you talking about the pg_id error?
**trex.common.trex_exceptions.TRexError: Port 0 : *** For payload rules: Can't have two streams with same pg_id, or same stream on more than one port**

PR #272, opened 2019-06-04
[test_this_on_travis] from #262

JunsubHan4rmEricsson:
Merged from #262 for travis regression test

PR #277, opened 2019-06-05
Solution for STL Regression test case, "test_latency_pause_resume_dynamic_profile" and its explanation. 

JunsubHan4rmEricsson:
Hi Hanoh. 
This thread is extended from #224 , #262 , #272 , #273 .
As you know, the problem has been with the new regression test case, **"test_latency_pause_resume_dynamic_profile"**
> AssertionError: Port 0 : *** For payload rules: Can't have two streams with same pg_id, or same stream on more than one port

The problem above has been taking place even though the assigned pg_ids are all unique in the new test case. 
However, I think we have found the solution for this problem. 

First of all, as you have mentioned in #262 trex-08 setup is like this. 
> 
> 0->3
> 1->2
> 2->1
> 3->0

In the previous test cases before **test_latency_pause_resume_dynamic_profile**,
test case, **test_core_pinning_latency** assigns pg_id, 3 to port 0 and pg_id, 4 to port 1. 

This is what's causing the problem. 
Instead of assigning each pg_id to self.tx_port and self.rx_port it uses **hard-coded integer** 0 and 1 as port numbers. 

**Neither cleanUp nor setUp function will reset port 1 in this case, because both functions only reset self.tx_port(0) and self.rx_port(3) in this trex-08 setup.** 

**Therefore pg_id, 4, is left uncleaned in port 1.** 

When I try the new test case "test_latency_pause_resume_dynamic_profile", I also assigned pg_id 4 to self.tx_port. 
**However, this pg_id is already occupied by port 1.** 
Therefore such error has taken place. 

I have updated the code so that it uses other pg_ids. 
I think it's going to pass the regression. 

However, there are several ways to solve this problem. 
1) **test_core_pinning_latency** uses designated ports, self.tx_port and self.rx_port, instead of 0 and 1 (as well as other test cases)
2) **test_latency_pause_resume_dynamic_profile** uses different pg_ids as implemented now. 
3) **setUp** and **cleanup** function reset all ports, not just self.tx_port and self.rx_port

### DIFF
--- a/scripts/automation/regression/stateless_tests/stl_client_test.py
+++ b/scripts/automation/regression/stateless_tests/stl_client_test.py
@@ -43,7 +43,9 @@ class STLClient_Test(CStlGeneral_Test):
         self.tx_port, self.rx_port = CTRexScenario.ports_map['bi'][0]
 
         self.c.connect()
-        self.c.reset(ports = [self.tx_port, self.rx_port])
+#        self.c.reset(ports = [self.tx_port, self.rx_port])
+        for tx_port, rx_port in CTRexScenario.ports_map['map'].items():
+            self.c.reset(ports = [tx_port, rx_port])
 
         port_info = self.c.get_port_info(ports = self.rx_port)[0]
 
@@ -65,7 +67,9 @@ class STLClient_Test(CStlGeneral_Test):
         
     def cleanup (self):
         self.c.remove_all_captures()
-        self.c.reset(ports = [self.tx_port, self.rx_port])
+#        self.c.reset(ports = [self.tx_port, self.rx_port])
+        for tx_port, rx_port in CTRexScenario.ports_map['map'].items():
+            self.c.reset(ports = [tx_port, rx_port])
         
             
     @classmethod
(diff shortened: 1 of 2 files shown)


JunsubHan4rmEricsson:
Furthermore, I have a request to make
**https://trex-tgn.cisco.com/trex/reports/** does not show my regression test.
I would like to see the traces I left. 
Could you please help me out with this?

JunsubHan4rmEricsson:
> Furthermore, I have a request to make
> **https://trex-tgn.cisco.com/trex/reports/** does not show my regression test.
> I would like to see the traces I left.
> Could you please help me out with this?

P

hhaim:
It seems there is an issue with Travis due to the last changes.  We are  working on this. 
It is clear why it happen only on trex-08 now.  Nice work !
I think we better do option 3 and option 2 
I will run the regression internally to validate it. 


However I think we are still missing this:
1. update the rates to be setup specific (not the lowest to match all) see example how to do it here 
(here)[https://github.com/cisco-system-traffic-generator/trex-core/blob/master/scripts/automation/regression/stateless_tests/stl_rx_test.py#L87]

```latency_9k_enable``` this is tuned per setup 

2. more resilience/random test for add/remove profiles in event driven see my old comments about the suggestion.

1+2 should be a new PR. This PR will be committed

**Comment by hhaim** [comment]
Regression works fine!

**Comment by JunsubHan4rmEricsson** [comment]
Hi Hanoh, I've applied solution #2, #3 as mentioned. 
Please run the regression and accept the PR. 

1+2 will be a new PR as mentioned
Thanks.

**Comment by hhaim** [comment]
Ack.

**Comment by JunsubHan4rmEricsson** [comment]
Are these pulled?

**Comment by hhaim** [comment]
reabsed

**Comment by hhaim** [comment]
see the fix for the test 
https://github.com/cisco-system-traffic-generator/trex-core/commit/b609ab460efeb59384cbcebc866b4e01b930d3cc

The golden is not accurate. only with burst you can do this math 
```
File "/tmp/trex-scripts/automation/regression/stateless_tests/stl_client_test.py", line 816, in test_basic_cont_dynamic_profile
assert get_error_in_percentage(golden, param) < 0.05, 'golden: %s, got: %s' % (golden, param)
AssertionError: golden: 500000, got: 525527
AssertionError: golden: 500000, got: 525510
```

PR #280, opened 2019-06-10
Fix mixed up Ether src/dst field offsets in Scapy service

Rhymmor:
Hi guys,
Small PR for fixing Ether offsets for Scapy server


hhaim:
Thanks. rebased

PR #283, opened 2019-06-13
ASTF dynamic profile is available in CP side

yonghwan007:
The summary of dynamic multiple profile in CP side.

1. Support multiple profiles in CP side
2. Support statisctics for each profile
3. Added new RPC command 'get_total_counter_values' for statistics summary of all profiles
4. Added new event through asynchronous port.
 - EVENT_ASTF_PROFILE_STATE_CHG = 60 : event for each profile state change
 - EVENT_ASTF_PROFILE_CLEARED   = 61 : event for each rofile clear(remove database)
5. Added TrexAstfDeleteDB message between CP and DP to remove CAstfDB.

I have done regression test in our environment.

### DIFF
--- a/src/main_dpdk.cpp
+++ b/src/main_dpdk.cpp
@@ -4229,9 +4229,14 @@ void CGlobalTRex::update_stats(){
     m_last_total_cps = m_cps.add(total_open_flows);
 
     bool all_init=true;
-    CSTTCp  * lpstt;
-    lpstt = m_fl.m_stt_cp;
-    if ( lpstt ){
+    vector<CSTTCp *> sttcp_list;
+    if ( get_is_interactive() && get_is_tcp_mode() ) {
+        sttcp_list = get_astf_object()->get_sttcp_list();
+    }
+    else if ( m_fl.m_stt_cp ) {
+        sttcp_list.push_back(m_fl.m_stt_cp);
+    }
+    for ( auto lpstt : sttcp_list ) {
         if (!lpstt->m_init){
             /* check that we have all objects;*/
             for (i=0; i<get_cores_tx(); i++) {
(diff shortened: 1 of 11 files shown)


yonghwan007:
I have finished full regression on my setups, and all tests were passed.

hhaim:
Hi @yonghwan007 the ASTF DP from @jsmoon has a crash., I'm waiting for a fix . After the fix worth verifying that both patches (CP+DP) does not have this issue.

yonghwan007:
I rebased from master including DP code.
I have done full regression test, and all tests were passed.

hhaim:
Hi @yonghwan007 
please consult with  @jsmoon it seems that reading the counters are not protected as it could be changed by DP. We had a crash in the DP change and the workaround was to create the default context once. 
This workaround won't work for the general case. 
My suggestion for now is to keep only the default profile for now (read only default) and in a diff PR to solve the counters issue 

```
@@ -52,11 +53,11 @@ void CSTTCpPerTGIDPerDir::update_counters(bool is_sum, uint16_t tg_id) {
        uint64_t *base_tcp;
        uint64_t *base_udp;
        if (is_sum) {
            base_tcp = (uint64_t *)&lpctx->get_tcpstat()->m_sts;
            base_udp = (uint64_t *)&lpctx->get_udpstat()->m_sts;
            base_tcp = (uint64_t *)&lpctx->get_tcpstat(m_stt_id)->m_sts;
            base_udp = (uint64_t *)&lpctx->get_udpstat(m_stt_id)->m_sts;

```

hhaim:
BTW regression is OK

hhaim:
```
    void Create(uint32_t stt_id=0, uint16_t num_of_tg_ids=1, bool first_time=true);
```

Understood. The code will work for default profile *but* not for multi-profile. 
Suggest to fix the general case and "remove" case as I proposed in the other thread

yonghwan007:
Okay, I understood.
I am out of office during this week, so I will implement a proper solution for general case on Monday.

hhaim:
rebased.

PR #289, opened 2019-06-25
revert __rte_raw_cksum change does not work properly at big-endian machine

jsmoon:
Hi,
Recently, I found my commit f5319632 can make a potential problem on big-endian machines.
There is no issue on little-endian machines.
It was my fault at that time. Now I want to revert the change to fix it.


hhaim:
How did you find it? Do you have ARM

jsmoon:
No, I don't have it.
When the DPDK version upgraded to 19.05 in TRex, I've recalled my DPDK change and checked it.
When I reviewed my code again, I realized that my change may cause a problem on big-endian machines.
So I think that it should be reverted to the original code.

hhaim:
Understood

hhaim:
rebased

PR #296, opened 2019-07-02
added a stateless test for raw packet stream statistics

jsmoon:
This PR is related to #293 "enable latency statistics of custom ethertype packet stream".

I added a condition to run this test like this:
```
        if not self.is_virtual or not self.is_multiqueue_mode:
            self.skip('Skip this for unsupported modes')
```

I think that ```is_multiqueue_mode``` is not enough because it may be true on some real NIC.
So, I added a condition ```is_virtual``` from port 0.

### DIFF
--- a/scripts/automation/regression/stateless_tests/stl_rx_test.py
+++ b/scripts/automation/regression/stateless_tests/stl_rx_test.py
@@ -47,7 +47,7 @@ class STLRX_Test(CStlGeneral_Test):
         self.num_cores = system_info.get('dp_core_count', 'Unknown')
         self.is_multiqueue_mode = system_info.get('is_multiqueue_mode', False)
         self.is_vxlan_supported = system_info['ports'][0]['is_vxlan_supported']
-        self.is_virtual = system_info['ports'][0]['is_virtual']
+        self.advanced_per_stream_stats = system_info.get('advanced_per_stream_stats', False)
         mbufs = self.c.get_util_stats()['mbuf_stats']
         # currently in MLX drivers, we use 9k mbufs for RX, so we can't use all of them for TX.
         if self.drv_name == 'net_mlx5':
@@ -441,7 +441,7 @@ class STLRX_Test(CStlGeneral_Test):
 
     @try_few_times_on_vm
     def test_raw_pkt_stream(self):
-        if not self.is_virtual or not self.is_multiqueue_mode:
+        if not self.advanced_per_stream_stats or not self.is_multiqueue_mode:
             self.skip('Skip this for unsupported modes')
 
         total_pkts = self.total_pkts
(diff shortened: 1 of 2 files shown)


hhaim:
Hi @jsmoon 
Any platform with is_multiqueue_mode==True should support this. 
So it is not relevant if it is not virtual  or not .

jsmoon:
As far as I know, TRex has four dpdk modes: ONE_QUE, MULTI_QUE, DROP_QUE_FILTER and RSS_DROP_QUE_FILTER.
When TRex is running on MUTLI_QUE and RSS_DROP_QUE_FILTER mode, **is_multiqueue_mode** is true.
But, my implementation used **! is_hardware_filter_needed()** to make it available and the result is **different** between MULTI_QUE and RSS_DROP_QUE_FILTER mode.
So, I thought that is_virtual could filter out the case which is_multiqueue_mode and is_hardware_filter_needed() is true and it is enough for this test.

ybrustin:
recheck

ybrustin:
recheck

hhaim:
You are right!

So better to add an explicit field  that specifies that advance per-stream stats is supported. 
something like this:

```
section["advace_per_stream_stats"] = get_dpdk_mode()->!is_hardware_filter_needed();
```

hhaim:
Hi @jsmoon 
Practically in STL mode, there could be only  mDROP_QUE_FILTER, tmMULTI_QUE,mONE_QUE modes. 
So advace_per_stream_stats would be like multi-queue. However, it is good to have it for future. 

```
    modes.push_back( {OP_MODE_STF, true,  { tmDROP_QUE_FILTER,tmONE_QUE } } );
    modes.push_back( {OP_MODE_STF, false, { tmDROP_QUE_FILTER } } );

    modes.push_back( {OP_MODE_STL, true,  { tmDROP_QUE_FILTER,tmONE_QUE } } );
    modes.push_back( {OP_MODE_STL, false, { tmDROP_QUE_FILTER,tmMULTI_QUE } } );

    modes.push_back( {OP_MODE_ASTF_BATCH, true,  { tmRSS_DROP_QUE_FILTER,tmONE_QUE } } ); /* only filter is needed here */
    modes.push_back( {OP_MODE_ASTF_BATCH, false, { tmRSS_DROP_QUE_FILTER } } );

    modes.push_back( {OP_MODE_ASTF, true,  { tmRSS_DROP_QUE_FILTER,tmONE_QUE } } ); 
    modes.push_back( {OP_MODE_ASTF, false, { tmRSS_DROP_QUE_FILTER,tmMULTI_QUE } } );

```

hhaim:
rebased.

jsmoon:
Hi Hanoh,
Thank you for your additional comment about the point I missed. I totally
agree with you.

PR #309, opened 2019-07-22
Cannot start new profile during latency profile running

yonghwan007:
If one profile started with latency enabled, then other new profile cannot be started during latency enabled profile running.

Reproduction:

trex>start -l 1 -f astf/udp1.py --pid udp
trex>start -f astf/http_simple.py --pid tcp

Last command failed: Latency state is not idle, should stop latency first


hhaim:
rebased

Issue #307, opened 2019-07-16
TRex error on retreiving statistics when connected to several instances


fperrin:
Hi,

When using the ASTF Python API, connected to several TRex instances, when running `get_xstats`, I get the following error:

```
>>> trex5.trex.get_xstats(0)                                                                                          
                                                                                                                      
Traceback (most recent call last):                                                                                    
  File "<console>", line 1, in <module>                                                                               
  File "/home/fperrin/src/trex-core-2.57/scripts/automation/trex_control_plane/interactive/trex/common/trex_api_annota
tors.py", line 51, in wrap2                                                                                           
    ret = f(*args, **kwargs)                                                                                          
  File "/home/fperrin/src/trex-core-2.57/scripts/automation/trex_control_plane/interactive/trex/common/trex_client.py"
, line 1153, in get_xstats                                                                                            
    xstats.update_sync(self.conn.rpc)                                                                                 
  File "/home/fperrin/src/trex-core-2.57/scripts/automation/trex_control_plane/interactive/trex/common/stats/trex_stat
s.py", line 122, in update_sync                                                                                       
    self._update_common(rc.data())                                                                                    
  File "/home/fperrin/src/trex-core-2.57/scripts/automation/trex_control_plane/interactive/trex/common/stats/trex_stat
s.py", line 245, in _update_common                                                                                    
    snapshot = self._pre_update(snapshot)                                                                             
  File "/home/fperrin/src/trex-core-2.57/scripts/automation/trex_control_plane/interactive/trex/common/stats/trex_port
_stats.py", line 228, in _pre_update                                                                                  
    % (len(self.names), len(values), self.names, values))                                                             
TRexError: Length of get_xstats_names: 158 and get_port_xstats_values: 110                                            
```

Some TRex instances use an Intel Corporation 82599ES 10-Gigabit SFI/SFP+ NIC; some use XXV710 for 25GbE SFP28. When I connect to only one TRex instance at a time, the length of get_xstat() for the first one is 158, and is 110 for the second one.

I think the issue is that in `trex_port_stats.py` the `get_names` function is a class method. However these two cards don't have the same set of available stat counters, so shouldn't share the list of stat names. The following diff fixes the issue for me, but I don';t know enough about TRex to offer it as a proper PR; in particular do other uses of TRexClient.any_port need changing too?

```
--- stats/trex_port_stats.py~   2019-07-15 11:46:28.081513032 +0100
[...]

PR #311, opened 2019-07-22
ASTF: allow different stat counters on clients

fperrin:
Closes: #307

Signed-off-by: Frédéric Perrin <fred@fperrin.net>

### DIFF
--- a/scripts/automation/trex_control_plane/interactive/trex/common/stats/trex_port_stats.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/common/stats/trex_port_stats.py
@@ -204,19 +204,25 @@ class PortStats(AbstractStats):
 ############################                #############################
 ############################                #############################
 class PortXStats(AbstractStats):
-    names = []
     def __init__(self, port_obj):
         super(PortXStats, self).__init__(RpcCmdData('get_port_xstats_values', {'port_id': port_obj.port_id}, ''))
 
         self.port_id = port_obj.port_id
+        self.names = []
 
-
-    @classmethod
-    def get_names(cls, port_obj):
-        rc = port_obj.rpc.transmit('get_port_xstats_names', params = {'port_id': port_obj.port_id})
+        
+    def get_names(self, all_ports):
+        port = all_ports[0]
+        rc = port.rpc.transmit('get_port_xstats_names', params = {'port_id': port.port_id})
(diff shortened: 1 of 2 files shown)


ybrustin:
We have an error in regression report script.
Please sync with latest master as it should fix the issue.

fperrin:
Hi @ybrustin , since the first job failed with an encoding error, I assumed the problem was my name `Frédéric` containing non-ASCII characters. So I resubmitted with the author name changed to `Frederic`. Well trex-bot is happy now, so I won't change it any further.

ybrustin:
Hi,

Encoding-wise its fine.

Regarding the PR itself:
For someone who has many ports (let's say 16) it will mean many RPC requests.

Better to keep it per class, but save the xstat names in dict with key of client_object.ctx.session_id
(Need also to cleanup names upon client disconnect)

Even better, the function get_names can be static and called by client [once] upon connection to return the names.
Afterwards, it will be assigned to all port objects as copy.

fperrin:
If I'm not mistaken get_names is only called from TRexClient._connect(). So the cost of the RPC calls will only be born at the time of connection. Is that too expensive?

I'm not sure you want to index by client ID though. Can you have port 1 on, say, a 10G interface and port 2 on a 25G interface? In that case, we could cache on interface type instead. (Using port_obj.driver? Are all interfaces using a given driver exporting the same set of stats?)

hhaim:
This is the issue, with 32 ports you will run the same RPC 32 times. 
```
   for port in self.ports.values():
            port.xstats.get_names(port)
```

The suggestion  is to read it once for all the ports as it is not supported to have multi type ports in one server.

ybrustin:
> I'm not sure you want to index by client ID though. Can you have port 1 on, say, a 10G interface and port 2 on a 25G interface? In that case, we could cache on interface type instead. (Using port_obj.driver? Are all interfaces using a given driver exporting the same set of stats?)

Per TRex server instance [i.e. per current connection] all names are the same for all ports.

fperrin:
Changed so that the names are fetched once and then copied to all ports accessed by the TRex client.

Second push: python3 dict.values() is not a list.

ybrustin:
Rebased to master with small refactor

fperrin:
Thanks @ybrustin

PR #312, opened 2019-07-23
crash when using rpc commands with undefined profile_id on astf mode

Cyborn-Minjae-Lee:
If Client sends Json RPC commands with undefined profile_id to Server of astf mode, the crash is occurred.

 The related commands :
 get_stats 
 get_tg_id_stats 
I add the code to validate profile_id.


hhaim:
rebased.

Issue #305, opened 2019-07-16
Crash issues from wrong API usage in ASTF mode


egwakim:
Hi,
I found following 5 issues in ASTF mode.
These are caused from wrong API usage from user, but, I think it needs some defense code to report proper error messages instead of crashing TRex server.

We have 2 options to fix it.
Option#1. Defense from trex server : Complex to fix, more robust and promising approach
Option#2. Defense from Python/Java API : Simpler to fix, reduce burden of trex server
Which approach would be preferred option in trex point of view?

Issue#1. TRex server crash when user used same client/server IP in ASTF mode
Reproduction>
ip_gen_c = ASTFIPGenDist(ip_range=["16.0.0.2", "16.0.0.2"], distributionn
="seq")
ip_gen_s = ASTFIPGenDist(ip_range=["16.0.0.2", "16.0.0.2"], distributionn
="seq")

Issue#2. TRex server crash when user used big set_var in ASTF UDP mode
If user used over than 88000 loops in UDP, then, trex server was crashed.
Reproduction>
loop = 88000
prog_c.set_var("var1",loop); #
prog_c.set_label("a:");
prog_c.send_msg('*')
prog_c.jmp_nz("var1","a:")

Issue#3. TRex server crash when same port used in the multiple template group in ASTF mode
If user used same port number in the multiple template group in ASTF mode, TRex server was crashed.
Reproduction>
temp_c1 = ASTFTCPClientTemplate(port=80, program=prog_c,ip_gen=ip_gen, cc
ps = 1)
temp_c2 = ASTFTCPClientTemplate(port=80, program=prog_c, ip_gen=ip_gen, 
cps = 2)
temp_s1 = ASTFTCPServerTemplate(program=prog_s, assoc=ASTFAssociationRull
e(80)) # using default association
temp_s2 = ASTFTCPServerTemplate(program=prog_s, assoc=ASTFAssociationRull
e(80)) # using default association

Issue#4. TRex server crash when user used get_stats(pid_input='*') in ASTF mode
Reproduction>
stats = c.get_stats(pid_input='*')

Issue#5. Not possible to find exact error reason when same port used between multiple profiles
If user used same port number in the different profile, then, TRex reports 'Unknown error' when starting TRex. It's not easy to find user fault(same port used) in the user point of view.
Reproduction>
Run 2 profiles with same port number.

Best Regards
Gwangmoon

hhaim:
It should be solved both in the server and in the client
[...]

PR #315, opened 2019-07-29
make recursive CEmulApp::process_cmd() iterative

jsmoon:
This PR is related to #305 the second issue((Issue#2. TRex server crash when the user used big set_var in ASTF UDP mode).
According to my examination, the reason for the crash is a stack overflow.
Since CEmulApp::process_cmd() and CEmulApp::next() were recursively called by the programmed loop, stack memory was exhausted by a large value of the loop.
So, I changed the recursive routine to iterative.

I hope my change will remove the massive stack usage caused by huge recursive calls.

### DIFF
--- a/src/44bsd/tcp_socket.cpp
+++ b/src/44bsd/tcp_socket.cpp
@@ -325,6 +325,12 @@ void CEmulApp::check_rx_pkt_condition(){
 
 */
 void CEmulApp::process_cmd(CEmulAppCmd * cmd){
+    while(cmd) {
+       cmd = process_cmd_one(cmd);
+    }
+}
+
+CEmulAppCmd* CEmulApp::process_cmd_one(CEmulAppCmd * cmd){
 
     EMUL_LOG(cmd, "CMD [%d] state : %d ,cmd_index [%d] -",m_debug_id,m_state,m_cmd_index);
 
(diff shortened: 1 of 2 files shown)


hhaim:
It looks as a better implementation.
Is there an example from our profiles that runs many commands in the stack?

hhaim:
I see the example. It is not typical to have such a loop. Usually we add delay that break the program. This type of loop could stall the overall events loop.

egwakim:
Yes, it's not typical, but, user may send 90000 successive burst packets in some test scenarios.
In current TRex, TRex will be crashed by 'stack overview' from recursive calls.
TRex should be not be crashed in any cases.

hhaim:
Understood. Maybe with the new implementation  you can limit the number of commands in one loop to 1000.  More could be done by adding delay command.

jsmoon:
Which value do you think can be reasonable for the implicit delay command?

egwakim:
In case of event starvation issue, I think it's better to let user to add delay in user profile.
But, we need to defense crash from stack overflow.

hhaim:
@egwakim agreed. Delay should be implicit. We just need to protect event loop. Without that there could be a watchdog that crash the server.

jsmoon:
@hhaim, I added an implicit 1 tick delay. Please check about it.

hhaim:
Nice!

hhaim:
rebased.

PR #316, opened 2019-07-30
fix crash when duplicated port used in templates

jsmoon:
This PR is related to #305 the third issue(Issue#3. TRex server crash when same port used in the multiple template group in ASTF mode).
It comes from a lack of defense code when multiple profile support implemented.
I added a verification code for this issue in the client side also.


hhaim:
rebased

Issue #318, opened 2019-08-07
truncated messages in server console log


jsmoon:
After performing run_regression, I checked the server log(/tmp/trex.txt) and found it does not contain full console messages.
Here are console output messages:
```
The ports are bound/configured.
Starting  TRex v2.53 please wait  ... 
 set driver name net_e1000_em 
 driver capability  : TCP_UDP_OFFLOAD 
 set dpdk queues mode to ONE_QUE 
 Number of ports found: 2
zmq publisher at: tcp://*:4500
 wait 1 sec .
port : 0 
------------
link         :  link : Link Up - speed 1000 Mbps - full-duplex
promiscuous  : 0 
port : 1 
------------
link         :  link : Link Up - speed 1000 Mbps - full-duplex
promiscuous  : 0 
 number of ports         : 2 
 max cores for 2 ports   : 1 
 tx queues per port      : 3 
 -------------------------------
RX core uses TX queue number 65535 on all ports
 core, c-port, c-queue, s-port, s-queue, lat-queue
 ------------------------------------------
 1        0      0       1       0      0  
 -------------------------------
Cleanup of old namespaces related to Linux-based stack
Cleanup Done
```
The contents in the /tmp/trex.txt are:
```
The ports are bound/configured.
Starting  TRex v2.53 please wait  ... 
 set driver name net_e1000_em 
 driver capability  : TCP_UDP_OFFLOAD 
 set dpdk queues mode to ONE_QUE 
 Number of ports found: 2
zmq publisher at: tcp://*:4500
 wait 1 sec .
```

I thought _t-rex-64 stdout is full-buffered mode.
When I used ```stdbuf -oL``` for _t-rex-64, I can get the same messages as the console messages.
So, I think _t-rex-64 stdout should be line-buffered mode or non-buffered mode.
@hhaim, what do you think of it?

hhaim:
There are two way we run the servers. 
One with ```--iom 0 ```  for the regression/automation to find mostly issue -- in this case line-buffer would be good. Without that it will influence the TUI server performance. 

If you could enable this (line buffer) only when --iom=0 it would be great.

PR #319, opened 2019-08-08
set stdout line buffered when --iom is 0

jsmoon:
This PR is related to issue #318 (truncated messages in the server console log.

As you suggested, I set stdout line-buffered mode only when --iom 0 is specified.


hhaim:
rebased

PR #323, opened 2019-08-12
Fix building with GCC 9

CtrlZvi:
GCC 9 added the no-address-of-packed-member check to Wall, which causes
the build to fail under GCC. Clang already had that check, and the
warning is explicitly disabled for the clang build, so we disable it
for GCC as well.

Signed-off-by: Zvi Effron <zeffron@hmc.edu>

### DIFF
--- a/linux_dpdk/ws_main.py
+++ b/linux_dpdk/ws_main.py
@@ -55,7 +55,8 @@ gcc_flags = ['-Wall',
              '-Werror',
              '-Wno-literal-suffix',
              '-Wno-sign-compare',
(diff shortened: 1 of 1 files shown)


CtrlZvi:
I've looked through the failed build, and I don't see how disabling a warning that GCC 6 doesn't support could have impacted performance? It also looks like the failure may be related to something else, as the failing test spews several lines of "i40e_ethertype_filter_set(): There's no corresponding ethertype filter!"

hhaim:
It's related to some other fix

Issue #328, opened 2019-08-28
Garbage info in multiple ASTFProfile()


egwakim:
Current Python API accumulate ASTFIPGen info when creating multiple profiles.
If we create large number of profiles, then, it'll create huge amount of garbage info in every RPC command.
User can call ASTFProfile.clear_cache() before creating new ASTFProfile().
But, it's usage is not the graceful approach.

We have 2 options,
Option#1. Let user use ASTFProfile.clear_cache() before creating new ASTFProfile() and update document about it.
Option#2. Change ASTFProfile API to clear_cache automatically

Do you have any opinion about these options?

Best Regards
Gwangmoon

hhaim:
Maybe it is a better to use a LRU with limited size. there is a Python LRU library that we already using

egwakim:
It's not possible to define the size of the cache because we cannot expect the buffer_list().
Instead, it's reasonable to clear inside load_profile() after convert to json because it'll be not used anymore. 
See line 475 below.
The only possible issue is that ASTFProfile() object usage after load_profile() call.
But, I cannot find such usage in current ASTFProfile().

`scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py

 449     @client_api('command', True)
 450     def load_profile(self, profile, tunables = {}, pid_input = DEFAULT_PROFILE_ID):
 451         """ Upload ASTF profile to server
 452 
 453             :parameters:
 454                 profile: string or ASTFProfile
 455                     Path to profile filename or profile object
 456 
 457                 tunables: dict
 458                     forward those key-value pairs to the profile file 
 459 
 460                 pid_input: string
 461                     Input profile ID
 462 
 463             :raises:
 464                 + :exc:`TRexError`
 465 
 466         """
 467 
 468         if not isinstance(profile, ASTFProfile):
 469             try:
 470                 profile = ASTFProfile.load(profile, **tunables)
 471             except Exception as e:
 472                 self.astf_profile_state.pop(pid_input, None)
 473                 raise TRexError('Could not load profile: %s' % e)
 474         profile_json = profile.to_json_str(pretty = False, sort_keys = True)
 **475 +       profile.clear_cache()**
 476 
 477         self.ctx.logger.pre_cmd('Loading traffic at acquired ports.')
 478         rc = self._upload_fragmented('profile_fragment', profile_json, pid_input = pid_input)
 `

hhaim:
Agreed

[...]

PR #330, opened 2019-08-30
Fix garbage info in multiple ASTFProfile

egwakim:
Issue #328 : Fix garbage info when multiple ASTFProfile creation

### DIFF
--- a/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
@@ -472,6 +472,7 @@ class ASTFClient(TRexClient):
                 self.astf_profile_state.pop(pid_input, None)
                 raise TRexError('Could not load profile: %s' % e)
         profile_json = profile.to_json_str(pretty = False, sort_keys = True)
+        profile.clear_cache()
 
         self.ctx.logger.pre_cmd('Loading traffic at acquired ports.')
         rc = self._upload_fragmented('profile_fragment', profile_json, pid_input = pid_input)


egwakim:
Regression failed, but, it seems issue in testing environment.
Can you check the issue?

hhaim:
It is not related to testing environment. It relates to template group.
I would look into the specific test to understand why this patch does not work in this case

hhaim:
The issue is that the cache (buffers) is static. I think it is better to make it per object

egwakim:
Current ASTF programming model is not designed as per object.
ASTFProfile using cached ASTFProgram, ASTFIPGen, ASTFTemplate.
To be able to support both existing test case which using cached ASTFProfile and cleaning up cached old profile, it's better clean-up multi-profile only.


       profile_json = profile.to_json_str(pretty = False, sort_keys = True)
+        if pid_input != DEFAULT_PROFILE_ID :
+            profile.clear_cache()

        self.ctx.logger.pre_cmd('Loading traffic at acquired ports.')

hhaim:
@egwakim I think it is better to make the cache per object and not global (static)
Your suggestion will work too, but it is a bypass to test case that test only DEFAULT_PROFILE_ID
This test case won't work with multi-profiles without changing the global to per object

ybrustin:
Per-object (and with this PR as well) the cache will not be shared with new profiles.
Each "start" from console, for example, will parse all the pcaps again.

I would suggest to add some sort of TTL to cache entries.
Entry is not used for N profiles => remove it.

hhaim:
It is not about reading the pcap file. it is about the the cache of program/buffers/generators
it should be per profile and not global as in current implementation

egwakim:
We can change class variables(ASTFIPGenDist.in_list, _ASTFTemplateBase.program_list, _ASTFTemplateBase.program_hash) into instance variables.
Then, we do not need to explicitly call clear_cache() because these will be part of object.

But, per-object will cause issues on many test cases because current test case is written by cached assumption. We need to change all these test cases. These test cases re-using templates after load_profiles. That's the issue.

Another option is adding explicit option in load_profile(clear_profile).
clear_profile : default = false, clear profile after load if it's true.
It's similar to first option(Let user to explicit call ASTFProfile.clear_cache() before generate profile).

hhaim:
Could you provide an example for this one?

> But, per-object will cause issues on many test cases because current test case is written by cached assumption. We need to change all these test cases. These test cases re-using templates after load_profiles. That's the issue.
> 

I think that the right solution is to keep the cache per profile (this was the design not sure from where come the global stuff). Explicitly clear of the cache does not make sense. The user doesn't care about our  internal optimization. 

I didn't notice this global cache and it seems to me as a big problem that should be fixed.

egwakim:
astf_template_group_test reuses ipgen object in 2 ASTFProfiles and called load_profile() twice.
I rechecked test case code, it'll not the issue because ipgen object will not be removed by Python GC, it's referenced by other objects. No issue after per-object change.

per-object requires quite a lot change on  trex_astf_profile codes because these are written for classmethod. Converting json code need to be changed and buf_list in ASTFIPGenDist,ASTFProgm ASTFTemplate need to be handled via object method.
And we need to remove all these classmethod after per-object change.

I need to discuss this topic with my team for the priority because it'll need some efforts for the proper change and we have workaround(calling clear_cache).

Best Regards
Gwangmoon

egwakim:
After discussion, we decided to postpone it because most of users are using Java APIs and there is workaround solution of it. It's not the urgent issue.
Change to per-object approach is proper solution.

Regards
Gwangmoon

Issue #327, opened 2019-08-28
Watchdog timeout issues with the thousands of ASTF profiles.


jsmoon:
When I tried to start thousands of ASTF profiles, I observed some WATCHDOG timeouts and found issues on watchdog timer handling.

1. The mismatch on CP master and RPC server timeout values(2 sec and 1 sec) make report the 'RPC server' watchdog even though CP master is causing it.
```
'WATCHDOG: task 'ZMQ sync request-response' has not responded for more than 1.00203 seconds - timeout is 1 seconds

*** traceback follows ***

1             0x54106a ./_t-rex-64() [0x54106a]
2       0x7f0c216f2890 /lib/x86_64-linux-gnu/libpthread.so.0(+0x12890) [0x7f0c216f2890]
3       0x7f0c216f110d /lib/x86_64-linux-gnu/libpthread.so.0(+0x1110d) [0x7f0c216f110d]
4       0x7f0c216ea098 pthread_mutex_lock + 248
5             0x66ceae TrexRpcServerReqRes::process_request_raw(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) + 430
6             0x66d6f8 TrexRpcServerReqRes::process_zipped_request(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) + 456
7             0x66daa5 TrexRpcServerReqRes::handle_request(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 453
8             0x66e1e3 TrexRpcServerReqRes::_rpc_thread_cb_int() + 1283
9             0x66e866 TrexRpcServerReqRes::_rpc_thread_cb() + 6
10      0x7f0c20ff841f so/libstdc++.so.6(+0xb941f) [0x7f0c20ff841f]
11      0x7f0c216e76db /lib/x86_64-linux-gnu/libpthread.so.0(+0x76db) [0x7f0c216e76db]
12      0x7f0c206b988f clone + 63
```
2. CP master thread handles all DP messages and slow path operations without tickle. I think this mechanism may cause watchdog timeout when a large number of DP messages are received at once.
```
WATCHDOG: task 'master' has not responded for more than 2.00402 seconds - timeout is 2 seconds

[...]

PR #333, opened 2019-09-03
improve watchdog timeout handling

jsmoon:
This PR is related to #327.

1. the number of ASTF flow_nodes is updated to 8192(8 times than before).
2. RPC server thread will not receive wrong watchdog timeout any more caused by CP master thread.
3. CP master thread will process up to 100 DP messages at once to avoid watchdog timeout.


hhaim:
looks good

hhaim:
@jsmoon I would revert the last change. I would like to have accurate DP CPU% in all cases.

jsmoon:
@hhaim I agree. I updated by force, is it OK to revert?

hhaim:
It is OK

hhaim:
rebased

PR #334, opened 2019-09-05
Improve update_astf_state() performance

jsmoon:
This PR is for improvement of DP event handling time.

There are 2 main functions to be improved:
1. update_astf_stat() checks all profiles state one by one. --> changed to use profile count for each state.
2. get_profile_id() does a sequential search to all profiles. --> changed to use id map.

According to my measurement, with 4000 profiles, it took about 30ms for processing 100 messages.
Now, it will take about 3ms.


hhaim:
Make sense. the read counters could be improved too. there is no need to read all the 4000 in one time. could be split

jsmoon:
My team has another plan to improve update_stats() performance too.
But it will take time.

jsmoon:
@hhaim should I do something further for this PR?

hhaim:
rebased

PR #335, opened 2019-09-05
Support VXLAN skip for latency

michaelspedersen:
* Adds support for skipping the VXLAN header for latency streams.
  * Sets a flag, `m_vxlan_fs` when a port is configured for VXLAN
  * When received packets are handled, the flag is used to determine if the outer header should be skipped
* If the port is configured for VXLAN, but the received packets don't have the header, the parser will end up returning `FSTAT_PARSER_E_UNKNOWN_HDR` (6) and packets will not be processed

### DIFF
--- a/src/stx/common/trex_latency_counters.cpp
+++ b/src/stx/common/trex_latency_counters.cpp
@@ -141,6 +141,10 @@ RXLatency::handle_pkt(const rte_mbuf_t *m) {
     uint8_t tmp_buf[sizeof(struct flow_stat_payload_header)];
     CFlowStatParser parser(CFlowStatParser::FLOW_STAT_PARSER_MODE_SW);
     int ret = parser.parse(rte_pktmbuf_mtod(m, uint8_t *), m->pkt_len);
+    if ( CGlobalInfo::m_options.m_ip_cfg[0].get_vxlan_fs() ) {
+        uint16_t vxlan_skip = parser.get_vxlan_payload_offset(rte_pktmbuf_mtod(m, uint8_t *), m->pkt_len);
+        ret = parser.parse(rte_pktmbuf_mtod_offset(m, uint8_t *, vxlan_skip), m->pkt_len - vxlan_skip);
+    }
 
     if (m_rcv_all ||  (ret == 0)) {
         uint32_t ip_id = 0;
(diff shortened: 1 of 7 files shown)


src/stx/common/trex_latency_counters.cpp
```diff
@@ -141,6 +141,10 @@ RXLatency::handle_pkt(const rte_mbuf_t *m) {
     uint8_t tmp_buf[sizeof(struct flow_stat_payload_header)];
     CFlowStatParser parser(CFlowStatParser::FLOW_STAT_PARSER_MODE_SW);
     int ret = parser.parse(rte_pktmbuf_mtod(m, uint8_t *), m->pkt_len);
+    if ( CGlobalInfo::m_options.m_ip_cfg[0].get_vxlan_fs() ) {
+        uint16_t vxlan_skip = parser.get_vxlan_payload_offset(rte_pktmbuf_mtod(m, uint8_t *), m->pkt_len);
```
hhaim: ```get_vxlan_payload_offset ``` can't be used in this location. ```get_vxlan_payload_offset ``` is a CP function that throw exceptions for RPC. 
In this case you are processing packets in the DP and it is not possible to crash. 
You need to change the parser. parses function to skip the vxlan header (if exits and enable) exactly like what the ASIC does

michaelspedersen:
I deliberately didn't update the parser directly, as both that and the offset function is used elsewhere:
https://github.com/cisco-system-traffic-generator/trex-core/blob/ee8e07573948bcd96896c8dff4db3c90db6dc237/src/stx/stl/trex_stl_fs.cpp#L550-L575

hhaim:
@michaelhans Understood. 
The code is good now but could be better.
The parser function should be called **once** and skip the VxLan if enabled.
I don’t see an issue to fix the other places that uses the offset function (that should be private). Could you help improving the code?
The regression code will verify the fix

ahothan:
@hhaim could any further optimization be done in a later commit by someone who is more knowledgeable about the Trex code? I think we have stretched our limited knowledge of the code without risking breaking it with more changes or spending more time that we can afford. Perhaps this can be merged in this state?

hhaim:
@ahothan I’m out for a business trip. Will handle it next week

hhaim:
reabsed + minor changes

Issue #338, opened 2019-09-18
Some DP to CP burst messages might be lost


jsmoon:
Hi,
During the test with a large number of ASTF profiles, I've met a situation that CP did not know some profiles were stopped in DP side.
I think the reason is DP to CP buffer overflow. While DP can generate maximum 8192 messages, DP to CP ring buffer can keep up to 1024. If DP to CP buffer is full and following messages will be discarded in the current implementation.
```
void TrexAstfDpCore::report_finished(profile_id_t profile_id) {
    TrexDpToCpMsgBase *msg = new TrexDpCoreStopped(m_flow_gen->m_thread_id, profile_id);
    m_ring_to_cp->Enqueue((CGenNode *)msg);   <-- msg will be lost if m_ring_to_cp buffer is full.
}
```

I think a simple solution that enlarges the buffer size to 8192.
What do you think of this?

hhaim:
We can do that but it won't solve the issue it will just mitigate it. This type of message is **important** and we should be able to have a way to make sure the message was sent by either blocking or reschedule until the message is sent

jsmoon:
I think **rescheduling** is more reasonable.
It is not a desirable situation that the DP core is blocked by the CP processing delay.

hhaim:
I agree. But blocking could solve it too in a simpler way, as no new messages will be sent and once **one** message is processed by  CP it would be possible to add to this message

jsmoon:
There are 3 CMessagingManager instances used: CP <-> DP, CP <-> RX, RX <-> DP.
CP <-> DP messages are important and should not be lost. But, I'm not sure for others.
I can focus on CP <-> DP messages only. or I can modify for all.
What do you think of it?

PR #342, opened 2019-09-24
fix DP to CP burst msg lost issue

jsmoon:
This PR is related to #338 (Some DP to CP burst messages might be lost).

I modified CNodeRing to append SecureEnqueue() and Reschedule() methods.
SecureEnqueue() will be blocked until rte_ring buffer is available by default.
It also supports rescheduling when rte_ring buffer is full. CP<->DP messages use it.
To support rescheduling, I added Reschedule() method which is called at CP and DP scheduler.

### DIFF
--- a/src/bp_sim.cpp
+++ b/src/bp_sim.cpp
@@ -4157,6 +4157,8 @@ void   CFlowGenListPerThread::no_memory_error(){
 
 
 bool CFlowGenListPerThread::check_msgs_from_rx() {
+    m_ring_to_rx->Reschedule();
+
     if ( likely ( m_ring_from_rx->isEmpty() ) ) {
         return false;
     }
--- a/src/pal/linux/CRing.h
+++ b/src/pal/linux/CRing.h
@@ -89,6 +89,14 @@ public:
     int Dequeue(T * & obj){
         return (CRingSp::Dequeue(*((void **)&obj)));
     }
+
+    bool Reschedule() {
+        return true;
+    }
+
+    void SecureEnqueue(T *obj, bool use_resched=false) {
+        CRingSp::Enqueue((void*)obj);
+    }
 };
 
 
--- a/src/pal/linux_dpdk/CRing.h
+++ b/src/pal/linux_dpdk/CRing.h
@@ -27,6 +27,7 @@ limitations under the License.
 #include <string>
 #include <rte_config.h>
 #include <rte_ring.h>
+#include <queue>
 
 class CRingSp {
 public:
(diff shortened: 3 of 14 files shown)


hhaim:
The idea looks good. I have one concern. The ring is shared memory betwean two cores. 
Most of the time the high priority reschedule queue is empty. 
I would add a boolean flag in a separate cache line to elinirate false sharing for the CP that pool the stats.

jsmoon:
Since the m_resched_q is dedicated to only one thread which requests SecureEnqueue(), I didn't care about the false sharing case.
Am I missing something?

hhaim:
The queue is shared by two cores. One adding to the queue and one removing from the queue. 
The one that pool (which is more freq) should bail out quickly and not share a cache line in the common case. 

Adding 
```
m_resched_q.push(obj);

```

jsmoon:
> The queue is shared by two cores. One adding to the queue and one removing from the queue.

**m_resched_q is not shared by two cores.** It is dedicated to one core only.

As you know, a pair of rte_ring is used for communication between two cores. A rte_ring has a m_resched_q and it is used only by sending core.

For example, when a CP to DP rte_ring is full:
-> push msg to **m_resched_q** _by SecureEnque(msg, true) **in CP core**_
-> pop msg from **m_resched_q** _by Reschedule() **in CP core** TrexSTX::check_for_dp_messages()_
-> push msg to **rte_ring**
-> pop msg from **rte_ring** _in **DP core**_

hhaim:
The issue is that  ```Reschedule();``` function is  called from both sides and it should be called only from the  sender side. You can't call it from both sides 

So let's look into this code 

CP->RX (Rx side call ```Reschedule()``` while only Tx should ) 

```
uint32_t CRxCore::handle_msg_packets(void) {
    uint32_t pkts = 0;
    for (uint8_t thread_id=0; thread_id<m_tx_cores; thread_id++) {
        m_rx_dp->getRingCpToDp(thread_id)->Reschedule();  <<<<<<<<<<<<<<

        CNodeRing *r = m_rx_dp->getRingDpToCp(thread_id);
        pkts += handle_rx_one_queue(thread_id, r);
    }
```

If you call it only from sender side there is no issue of cache false sharing. I agree. 

Am I missing something?


the CP side pul a messgae

jsmoon:
CP->RX doesn't use reschedule. It uses the default blocking mode.

> m_rx_dp->getRingCpToDp(thread_id)->Reschedule();  <<<<<<<<<<<<<<

This is for RX->DP rte_ring, not for CP->DP.

hhaim:
The point is different 

This is DP->RX called from Rx core. 
My point that ```reschedule``` function should be called **only** from the sender (in this case DP)

jsmoon:
m_rx_dp->getRingCpToDp(thread_id) is for RX->DP rte_ring.
i.e. RX core is the sender of it.

hhaim:
Are we in agreement that ```Reschedule();``` API should be called **only** from sender side (the side that call Enqueue)? 

If Yes, the code I've sent has an issue: 

```
uint32_t CRxCore::handle_msg_packets(void) {
    uint32_t pkts = 0;
    for (uint8_t thread_id=0; thread_id<m_tx_cores; thread_id++) {
        m_rx_dp->getRingCpToDp(thread_id)->Reschedule();  <<<<<<<<<<<<<<

        CNodeRing *r = m_rx_dp->getRingDpToCp(thread_id);
        pkts += handle_rx_one_queue(thread_id, r);
    }

```

the function ```handle_rx_one_queue``` call ```Dequeue``` (rx side) and this is not allowed . 

```
uint32_t CRxAstfCore::handle_rx_one_queue(uint8_t thread_id, CNodeRing *r) {
    CGenNode * node;
    uint32_t got_pkts;
    CFlowGenListPerThread* lpt = get_platform_api().get_fl()->m_threads_info[thread_id];
    uint8_t port1, port2, port_pair_offset;
    lpt->get_port_ids(port1, port2);
    port_pair_offset = port1 & port2;

    for ( got_pkts=0; got_pkts<64; got_pkts++ ) { // read at most 64 packets
        if ( r->Dequeue(node)!=0 ){
            break;
        }
        assert(node);
```

If not please explain why it is OK

jsmoon:
```
uint32_t CRxCore::handle_msg_packets(void) {
    uint32_t pkts = 0;
    for (uint8_t thread_id=0; thread_id<m_tx_cores; thread_id++) {
        m_rx_dp->getRingCpToDp(thread_id)->Reschedule();  <<<<<<< RX->DP

        CNodeRing *r = m_rx_dp->getRingDpToCp(thread_id);
        pkts += handle_rx_one_queue(thread_id, r);        <<<<<<< DP->RX
    }
```
**They are different rte_ring queues.** ```handle_rx_one_queue``` does not handle RX->DP rte_ring which is called ```Reschedule()```.

In summary, ```CRxCore::handle_msg_packets()``` in RX core calls ```Reschedule()``` for RX->DP as a sender and calls ```Dequeue()``` for DP->RX as a receiver.

hhaim:
Understood. This is good.  Let me review the other places of the code

hhaim:
Looks good. Nice work.

ybrustin:
recheck

hhaim:
rebased

PR #343, opened 2019-09-27
fix crash in rx_handle_packet and remove default profile in DP

yonghwan007:
Crash was observed as the following backtrace, becasue server accessed a null pointer in the idle state of default profile in DP side.
So I have added a handling routine for active profiles in DP side to prevent crash.

Major changes are below.
1. removed the default profile in DP side.
 - Now it is possible to remove the default profile in DP side because 'Secure ASTF statistics update for multiple profiles'(https://github.com/cisco-system-traffic-generator/trex-core/commit/365973b365479d9c69573518af6f946a36177ab5) commit already covers a crash issue caused by the previous racing condition.
2. Handling the list of active profiles to avoid crash.
 - append an active profile into the list when load_tcp_profile.
 - remove an active profile from the list when unload_tcp_profile.
 - FALLBACK_PROFILE_CTX(ctx) will find the first profile context in the active profile list instead of the previousu profile list.

```
#0  std::vector<CTcpDataAssocTransHelp, std::allocator<CTcpDataAssocTransHelp> >::size (this=<optimized out>) at stl_vector.h:656
#1  CTcpDataAssocTranslation::get_server_info (params=<synthetic pointer>..., this=0x78)
     at ../../../../trex-core/src/astf/astf_db.cpp:105
#2  CAstfDbRO::get_server_info_by_port (this=this@entry=0x0, port=port@entry=67, stream=stream@entry=false) at ../../../../trex-core/src/astf/astf_db.cpp:1548
#3  0x00000000004f77ba in CFlowTable::rx_handle_packet_udp_no_flow (this=this@entry=0x2eddd38, ctx=ctx@entry=0x2eddc40, mbuf=mbuf@entry=0x1b6ce2e80, lpflow=lpflow@entry=0x0, parser=..., tuple=..., ftuple=..., hash=3430765316)
     at ../../../../trex-core/src/44bsd/flow_table.cpp:512
```

### DIFF
--- a/src/44bsd/tcp_subr.cpp
+++ b/src/44bsd/tcp_subr.cpp
@@ -717,8 +717,6 @@ bool CTcpPerThreadCtx::Create(uint32_t size,
         return(false);
     }
 
-    create_profile_ctx(0);  // create default profile
-
     return(true);
 }
 
--- a/src/44bsd/tcp_var.h
+++ b/src/44bsd/tcp_var.h
@@ -1038,6 +1038,7 @@ public:
     /* profile management */
 private:
     std::unordered_map<profile_id_t, CPerProfileCtx*> m_profiles;
+    std::unordered_map<profile_id_t, CPerProfileCtx*> m_active_profiles;
 
 public:
     bool is_profile_ctx(profile_id_t profile_id) { return m_profiles.find(profile_id) != m_profiles.end(); }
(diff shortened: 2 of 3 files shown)


yonghwan007:
@hhaim 
Do you have any comment?

hhaim:
Looks good.

hhaim:
rebased.

PR #346, opened 2019-10-01
Added stack name info to get_system_info

EgorBlagov:

- Added capability to get info about current stack from API

NOTE: Though stack is a global option for now, as I saw each port has it's own method to get stack name, hence I've added it under system info's port object
NOTE2: I didn't find a way how to start rpc_test.cpp tests (gtests), but edited it.


hhaim:
rebased.

Issue #336, opened 2019-09-09
trex-console --tui does not update state properly during the state is updated by another client.


jsmoon:
Hi,
When I start and stop a profile by ASTFClient API, I found that `trex-console --tui` does not update the state change properly.
I found the reason is that the state update ignored by epoch value mismatch.
```
    def _on_astf_state_chg(self, ctx_state, error, epoch):
        if ctx_state < 0 or ctx_state >= len(astf_states):
            raise TRexError('Unhandled ASTF state: %s' % ctx_state)

        if epoch is None or epoch != self.epoch:
            return

        ....
```
Since ASTFClient.start and ASTFClient.stop block parameter's default values are True, _transmit_async will be used internally and it requests "inc_epoch" to server everytime.
Other clients that already synched by previous epoch value won't update the new state by the above code.

I'd like to know what is the purpose of the **epoch** value and how I handle this issue properly.

hhaim:
There was an assumption that there is only one active profile. The epoch helps to identify that the **async** messages  match to the current active profile and only then it is safe to process it (else it is an old message)
If you change the epoc only when moving from zero profiles to 1 (CP) and not in each start it will solve the issue.

hhaim:
@jsmoon I've discussed this with @ybrustin 
The objective of the epoch is to verify that the async events are valid and new and not relates to the past.  

There should be two types of epochs

1. per profile 
2. global - when moving from 0->1 profile need to inc it

jsmoon:
@hhaim @ybrustin I understood the epoch value is used to figure out the received async event message is valid or not. But I'm still confusing how I could handle the value properly. So I have several questions from your comments.

1. At first, this issue is that the old connected client does not update the changed state by the async event messages **when the epoch value is changed**. What do you think of that? Should it be updated by the new one? or keep the old one?

2. How could I know the epoch value from an async event is not related to the past? I know that the client got the epoch value already by `sync` request when it began.
3. Who should have the responsibility for the epoch value handling? Currently, the server keeps the value but it is handled by the client's `inc_epoch` request.
4. `when moving from 0 -> 1 profile`: I'm not sure `0 -> 1` means. It could be the first profile loaded. It could be the first profile `start` request and more situation.
5. `per profile`: I cannot understand the reason why the `per profile` handling is needed. Could you explain it in detail?
If I understood correctly, the epoch value is used to keep the client's data consistency from the server. So, I cannot think the case that `per profile` value is needed.

[...]

PR #347, opened 2019-10-02
update epoch handling and reduce sync() calls in ASTFClient for multiple profiles usage

jsmoon:
This update is for performance improvement of _transmit_async() routine.
It calls self.sync() 5 times on each transaction. With the thousands of profiles, it takes a long time to get whole states of profiles. So I reduce them when it really needed.

This change includes a simple update for issue #336. Now ```_on_astf_state_chg()``` updates the state even though epoch value does not match.

Please check the code and give your feedback.


hhaim:
rebased.

PR #349, opened 2019-10-07
remove several CPU architecture dependencies

drchristensen:
Remove several CPU architecture specific features in the code
to allow additional CPU architectures to be added more easily.

- Move some external libraries to a <library>/<CPU arch> format
- Eliminated the use of /proc/cpuinfo for CPU topology detection
  since it isn't portable across CPUs (Intel specific)
- Added support for 20/25/50/100Gb NIC speeds
- Moved non-CPU specific i40e PMD source files into a new SrcGroup
- Switched DPDK lcore assignment from "-c <bitmask>" to "-l <string>"
  to support larger numbers of lcores in the future

Signed-off-by: David Christensen <drc@linux.vnet.ibm.com>


scripts/automation/trex_control_plane/server/outer_packages.py
```diff
@@ -13,7 +13,15 @@
...
 
-    cpu_vendor = 'arm' if os.uname()[4] == 'aarch64' else 'intel'
+    march = os.uname()[4]
+
+    if march == 'aarch64':
+        cpu_vendor = 'arm'
+    elif march == 'x86_64' or march == 'i386' or march == 'i686':
+        cpu_vendor = 'intel'
+    else:
+        assert(cpu_vendor, 'Unknown CPU architecture')
```
ybrustin: This assert will never raise Exception

scripts/external_libs/pyzmq-ctypes/zmq/bindings.py
```diff
@@ -11,7 +11,15 @@
...
 
-cpu_vendor = 'arm' if os.uname()[4] == 'aarch64' else 'intel'
+march = os.uname()[4]
+
+if march == 'aarch64':
+    cpu_vendor = 'arm'
+elif march == 'x86_64' or march == 'i386' or march == 'i686':
+    cpu_vendor = 'intel'
+else:
+    assert(cpu_vendor, 'Unknown CPU architecture')
```
ybrustin: This assert will never raise Exception

scripts/dpdk_setup_ports.py
```diff
@@ -1031,30 +1043,45 @@ def split_pci_key(self, pci_id):
...
-                if numa not in cpu_topology:
-                    cpu_topology[numa] = OrderedDict()
-                core = int(lcore_dict.get('core id', lcore_dict['processor']))
-                if core not in cpu_topology[numa]:
-                    cpu_topology[numa][core] = []
-                cpu_topology[numa][core].append(int(lcore_dict['processor']))
+
+        # Find the total number of CPUs (logical cores)
+        base_path = "/sys/devices/system/cpu"
+        fd = open("{}/kernel_max".format(base_path))
```
ybrustin: I would open file with "with open() ..." and remove "close"

scripts/dpdk_setup_ports.py
```diff
@@ -1031,30 +1043,45 @@ def split_pci_key(self, pci_id):
...
+        max_cpus = int(fd.read())
+        fd.close()
+
+        for cpu in xrange(max_cpus + 1):
```
ybrustin: max_cpus in CentOS is 5119.
Potentially it can be very large.
Are you sure you want to try to open all those files?
I would use glob with "cpu*" pattern and then check additionally in Python that filename is "cpu" with number afterwards.

scripts/dpdk_setup_ports.py
```diff
@@ -1031,30 +1043,45 @@ def split_pci_key(self, pci_id):
...
+
+            # Find the socket ID of the current CPU
+            try:
+                fd = open("{}/cpu{}/topology/physical_package_id".format(base_path, cpu))
```
ybrustin: I would open file with "with open() ..." and remove "close"

scripts/dpdk_setup_ports.py
```diff
@@ -1031,30 +1043,45 @@ def split_pci_key(self, pci_id):
...
+            except:
+                break
+            socket = int(fd.read())
+            fd.close()
+            if socket not in cpu_topology:
+                cpu_topology[socket] = OrderedDict()
+
+            # Find the core ID of the current CPU
+            try:
+                fd = open("{}/cpu{}/topology/core_id".format(base_path, cpu))
```
ybrustin: I would open file with "with open() ..." and remove "close"

scripts/dpdk_setup_ports.py
```diff
@@ -1031,30 +1043,45 @@ def split_pci_key(self, pci_id):
...
+            fd.close()
+            if core not in cpu_topology[socket]:
+                cpu_topology[socket][core] = []
+
+            # Capture the socket/core of the current CPU
+            cpu_topology[socket][core].append(cpu)
+
         if not cpu_topology:
-            raise DpdkSetup('Could not determine CPU topology from %s' % cpu_topology_file)
+            raise DpdkSetup('Could not determine CPU topology')
```
ybrustin: Please add "base_path" to exception so that user will know which path was checked.

scripts/dpdk_setup_ports.py
```diff
@@ -48,7 +52,7 @@ class ConfigCreator(object):
     mandatory_interface_fields = ['Slot_str', 'Device_str', 'NUMA']
     _2hex_re = '[\da-fA-F]{2}'
     mac_re = re.compile('^({0}:){{5}}{0}$'.format(_2hex_re))
-    MAC_LCORE_NUM = 63 # current bitmask is 64 bit
+    MAX_LCORE_NUM = 63 # current bitmask is 64 bit
```
hhaim: I think this should be per platform make file configuration. The code is cross compiled and in any x86 machine that we compile the code we would like 64 maximum CORES

hhaim:
@drchristensen in general very nice work. 
It shows how hard it is to add new platform support, which is no a good thing for this project. 
Please have a look into @ybrustin comments (he is the writer of dpdk_setup_port). 
We will need to do some manual tests as not all the capabilities of this scripts are tested.

scripts/automation/trex_control_plane/server/outer_packages.py
drchristensen: Changed the assert to "raise Exception()".  Unsure how to test the code.  If I remove the ppc64le support neither the unit test nor the functional tests documented in [how-to-build-trex](https://github.com/cisco-system-traffic-generator/trex-core/wiki#how-to-build-trex) fail.

scripts/external_libs/pyzmq-ctypes/zmq/bindings.py
drchristensen: Changed the assert to "raise Exception()".  Verified the code by commenting out the ppc64le support and running the functional regression test ("./run_regression --func") which generated the expected error.

scripts/dpdk_setup_ports.py
drchristensen: Done.

scripts/dpdk_setup_ports.py
drchristensen: Interesting.  On my PPC system the value is 2047 and my x86 system uses 8191 (both are running RHEL8 with kernel 4.18.0-80.el8).  I modified the code as suggested and the time dropped by 40x on the x86 system:
old run time =  0.041648149490356445
new run time =  0.001898050308227539

scripts/dpdk_setup_ports.py
drchristensen: Done.

scripts/dpdk_setup_ports.py
drchristensen: Done

scripts/dpdk_setup_ports.py
drchristensen: Done.

scripts/dpdk_setup_ports.py
drchristensen: Done.

drchristensen: I would note that I removed the bitmask used to configure logical cores for the DPDK EAL, switching from the "-c" option to the "-l" option which uses a string.  The maximum lcore limit still applies, but will be simpler to raise in the future.

hhaim:
> Thanks for the review. Assuming the code is finally accepted I'd like to hear your thoughts on testing and releasing for PPC in the future. The existing Travis configuration seems to use cloud storage to pass test requests/results, suggesting you're running CI internally on dedicated machines.

We are using *one* x86 server in the DMZ for unit-tests/regression and we are using a few setup internally. The Travis is only to sync the result. 

There are a few levels of tests:
1. Build 
2. Unit-test 
3. Regression  with real network setup (e.g. POWER + XL710)

1,2 can be done on a remote machine (travis) however the regression should use a real hardware and we don't have the resource to support it. 

current internal setups 
https://github.com/cisco-system-traffic-generator/trex-core/wiki/Regression-and-Setups

I would try to run full regression  on your Power setup to verify that there is no other issues 
https://github.com/cisco-system-traffic-generator/trex-core/wiki#running-regression

hhaim:
I've reviewed the code. Nice work. Will run full regression.

hhaim:
@drchristensen rebased. thanks for the effort.
Let me know what you think about this?

> We are using one x86 server in the DMZ for unit-tests/regression and we are using a few setup 
  internally. The Travis is only to sync the result.
  There are a few levels of tests:
   Build
   Unit-test
   Regression with real network setup (e.g. POWER + XL710)
   1,2 can be done on a remote machine (travis) however the regression should use a real hardware 
   and we don't have the resource to support it.
   current internal setups
  https://github.com/cisco-system-traffic-generator/trex-core/wiki/Regression-and-Setups
  I would try to run full regression on your Power setup to verify that there is no other issues
  https://github.com/cisco-system-traffic-generator/trex-core/wiki#running-regression

PR #355, opened 2019-10-22
fix server down issue when dynamic profile add/remove randomly

jsmoon:
This PR is related to #324.

Followings are implemented as discussed at #324:
- DP core now exits from its scheduler by TrexAstfDpScheduler CP to DP message explicitly.
CP now sends TrexAstfDpScheduler message to DP when it decided to make DP scheduler exited.
DP core reports its scheduler state by TrexDpCoreState DP to CP message when the state is changed.
CP will hold the following CP to DP messages until DP reports its state is IDLE.
- Per profile --nc option now works.
CP start_transmit() delivers the --nc option which is requested by the user to DP and it triggers the active flows to be closed.

Please check my code and give your feedback.

### DIFF
--- a/src/44bsd/flow_table.cpp
+++ b/src/44bsd/flow_table.cpp
@@ -507,11 +507,16 @@ bool CFlowTable::rx_handle_packet_udp_no_flow(CTcpPerThreadCtx * ctx,
 
     uint16_t dst_port = lpUDP->getDestPort();
 
+    if (!ctx->is_any_profile()) {
+        rte_pktmbuf_free(mbuf);
+        FT_INC_SCNT(m_err_no_template);
+        return(false);
+    }
     CPerProfileCtx * pctx = ctx->get_profile_by_server_port(dst_port,false);
     CAstfDbRO *tcp_data_ro = pctx->m_template_ro;
     CTcpServreInfo *server_info = tcp_data_ro->get_server_info_by_port(dst_port,false);
 
-    if (! server_info) {
+    if (!server_info || (!pctx->is_active() && pctx->get_nc())) {
         rte_pktmbuf_free(mbuf);
         FT_INC_SCNT(m_err_no_template);
         return(false);
@@ -615,6 +620,11 @@ bool CFlowTable::rx_handle_packet_tcp_no_flow(CTcpPerThreadCtx * ctx,
 
     uint16_t dst_port = lpTcp->getDestPort();
 
+    if (!ctx->is_any_profile()) {
+        rte_pktmbuf_free(mbuf);
+        FT_INC_SCNT(m_err_no_template);
+        return(false);
+    }
         /* not found in flowtable , we are generating the flows*/
     if ( m_client_side ){
         if ( ctx->tcp_blackhole ==0 ){
@@ -671,7 +681,7 @@ bool CFlowTable::rx_handle_packet_tcp_no_flow(CTcpPerThreadCtx * ctx,
     CAstfDbRO *tcp_data_ro = pctx->m_template_ro;
     CTcpServreInfo *server_info = tcp_data_ro->get_server_info_by_port(dst_port,true);
 
-    if (! server_info) {
+    if (!server_info || (!pctx->is_active() && pctx->get_nc())) {
         if (ctx->tcp_blackhole ==0 ){
           generate_rst_pkt(pctx,
                          dest_ip,
--- a/src/44bsd/tcp_var.h
+++ b/src/44bsd/tcp_var.h
@@ -717,11 +717,14 @@ private:
         return (tcp_pktmbuf_alloc(m_mbuf_socket,size));
     }
 
+    void keepalive_timer_start(bool init);
+
 public:
     CHTimerObj        m_keep_alive_timer; /* 32 bytes */ 
-    uint32_t          m_keepaive_ticks;
+    uint32_t          m_keepalive_ticks;
     uint8_t           m_mbuf_socket;    /* mbuf socket */
     uint8_t           m_keepalive;      /* 1- no-alive 0- alive */
+    uint32_t          m_remain_ticks;
     bool              m_closed;
 
 };
@@ -883,6 +886,7 @@ public:
 
 private:
     bool                m_stopped;
+    bool                m_nc_flow_close;
 
 public:
     ~CPerProfileCtx() {
@@ -894,6 +898,9 @@ public:
     void activate() { m_stopped = false; }
     void deactivate() { m_stopped = true; }
     bool is_active() { return m_stopped == false; }
+
+    void set_nc(bool nc) { m_nc_flow_close = nc; }
+    bool get_nc() { return m_nc_flow_close; }
 
     void on_flow_close() {
         if (m_flow_cnt == 0 && !is_active()) {
@@ -924,6 +931,8 @@ public:
     void handle_udp_timer(CUdpFlow * flow){
         if (flow->is_can_closed()) {
           m_ft.handle_close(this,flow,true);
+        } else if (!flow->m_pctx->is_active() && flow->m_pctx->get_nc()) {
+          m_ft.terminate_flow(this,flow,true);
         }
     }
 
@@ -931,10 +940,13 @@ public:
         if (flow->is_can_close()) {
             /* free the flow in case it was finished */
             m_ft.handle_close(this,flow,true);
-            return(RC_HTW_OK);
+        } else if (!flow->m_pctx->is_active() && flow->m_pctx->get_nc()) {
+            /* terminate the flow in case --nc specified */
+            m_ft.terminate_flow(this,flow,true);
         }else{
             return (m_timer_w.timer_start(&flow->m_timer,tcp_fast_tick_msec));
         }
+        return(RC_HTW_OK);
     }
 
 
@@ -1041,6 +1053,7 @@ private:
 
 public:
     bool is_profile_ctx(profile_id_t profile_id) { return m_profiles.find(profile_id) != m_profiles.end(); }
+    bool is_any_profile(void) { return (m_active_profiles.size() != 0); }
 #define FALLBACK_PROFILE_CTX(ctx)   ((ctx)->get_first_profile_ctx())
 #define DEFAULT_PROFILE_CTX(ctx)    ((ctx)->get_profile_ctx(0))
     CPerProfileCtx* get_profile_ctx(profile_id_t profile_id) {
@@ -1103,6 +1116,9 @@ public:
 
     void activate_profile_ctx(profile_id_t profile_id) { get_profile_ctx(profile_id)->activate(); }
     void deactivate_profile_ctx(profile_id_t profile_id) { get_profile_ctx(profile_id)->deactivate(); }
+
+    void set_profile_nc(profile_id_t profile_id, bool nc) { get_profile_ctx(profile_id)->set_nc(nc); }
+    bool get_profile_nc(profile_id_t profile_id) { return get_profile_ctx(profile_id)->get_nc(); }
 
     void set_profile_cb(profile_id_t profile_id, void *cb_data, on_stopped_cb_t cb) {
         CPerProfileCtx *pctx = get_profile_ctx(profile_id);
--- a/src/44bsd/tcp_var.h
+++ b/src/44bsd/tcp_var.h
@@ -899,7 +899,6 @@ public:
         if (m_flow_cnt == 0 && !is_active()) {
             if (m_on_stopped_cb) {
                 m_on_stopped_cb(m_cb_data, m_profile_id);
-                m_on_stopped_cb = nullptr;
             }
         }
     }
--- a/src/44bsd/udp.cpp
+++ b/src/44bsd/udp.cpp
@@ -67,10 +67,8 @@ void CUdpFlow::init(){
 
     m_closed =false;
     /* TBD need to fix this to tunebale */
-    m_keepaive_ticks  =  tw_time_msec_to_ticks(1000);
-
-    m_keepalive=1;
-    m_pctx->m_ctx->m_timer_w.timer_start(&m_keep_alive_timer,m_keepaive_ticks);
+    m_keepalive_ticks = tw_time_msec_to_ticks(1000);
+    keepalive_timer_start(true);
 }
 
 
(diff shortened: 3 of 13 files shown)


jsmoon:
Hi, I cannot access the FAILURE result http://81.218.86.50:8080/job/trex_build/193/.
I'd like to know which test case was not passed.

hhaim:
Try again, there was a ACL issue that was fixed. 
I'm looking into the code now
```test_astf_params (astf_tests.astf_resilience_test.ASTFResilience_Test) ... Build timed out (after 15 minutes). Marking the build as failed.
./run_regression: line 38: 26895 Terminated              $PYTHON trex_unit_test.py "${ARGS[@]}"```

**Comment by hhaim** [comment]
recheck

**Comment by hhaim** [comment]
**Code review**

@jsmoon, nice work. the ```--nc ``` is per profile now which is another nice thing to have

1. 

I think it is not necessary to add the  ```!flow->m_pctx->is_active() && flow->m_pctx->get_nc()``` for every Rx packet. The timer code is enough (~a few msec) to solve this case. The numbers of packets are way too high than the number of ticks

```
CFlowTable::process_tcp_packet(CTcpPerThreadCtx * ctx,

    if (flow->is_can_close()) {
        handle_close(ctx,flow,true);
    } else if (!flow->m_pctx->is_active() && flow->m_pctx->get_nc()) {
        terminate_flow(ctx,flow,true);
    }
}
```

2. ```--nc``` 
This is what I would expect and I don't see how the code solves this
1. ```--nc``` should be a field in the start command, per profile 
2. The flow termination should behave using this field -   (In case the spontaneous stop -- after duration or by explicit stop)

I don't see why ````m_opts->preview.setNoCleanFlowClose(true);```` this line is in this code and why?
Could you explain 
```
void TrexAstf::stop_dp_scheduler() {
    if (is_dp_core_state(TrexAstfDpCore::STATE_TRANSMITTING, true)) {
        m_opts->preview.setNoCleanFlowClose(true);
        TrexCpToDpMsgBase *msg = new TrexAstfDpScheduler(false);
        send_message_to_all_dp(msg);
        m_stopping_dp = true;
    }
}
```

**Comment by jsmoon** [comment]
1. I'll remove it from all packet-based processing.
2. `--nc` option from the client is delivered through `TrexAstfDpStart` message's `m_nc_flow_close` and it saved to each profile in DP. It can be checked by `CPerProfileCtx::get_nc()` method.
3. `CPerProfileCtx::m_nc_flow_close`  added, per profile.
4. `(!flow->m_pctx->is_active() && flow->m_pctx->get_nc())` is for check the condition.
m_nc_flow_close is true when explicit stop requested or start the profile with `--nc` option.
m_nc_flow_close is false only when start the profile without `--nc` option and duration is specified.

`m_opts->preview.setNoCleanFlowClose(true);` is used at `TrexAstfDpCore::start_scheduler()`
```
        if ( !m_flow_gen->is_terminated_by_master() && !go->preview.getNoCleanFlowClose() ) { // close gracefully
            m_flow_gen->m_node_gen.flush_file(-1, d_time_flow, true, m_flow_gen, old_offset);
        }
```
This code is remained for global 'no clean flow close' behavior.
`stop_dp_scheduler()` can be requested during some profiles are still stopping in DP. In this case, DP will exit scheduler immediately like old behavior.

**Comment by hhaim** [comment]
One more question 

We are gettig this error from time to time on random setup. I wonder if this commit could solve this too 
image

**Comment by hhaim** [comment]
> or global behavior, the above code might be useless. If it can be removed from `TrexAstfDpCore::start_scheduler()`, `m_opts->preview.setNoCleanFlowClose(true);` is not required.

Good point. In that case wouldn't be better to replace gas->preview. getNoCleanFlowClose()``` that works for one profile to ```is_all_the_working_profiles_are_nc()```. This function could be calculated in the DP.

**Comment by hhaim** [comment]
@jsmoon I've looked into the issue. it is easy to reconstruct. it seems that with small delay and sfr profile it is stuck forever 
see  how to reconstruct 
```test_astf_prof_profiles``` test stuck with the profile sfr_full_2k.py for duration 1. It might be related to the --nc 



```        
[csi-sceasr-b81]> /router/bin/python-2.7.4 trex_unit_test.py --astf --skip-clean --no-daemon -t test_astf_prof_profiles
        !!!     Running with modes: ['loopback'], not suitable tests will be skipped.   !!!
Done instantiating TRex scenario!

test_connectivity (astf_tests.astf_general_test.ASTFBasic_Test) ... 
Starting TRex... done. (0s)
Started TRex
Connecting.
Connected
Mapping ports.
Got ports mapping: {'map': {0: 3, 1: 2, 2: 1, 3: 0}, 'bi': [(0, 3), (1, 2)], 'unknown': []}
passed
test_astf_prof_profiles (astf_tests.astf_profile_test.ASTFProfile_Test) ...  running http_advanced.py
starting profile http_advanced.py for duration 1
 skipping http_by_l7_percent.py
 running http_eflow.py
starting profile http_eflow.py for duration 1
 running http_eflow2.py
starting profile http_eflow2.py for duration 1
 skipping http_eflow3.py
 running http_https.py
starting profile http_https.py for duration 1
 running http_manual_cli_tunable.py
starting profile http_manual_cli_tunable.py for duration 1
 running http_manual_commands.py
starting profile http_manual_commands.py for duration 1
 running http_manual_commands_connect_rst.py
starting profile http_manual_commands_connect_rst.py for duration 1
 running http_manual_commands_delay.py
starting profile http_manual_commands_delay.py for duration 1
 running http_manual_commands_delay_client.py
starting profile http_manual_commands_delay_client.py for duration 1
 running http_manual_commands_delay_client2.py
starting profile http_manual_commands_delay_client2.py for duration 1
 running http_manual_commands_delay_rand.py
starting profile http_manual_commands_delay_rand.py for duration 1
 running http_manual_commands_fin_ack.py
starting profile http_manual_commands_fin_ack.py for duration 1
 running http_manual_commands_loop.py
starting profile http_manual_commands_loop.py for duration 1
 running http_manual_commands_many_flows.py
starting profile http_manual_commands_many_flows.py for duration 1
 running http_manual_commands_pipeline.py
starting profile http_manual_commands_pipeline.py for duration 1
 running http_manual_commands_pipeline2.py
starting profile http_manual_commands_pipeline2.py for duration 1
 running http_manual_commands_pipeline3.py
starting profile http_manual_commands_pipeline3.py for duration 1
 running http_manual_commands_pipeline4.py
starting profile http_manual_commands_pipeline4.py for duration 1
 running http_manual_commands_rst.py
starting profile http_manual_commands_rst.py for duration 1
 running http_manual_tunable2.py
starting profile http_manual_tunable2.py for duration 1
 running http_manual_tunables.py
starting profile http_manual_tunables.py for duration 1
 running http_manual_tunables3.py
starting profile http_manual_tunables3.py for duration 1
 skipping http_manual_tunables4.py
 running http_manual_tunables5.py
starting profile http_manual_tunables5.py for duration 1
 running http_simple.py
starting profile http_simple.py for duration 1
 running http_simple2.py
starting profile http_simple2.py for duration 1
 skipping http_simple_cc.py
 skipping http_simple_cc1.py
 running http_simple_different_ip_range.py
starting profile http_simple_different_ip_range.py for duration 1
 running http_simple_limit.py
starting profile http_simple_limit.py for duration 1
 skipping http_simple_rss.py
 skipping sfr_full.py
 running udp_pcap.py
starting profile udp_pcap.py for duration 1
 skipping http_simple_src_mac.py
 skipping nginx_wget.py
 running param_ipv6.py
starting profile param_ipv6.py for duration 1
 running param_ipv6_rss.py
starting profile param_ipv6_rss.py for duration 1
 running param_mss_err.py
starting profile param_mss_err.py for duration 1
 skipping param_mss_err2.py
 running param_mss_initwnd.py
starting profile param_mss_initwnd.py for duration 1
 running http_simple_todd.py
starting profile http_simple_todd.py for duration 1
 running param_tcp_delay_ack.py
starting profile param_tcp_delay_ack.py for duration 1
 running param_tcp_keepalive.py
starting profile param_tcp_keepalive.py for duration 1
 running param_tcp_no_delay.py
starting profile param_tcp_no_delay.py for duration 1
 running param_tcp_no_timestamp.py
starting profile param_tcp_no_timestamp.py for duration 1
 running param_tcp_rxbufsize.py
starting profile param_tcp_rxbufsize.py for duration 1
 running param_tcp_rxbufsize_8k.py
starting profile param_tcp_rxbufsize_8k.py for duration 1
 running param_tcp_rxbufsize_pt.py
starting profile param_tcp_rxbufsize_pt.py for duration 1
 skipping sfr.py
 running citrix_reverse.py
starting profile citrix_reverse.py for duration 1
 running sfr_single.py
starting profile sfr_single.py for duration 1
 running tcp_param_change.py
starting profile tcp_param_change.py for duration 1
 running udp1.py
starting profile udp1.py for duration 1
 running udp2.py
starting profile udp2.py for duration 1
 running udp_mix.py
starting profile udp_mix.py for duration 1
 running http_reverse2.py
starting profile http_reverse2.py for duration 1
 skipping udp_topo.py
 skipping udp_topo_traffic.py
 skipping http_simple_split.py
 skipping http_simple_split_per_core.py
 running http_simple_2_pools.py
starting profile http_simple_2_pools.py for duration 1
 skipping wrapping_it_up_example.py
 running template_groups.py
starting profile template_groups.py for duration 1
 running ftp_multiple_flows_sim.py
starting profile ftp_multiple_flows_sim.py for duration 1
 running http_high_active_flows.py
starting profile http_high_active_flows.py for duration 1
delay 1
 running http_manual_cli_tunable_4k.py
starting profile http_manual_cli_tunable_4k.py for duration 1
 running param_ip_tos_ttl_4.py
starting profile param_ip_tos_ttl_4.py for duration 1
 running res.py
starting profile res.py for duration 1
 running param_ip_tos_ttl_global.py
starting profile param_ip_tos_ttl_global.py for duration 1
 running param_ip_tos_ttl_pt.py
starting profile param_ip_tos_ttl_pt.py for duration 1
 running nat.py
starting profile nat.py for duration 1
 running param_sch_accurate.py
starting profile param_sch_accurate.py for duration 1
 running param_sch_rampup.py
starting profile param_sch_rampup.py for duration 1
 running sfr_full_2k.py
starting profile sfr_full_2k.py for duration 1
ERROR
test_astf_prof_profiles_dynamic_profile (astf_tests.astf_profile_test.ASTFProfile_Test) ... ERROR
-----------------------
```

**Comment by jsmoon** [comment]
According to my examination, my last change regarding the removal of packet-based processing caused the following issues.
1. some profile flows won't be terminated. (flows remained in the server-side)
2. some packets are received after profile stopped and removed and DP can't assign a profile to process received packets.
 
I think we should find a way to solve these issues.

**Comment by hhaim** [comment]
The issue happens before https://github.com/cisco-system-traffic-generator/trex-core/pull/355/commits/28c29088702b1e290ff879c5dc0e8bf02ebee016

I don't understand how flow won't be terminated. The timer is alway on for all flows, so I don't see how the timer code does not terminate the flows (after a few msec). In any case terminating packes based on Rx packet is wrong (you might not get them)

**Comment by hhaim** [comment]
> Since there is only 'no clean flow close' for global behavior, the above code might be useless. If it can 
 be removed from TrexAstfDpCore::start_scheduler(), m_opts

More thinking about it, I think you are right. We should remove the code as it is not needed anymore (and remove the code that set the variable)  as it is per profile now and handled in the scheduler. This code was needed only in the old implementation when there was one profile and one configuration of --nc (true or false).
This code should be removed only for ASTF and *not* for STF

**Comment by jsmoon** [comment]
> I don't understand how flow won't be terminated.

According to my investigation, the UDP keepalive value is too large as following:
```
{\"commands\": [{\"msec\": 1400000, \"name\": \"keepalive\"}, {\"buf_index\": 337, \"name\": \"tx_msg\"}, 
```
Since 1400 seconds is about 23 minutes, 15 minutes timeout will be occurred.

**Comment by hhaim** [comment]
@jsmoon it explains the issue. It is easy to fix this. We can add a counter per UDP flow and set the timer to maximum of 100msec. So  for 1400,000 msec the counter would be 14,000
for TCP it works the same. This counter is a watchdog it does not need to be accurate

**Comment by jsmoon** [comment]
@hhaim After I implemented UDP keepalive timer handling, issue 1 seems to be solved.
But, issue 2 still remained.
> 2. some packets are received after profile stopped and DP can't assign a profile to process received packets.

There is a packet received after all flows are closed. This causes the server down.
```
assert: ../../../../trex-core/src/44bsd/tcp_var.h:1068 CPerProfileCtx* CTcpPerThreadCtx::get_first_profile_ctx() Assertion 'm_active_profiles.size() != 0' failed.
```
I thought that following ways to handle this issue:
1. reserve a static profile to process RX packet when there is no active profile.
2. no RX packets are guranteed when profile stop is done.
Which do you prefer?

**Comment by hhaim** [comment]
@jsmoon sure there is a need to handle this case. 
I don't have the backtrace but it is probably from the flow_table code that call this function (rx side). You should create a function ```is_any_profile()``` and it case there is no profile drop the packet ( and **count** it)

**Comment by jsmoon** [comment]
@hhaim, I'm in trouble with the reproduction of this test_astf_params test failure in my environment.
So, I'd rechecked this PR by rebasing the recent changes. But the result is the same.
Since I don't know the reason and I failed to the reproduction, I'll reset this changes and separate it to several features and push it again.

**Comment by hhaim** [comment]
The issue is that a profile with duration 28 sec stuck forever (more than 15 min) In this case it is HTTP simple 
I would try to reconstruct it with this test 

est_astf_prof_simple (astf_tests.astf_profile_test.ASTFProfile_Test) ... starting profile http_simple.py for duration 28

**Comment by jsmoon** [comment]
@hhaim I've just pushed with derived issue fix-ups and simple defense code.
But the FAILURE occurred at the STL regression test. 
Since the ASTF tests are all passed, I'll push the next `--nc` handling feature.

**Comment by hhaim** [comment]
> @hhaim I've just pushed with derived issue fix-ups and simple defense code.
> But the FAILURE occurred at the STL regression test.
> Since the ASTF tests are all passed, I'll push the next `--nc` handling feature.

I will check this in our regression, it might be a noise

**Comment by hhaim** [comment]
@jsmoon Is it ready to commit?  what was the issue?

**Comment by jsmoon** [comment]
@hhaim No, it just enables per profile '--nc' handling only. Single profile
'--nc' handling is the same as before. I need to push more commits to
implement the explicit EXIT_SCHEDULER.

Anyway, I have a question for you.
If the per profile '--nc' handling works well, I think there is no need to
exit DP scheduler even though all profiles are stopped.
Is there any other reason DP scheduler should be exit?

**Comment by hhaim** [comment]
@jsmoon could you elaborate. Do you want *not* to implement the "explicit EXIT_SCHEDULER"  
both modes should be supported (```--nc``` and *none* ```--nc```)  and you would need to support explicit EXIT_SCHEDULER for the later

**Comment by jsmoon** [comment]
I'll continue implementing the 'explicit EXIT_SCHEDULER' in this PR. As I
mentioned previous comment, to do this, I need to push more commits to
complete it.

I just thought about why the enter and exit is needed. (STL does not seem
to exit)
If my thought could be reasonable and possible, DP code would be more
simpler.

2019년 11월 4일 (월) 오후 9:12, Hanoh Haim <notifications@github.com>님이 작성:

**Comment by hhaim** [comment]
@jsmoon STL exits the scheduler and have a cold state to pull messages and sleep  

```
    if ( (m_state == STATE_IDLE) && (are_any_pending_cp_messages()) ) {
        idle_state_loop();
    }
```

What is your suggestion to simplify the code?

jsmoon:
@hhaim I think I misunderstood the STL scheduler's behavior.

My thought was that ASTF DP may reside in the scheduler(TrexAstfDpCore::start_scheduler()) until terminated by the master. In this case, the implicit and explicit EXIT_SCHEDULER is not needed.
It can remove the STATE_STOPPING state which causes this issue.
It can remove my implementations except `--nc` handling support in this PR.
CP does not need to know the DP scheduler state.
As we discussed earlier if sync_barrier() is not needed for ASTF, there is no need to check the first profile also. It can remove the STATE_STARTING state also.

Anyway, I'll continue implementing the "explicit EXIT_SCHEDULER."
My next step will be the removal of the implicit EXIT_SCHEDULER.
And the last step will be the addition of the "explicit EXIT_SCHEDULER."

jsmoon:
@hhaim I don't understand why the `test_traffic_vlan` failed.
I think my change is not related to the fail test cases.
Should I look into the reason?

Anyway, I found a change missed (removal of global `--nc` in DP scheduler) and update it. (e495d06)

hhaim:
@jsmoon is it ready?

jsmoon:
@hhaim I've done all my work for this PR.

hhaim:
@jsmoon great. What was the first issue ?

jsmoon:
@hhaim The issue was the server down when CP requests start to DP when DP is exiting scheduler.
This PR will prevent that situation.

hhaim:
@jsmoon sorry for not being clear 
I've asked what was the issue in regards to this comment (the regression failed with strange failure)

> @hhaim, I'm in trouble with the reproduction of this test_astf_params test failure in my environment.
So, I'd rechecked this PR by rebasing the recent changes. But the result is the same.
Since I don't know the reason and I failed to the reproduction, I'll reset this changes and separate it to several features and push it again.

jsmoon:
@hhaim I think my original change had some hole in the logic. I found some CP to DP requests can be silently ignored and cannot be recovered. I think it could be the reason for the issue.
I changed the condition to send TrexAstfDpScheduler message from when one of the DP core is transmitting to when all the DP cores are transmitting. Before it changed, DP core which is not transmitting state will ignore the request and the DP core cannot be IDLE. In this situation, CP will hold the profile start requests from the client.

hhaim:
@jsmoon nice works. It was stable for a few runs. 
rebased

PR #359, opened 2019-10-25
Crash issues from wrong API usage in ASTF mode #305 - Issue1

ejangle:
TRex server crash when user used same client/server IP in ASTF mode
Added defensive code to report error messages instead of crashing TRex server.


hhaim:
rebased.

Issue #360, opened 2019-10-27
ASTF/STL multi-profile is not stable 


hhaim:
The tests are going to be disabled for now 

```
(NOT test.result:PASS) AND (test.name:test_random_dynamic_add_remove)
(NOT test.result:PASS) AND (test.name:test_stress_start_stop_dynamic_profile)
(NOT test.result:PASS) AND (test.name:test_double_start_stop)
```

```
stateless_tests.stl_client_test.STLClient_Test.test_random_dynamic_add_remove
astf_tests.astf_resilience_test.ASTFResilience_Test.test_double_start_stop
```

```
id_input)
  File "/tmp/trex-scripts/automation/trex_control_plane/interactive/trex/astf/stats/traffic.py", line 180, in _get_stats_values
    raise TRexError(rc.err())
TRexError: Invalid profile : febanjrcak
```

Kibana results 
![image](https://user-images.githubusercontent.com/13034429/67632781-f59ab480-f8b0-11e9-9b50-db6a4dcdfd15.png)

@jsmoon could you have a look

hhaim:
Need to revert this commit:

https://github.com/cisco-system-traffic-generator/trex-c
core/commit/5f3c53b2057f14ef15ffdfc1e75eba0e1751c551

It  seems that ```test_double_start_stop``` stability issue is rather new with one of the last commits

hhaim:
Example of the failures 
```
Failure:
Traceback (most recent call last):
File "/usr/lib64/python3.6/unittest/case.py", line 60, in testPartExecutor
yield
File "/usr/lib64/python3.6/unittest/case.py", line 622, in run
testMethod()
File "/tmp/trex-scripts/automation/regression/stateless_tests/stl_client_test.py", line 1072, in test_random_dynamic_add_remove
test.run_test()
File "/tmp/trex-scripts/automation/regression/stateless_tests/stl_client_test.py", line 143, in run_test
assert dropped_packets < 2
AssertionError


Stdout:
1 new profile 0.profile_0 44 1
3 new profile 0.profile_1 34 2
5 new profile 0.profile_2 48 3
7 new profile 0.profile_3 12 4
9 new profile 0.profile_4 26 5
11 new profile 0.profile_5 9 6
13 new profile 0.profile_6 15 7
15 new profile 0.profile_7 41 8
17 new profile 0.profile_8 33 9
19 del profile 3 8
19 new profile 0.profile_9 5 9
20 del profile 5 8
[...]

Issue #361, opened 2019-10-28
astf_tests.astf_resilience_test.ASTFResilience_Test.test_stress_start_stop fails


hhaim:
@jsmoon yet another issue in ASTF
![image](https://user-images.githubusercontent.com/13034429/67659595-64394a00-f965-11e9-8b99-865b3dcf3a5b.png)

jsmoon:
I think this issue is the same as the #360 ASTF issue.
I hope #362 might solve this issue also.

hhaim:
@jsmoon I'm testing it, will let you know shortly

hhaim:
duplicate #360

PR #362, opened 2019-10-28
fix ASTF multi-profile is not stable of issue #360

jsmoon:
This PR is related to #360.
This change is for the ASTF issue only.
I've found my last suggestion was not enough to solve this. It needed more change to update the ASTF state properly.
I hope this change may solve the ASTF issue.

### DIFF
--- a/scripts/automation/regression/astf_tests/astf_resilience_test.py
+++ b/scripts/automation/regression/astf_tests/astf_resilience_test.py
@@ -136,7 +136,7 @@ class ASTFResilience_Test(CASTFGeneral_Test):
         self.astf_trex.reset()
 
 
-    def _test_double_start_stop(self):
+    def test_double_start_stop(self):
         print('')
         c = self.astf_trex
         c.load_profile(os.path.join(CTRexScenario.scripts_path, 'astf', 'udp1.py'))
@@ -147,7 +147,7 @@ class ASTFResilience_Test(CASTFGeneral_Test):
         c.stop()
 
 
-    def _test_double_start_stop_dynamic_profile(self):
+    def test_double_start_stop_dynamic_profile(self):
         print('')
         c = self.astf_trex
         random_profile_1 = self.randomString()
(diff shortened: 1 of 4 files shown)


hhaim:
@jsmoon I took the most problematic setup and run those tests in loop without success to reproduce it (before your commit)
I will merge this and re-enable the test (astf) for now

hhaim:
image

Almost all the setup failed with this commit. I will remove the test again

hhaim:
the server crash (I'm not sure it related to the fix your sent)

image

jsmoon:
I think it came from my change.
m_error should not be accessed after removed.
I'll fix it by tomorrow.

hhaim:
yes, It point to your commit. I've revert it and disable the tests again.

jsmoon:
@hhaim The last FAILURE occurred at the STL performance test. It doesn't seem like originated my code change.
Anyway, I found a possible hanging issue in the ASTFClient stop. My last commit will remove the possibility.

hhaim:
@jsmoon it does not seem to solve the issue. The test failed

hhaim:
@jsmoon BTW it is something new, I need to track the failures to specific commit

hhaim:
@jsmoon isn't it related to the same transient state solved in 
https://github.com/cisco-system-traffic-generator/trex-core/pull/359
Does it make sense to solve both issues in one PR?

jsmoon:
> isn't it related to the same transient state solved in
> #359 

What does `it` mean?
I think the test FAILURE reason is related to issue #360 and #361.
So, this PR can make the PR #359 test passed.

> Does it make sense to solve both issues in one PR?

If the reason is the same, I think it makes sense.

hhaim:
From Elastic Search analysis it point to this commit 

https://github.com/cisco-system-traffic-generator/trex-core/commit/f56ee31d41a9a306789334f0e3e7b0799bb56eed

I've created a new branch (with this commit blackout) to verify it, will let you know.

hhaim:
@jsmoon Ok, I've ran it a few times and it seems more stable (with Reset commit)
I will merge it.

PR #363, opened 2019-10-29
Trim trailing whitespace in source code

mateumann:
Coding style used throughout the codebase seems a little bit erratic.
Working on it with a modern IDE (or Vim with editorconfig's
`trim_trailing_whitespace = true`) results in a lot of modifications
just because of this whitespace trimming.  As I like using fairly
user-friendly tools (IDEs/editors) and do not like having commits
littered with just trimmed whitespaces here and there, I thought a
single clean-up commit might be useful.

This is a result of running `sed -E 's|\s+$||'` on `*.{c,cc,cpp,h,py}`
files in `src`, `linux_dpdk` and `linux` directories.


ybrustin:
I'm not sure if we will merge this, but even if we do, please do not change the dpdk files.
@hhaim Any other files? tcp stack etc.?

hhaim:
@mateumann we will not approve this.

PR #367, opened 2019-11-05
Cant not clear ASTF Traffic Statistics sometimes

Cyborn-Minjae-Lee:
I fixed for not cleared statistics of deleted profiles in ASTF mode
You can see TRex console log which statistics are not cleared although no profile in ASTF mode.

`
Profile States

         ID          |        State         
---------------------+---------------------
         _           |      STATE_IDLE      

trex>stats -a
Traffic stats summary.

                        |      client       |      server       |                         
------------------------+-------------------+-------------------+------------------------
         m_active_flows |                 0 |                 0 | active open flows       
            m_est_flows |                 0 |                 0 | active established flows 
           m_tx_bw_l7_r |             0 bps |             0 bps | tx L7 bw acked          
     m_tx_bw_l7_total_r |             0 bps |             0 bps | tx L7 bw total          
           m_rx_bw_l7_r |             0 bps |             0 bps | rx L7 bw acked          
             m_tx_pps_r |             0 pps |             0 pps | tx pps                  
             m_rx_pps_r |             0 pps |             0 pps | rx pps                  
             m_avg_size |               0 B |               0 B | average pkt size        
             m_tx_ratio |             100 % |          109.29 % | Tx acked/sent ratio     
       tcps_connattempt |               225 |                 0 | connections initiated   
           tcps_accepts |                 0 |               493 | connections accepted    
          tcps_connects |               128 |               128 | connections established 
            tcps_closed |               225 |               493 | conn. closed (includes drops) 
         tcps_segstimed |               481 |               877 | segs where we tried to get rtt 
        tcps_rttupdated |               384 |               512 | times we succeeded      
            tcps_delack |               128 |                 0 | delayed acks sent       
          tcps_sndtotal |              1005 |              3565 | total packets sent      
           tcps_sndpack |               128 |              2944 | data packets sent       
           tcps_sndbyte |             56025 |           4108032 | data bytes sent by application 
        tcps_sndbyte_ok |             31872 |           4108032 | data bytes sent by tcp  
           tcps_sndctrl |               493 |                 0 | control (SYN|FIN|RST) packets sent 
           tcps_sndacks |               384 |               621 | ack-only packets sent   
           tcps_rcvpack |              3072 |               256 | packets received in sequence 
           tcps_rcvbyte |           4108032 |             31872 | bytes received in sequence 
        tcps_rcvackpack |               256 |               512 | rcvd ack packets        
        tcps_rcvackbyte |             31872 |           4108032 | tx bytes acked by rcvd acks  
     tcps_rcvackbyte_of |               128 |               256 | tx bytes acked by rcvd acks - overflow acked 
           tcps_preddat |              2816 |                 0 | times hdr predict ok for data pkts  
             tcps_drops |                 0 |               365 | connections dropped     
         tcps_conndrops |                83 |                 0 | embryonic connections dropped 
    tcps_rexmttimeo_syn |               268 |                 0 | retransmit SYN timeouts 
         tcps_keeptimeo |                83 |                 0 | keepalive timeouts      
         tcps_keepdrops |                83 |                 0 | connections dropped in keepalive 
         tcps_testdrops |                17 |                 3 | connections dropped by user at timeout (no-close flag --nc) 
                err_cwf |               602 |                 0 | client pkt without flow 
         redirect_rx_ok |                 8 |                 8 | redirect to rx OK       

trex>
`

### DIFF
--- a/src/stt_cp.cpp
+++ b/src/stt_cp.cpp
@@ -484,11 +484,11 @@ void CSTTCp::Delete(bool last_time){
     
 }
 
-void CSTTCp::clear_counters(void) {
+void CSTTCp::clear_counters(bool epoch_increment) {
     for (auto &sts : m_sts) {
         sts.clear_counters();
     }
-    m_epoch++;
+    if (epoch_increment) m_epoch++;
 }
 
 void CSTTCp::Resize(uint16_t new_num_of_tg_ids) {
--- a/src/stt_cp.h
+++ b/src/stt_cp.h
@@ -99,7 +99,7 @@ public:
     void Update();
     void DumpTable();
     bool dump_json(std::string &json);
-    void clear_counters();
+    void clear_counters(bool epoch_increment=true);
     void Resize(uint16_t new_num_of_tg_ids);
     void DumpTGNames(Json::Value &result);
     void UpdateTGNames(const std::vector<std::string>& tg_names);
--- a/src/stx/astf/trex_astf.cpp
+++ b/src/stx/astf/trex_astf.cpp
@@ -353,12 +353,6 @@ void TrexAstf::stop_transmit(cp_profile_id_t profile_id) {
 void TrexAstf::profile_clear(cp_profile_id_t profile_id){
     TrexAstfPerProfile* pid = get_profile(profile_id);
     pid->profile_check_whitelist_states({STATE_IDLE, STATE_LOADED});
-
-    if (profile_id == DEFAULT_ASTF_PROFILE_ID) {
-        pid->profile_init();
-        return;
-    }
-
     pid->profile_change_state(STATE_DELETE);
 
     TrexCpToDpMsgBase *msg = new TrexAstfDeleteDB(pid->get_dp_profile_id());
@@ -560,8 +554,16 @@ void TrexAstfProfile::add_profile(cp_profile_id_t profile_id) {
 }
 
 bool TrexAstfProfile::delete_profile(cp_profile_id_t profile_id) {
-    if (is_valid_profile(profile_id)) {
-        auto profile = m_profile_list[profile_id];
+    if (!is_valid_profile(profile_id)) {
+        return false;
+    }
+
+    auto profile = m_profile_list[profile_id];
+    profile->clear_counters(false);
+    if (profile_id == DEFAULT_ASTF_PROFILE_ID) {
+        profile->profile_change_state(STATE_IDLE);
+        profile->profile_init();
+    } else {
         m_profile_id_map.erase(profile->get_dp_profile_id());
 
         auto state = profile->get_profile_state();
@@ -570,10 +572,10 @@ bool TrexAstfProfile::delete_profile(cp_profile_id_t profile_id) {
 
         delete m_profile_list.find(profile_id)->second;
         m_profile_list.erase(profile_id);
-        return true;
+
     }
 
-    return false;
+    return true;
 }
 
 bool TrexAstfProfile::is_valid_profile(cp_profile_id_t profile_id) {
--- a/src/stx/astf/trex_astf.h
+++ b/src/stx/astf/trex_astf.h
@@ -128,6 +128,11 @@ public:
     void    set_duration(double duration) { m_duration = duration; }
     void    set_factor(double mult) { m_factor = mult; }
 
+    /*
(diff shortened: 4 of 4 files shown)


hhaim:
@Cyborn-Minjae-Lee 

Your patch raises a question of what happen even **before** this patch in the following scenario 

1. There is the global counters (sum of all profiles) ```TrexRpcCmdAstfTotalCountersValues``` and per profile counters set 
2. There is an assumption that the global counters set is always going up or cleared (when we start the first profile)
3. in case one of the  profiled is **removed** (not stopped) in the middle of a test with more active profiles the global counters set will go backward and this is not nice  

example (time:profile-name,ops)
```
0  : t0  - start 
10: t2  - start 

20: t2  - stop
30: t3  - delete  << global counter go backward 

```

looking into this I suspect the counter go backward 
```
TrexRpcCmdAstfTotalCountersValues::_run(const Json::Value &params, Json::Value &result) {
```

In regard to this commit I think it does not make the state worse in the above regards

Cyborn-Minjae-Lee:
I dicussed with my colleagues.
We decide to hold this commit for a while.

hhaim:
@Cyborn-Minjae-Lee .working with diffs will solve this issue, meaning calculate the diff per profile (save the old counters) in case of removing a profile flush the diff from prev read

Cyborn-Minjae-Lee:
I'm checking the overall global statistics.
Until now, each profile statistics is cleared every start of profiles 
And also, global stats are cleared when each profiles are removed because stt is removed if profile instance is deleted.
example (time:profile-name,ops)
`0  : t1  - start 
10 : t2  - start 
20 : t1  - stop
30 : t1  - start
40 : t1  - stop
50 : t2  - stop
 ------------ 
60 : t1  - start
70 : t2  - start
80 : t1, t2  - remove (default profile '_" : idle state)
`
So,  the global statistics of time"30" will go down than time"20"(backward) in example.

Please let me know how manage global statistics.
1. keep statistics of deleted profiles?
2. clear only when first start of traffic in server overall(all profiles) or each profiles ? 
 or clear every start of profiles as now?

hhaim:
@Cyborn-Minjae-Lee 
We can't keep stats of deleted profiles, this will create a leakage.
My suggestion is to have 2 sets of counters per stt (per profile) let's call it   ```old``` ,```active``` 

1. When new profile is loaded both ```old```=0 and ```active```=0

2. update and remove will work like this (```global_c``` is the global counters set)

```
update:

  for profile in range(profiles):
    diff  = profile.stt.active - profile.stt.old
    profile.stt.old = profile.stt.active
    global_c += diff


profile_is_deleted(profile):

    diff  = profile.stt.active - profile.stt.old
    profile.stt.old = profile.stt.active
    global_c += diff

now it would be safe to remove the profile as the diff was flushed to the global counters
   
```

Make sense?

Cyborn-Minjae-Lee:
I understood your suggestion. I'd like to discuss not only this commit but also overall astf operation.

Do you agree with the current astf behavior that "each profile stat is cleared each time a profile starts" commented previously?
I think that you want to be cleared only when the ASTF state transit to TRASMIT state like as time "60". 
Additionally, I wonder how should be operating if  time "20" be  stop and remove ("20 : t1 - stop --remove")

hhaim:
@Cyborn-Minjae-Lee 


1. I agree. The profile counters API should start with zeros (read the active) after TRANSMIT state. We can't clear it when it to move to STOP  
2. 

> I think that you want to be cleared only when the ASTF state transit to TRANSMIT state like as time "60"."  

Correct

3. 
>I wonder how should be operating if time "20" be stop and remove ("20 : t1 - stop --remove")
 
from profile prospective API you will loose the teardown counters from global counter there is no issue

Cyborn-Minjae-Lee:
- Respond 1,2:
In fact, your responses to "1" and "2" are in conflict. "1" is start Transmit of each profile and "2" is start Transmit of ASTF(sum of all profile).
If you want "2", global statistics should be cleared on time "0" and "60".
On the other hand, in the current feature "1", global statistics are affected at the beginning of each profile.
- Respond 3:
in the current feature,  global statistics can go backward(down) with removing of each profile by stop_and_remove or reset .

hhaim:
rebased. Are we sync on hot to solve the ```reverse`` counters

PR #368, opened 2019-11-07
Stop timesync Tx stream when test (sim) finishes

mateumann:
Timesync is a very strange stream.  It is very much connected with a latency stream, but should work the other way round.  TRex instance that is a transmitting part (Tx) for the latency stream should be a receiving part (Rx) for the timesync stream.  And vice-versa. Tx for timesync is also called a master, while Rx is called a slave.

This patch does some refactoring of when and how the timesync Tx node is initialized.  It has been moved to `TrexStatelessDpCore::add_stream` where latency streams are created.  A timesync Tx node is created quite similarly to a corresponding latency node.  This allows for it to be similarly stopped and removed by `stop_traffic` method.

This patch is a follow up for what @hhaim mentioned (codilime/trex-core#7) as a *restart issue*.


mateumann:
Sorry, my mistake, I have created PR to the wrong repository.

src/main_dpdk.cpp
```diff
@@ -910,6 +922,23 @@ COLD_FUNC static int parse_options(int argc, char *argv[], bool first_time ) {
             case OPT_SLEEPY_SCHEDULER:
                 CGlobalInfo::m_options.m_is_sleepy_scheduler = true;
                 break;
+            case OPT_LATENCY_MEASUREMENT_METHOD:
```
hhaim: Could we  move this configuration to RPC command? CLI is overloaded and should be used only if it is something that we can't change dynamically

src/main_dpdk.cpp
```diff
@@ -1906,7 +1959,8 @@ HOT_FUNC int CCoreEthIFStateless::send_node_flow_stat(rte_mbuf *m, CGenNodeState
     lp_s->add_bytes(mi->pkt_len + 4); // We add 4 because of ethernet CRC
 
     if (hw_id >= MAX_FLOW_STATS) {
-        fsp_head->time_stamp = os_get_hr_tick_64();
+        fsp_head->time_stamp = CGlobalInfo::m_options.m_get_latency_timestamp();
```
hhaim: Could be wrapped under different name why it is in ```CGlobalInfo::m_options.```
CGlobalInfo::m_options.m_get_latency_timestamp();

src/main_dpdk.cpp
```diff
@@ -1955,7 +2009,11 @@ CCoreEthIFStateless::send_node_packet(CGenNodeStateless      *node_sl,
         }
         return send_node_flow_stat(m, node_sl, lp_port, lp_stats, (node_sl->get_cache_mbuf()) ? true : false);
     } else {
-        return send_pkt(lp_port, m, lp_stats);
+        if(unlikely(node_sl->should_send_immediately())) {
```
hhaim: flow_stats is always immediately

src/stx/stl/trex_stl_dp_core.cpp
```diff
@@ -807,16 +807,28 @@ bool PerPortProfile::stop_traffic(uint8_t  port_id,
...
         CGenNodeStateless * node =dp_stream.m_node;
         assert(node->get_port_id() == port_id);
-        if ( node->get_state() == CGenNodeStateless::ss_ACTIVE) {
-            node->mark_for_free();
-            m_active_streams--;
-            dp_stream.DeleteOnlyStream();
-
-        }else{
-            dp_stream.Delete(m_port->m_core);
+        if (node->m_type != CGenNode::TIMESYNC) {
```
hhaim: Why? it seems it is the same code why CGenNodeTimesync needs a diff treatment

src/stx/stl/trex_stl_stream_node.h
```diff
@@ -703,5 +703,85 @@ friend class TrexStatelessDpPerPort;
 
 static_assert(sizeof(CGenNodePCAP) == sizeof(CGenNode), "sizeof(CGenNodePCAP) != sizeof(CGenNode)" );
 
-#endif /* __TREX_STL_STREAM_NODE_H__ */
+/* this is a event for time synchronization. */
+struct CGenNodeTimesync : public CGenNodeBase {
```
hhaim: This node is called from all the DP cores. My understanding that you should have only **one** for that.

mateumann: For TRex instances using high res ticks or system RTC one CGenNodeTimesync will be enough.  But I am working on a potentially different case.  If I get it correctly, a TRex core kind of represents DPDK port (physical port on a NIC).  If this is the case, then I would stick to this proposal.  From my understanding, some NICs have RTCs attached to ports which may be used in PTP time synchronization.

src/stx/stl/trex_stl_dp_core.cpp
mateumann: Sure, I am working on a refactor.

src/main_dpdk.cpp
mateumann: Thanks

src/main_dpdk.cpp
mateumann: I do assume latency measurement or time synchronization will not be changed dynamically.

src/main_dpdk.cpp
MehTheHedgehog: Hi @hhaim, thanks you for review!
As far as I understand, this part of code is referring to any other nodes as check of ->is_stat_needed() is above it (1950/2004). As normally if node don't has stat_needed it will go buffered path.
This part is trying to allow normal node with additional flag to be send immediately without buffering.

PR #372, opened 2019-11-11
Improved ASTF reset and clear routine

JunsubHan4rmEricsson:
Several issues have been reported by our customers
1) When using **"stats -a"** command, negative statistics values were seen 
2) clear -a does not clear all profiles

In order to fix the reported errors, following changes have been made accordingly, 
**1. "clear_stats" to be applied to ALL_PROFILE_ID by default (no pid parameter given)**
-> by using "clear -a" command without a profile param, user can remove all profiles
**2. "reference stats values (_ref)" in CAstfTrafficStats(traffic.py) to manage all reference values in dictionary by PID**
**3. traffic_stats.reset() and latency_stats.reset() when reset() used in ASTFClient (trex_astf_client.py)**

### DIFF
--- a/scripts/automation/trex_control_plane/interactive/trex/astf/stats/traffic.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/astf/stats/traffic.py
@@ -13,32 +13,47 @@ class CAstfTrafficStats(object):
 
 
     def reset(self):
-        self.is_init = False
         self._ref = {}
         self.tg_names_dict = {}
 
+        self._ref_global = {}
+        self._epoch_global = 0
+        self.is_init = False
 
-    def _init_desc_and_ref(self, pid_input = DEFAULT_PROFILE_ID):
-        if self.is_init:
-            return
+
+    def _init_desc_and_ref(self, pid_input = DEFAULT_PROFILE_ID, is_sum = False):
+        if is_sum:
+            if self._ref_global:
+                return
+        else:
+            if pid_input in self._ref.keys():
+                return
+
         params = {'profile_id': pid_input}
         rc = self.rpc.transmit('get_counter_desc', params = params)
-        
         if not rc:
             raise TRexError(rc.err())
 
         data = rc.data()['data']
-        self._desc = [0] * len(data)
-        self._err_desc = {}
-        for section in self.sections:
-            self._ref[section] = [0] * len(data)
-        self._max_desc_name_len = 0
-        for item in data:
-            self._desc[item['id']] = item
-            self._max_desc_name_len = max(self._max_desc_name_len, len(item['name']))
-            if item['info'] == 'error':
-                self._err_desc[item['name']] = item
-        self.is_init = True
+
+        if is_sum:
+            for section in self.sections:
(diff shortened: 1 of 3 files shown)


JunsubHan4rmEricsson:
@hhaim 
Could you please check the PR for me?

hhaim:
@JunsubHan4rmEricsson Is this the same issue we talked about last week (counters with multi ASTF profile go backward)?

If yes, I don't understand how this PR fix this, the fix should be in the server not in Client side

JunsubHan4rmEricsson:
@hhaim 
I've never been involved in the counters discussion.
I started this thread due to my customer request about negative values being shown when **stats -a** is used. 
As I have explained previously, that was because _ref parameter in traffic.py was not kept in track for each profile id.
And I've changed it so that each profile id can have its own _ref values for statistics comparison.

hhaim:
@JunsubHan4rmEricsson you are right !
see this 
https://github.com/cisco-system-traffic-generator/trex-core/pull/367 (the real issue wasn't solved yet)

This is the right solution to this problem, it is interesting that your raised this issue as we just discussed it last week

JunsubHan4rmEricsson:
@hhaim 
Could you be specific about "**what is the right solution**" to "**which problem**"
Are you saying that my PR is appropriate to the **negative values shown when stats -a is used?**

My issue is not derived from #367 
So I'm a little bit confused about your point

My PR is different from **the counters with multi ASTF profile go backward** issue.

hhaim:
The issue is this: 
With multiple profiles the global tcp statistic counters could go backward because when one profile is removed in the middle. 

Isn't it the same issue you are trying to solve

JunsubHan4rmEricsson:
@hhaim I do understand.
The issue you're talking about is different from the issue that I PRed for. 
That issue is in discussion and being planned within the team.

hhaim:
Understood. Let me see if I understand 

1. Clear stats - clear only default profile or some specific profile but not all, am I right?
2. Why ```IsSum``` arg is added? Do we want to clear each profile **and** global?

JunsubHan4rmEricsson:
1. clear stats -> when is_sum parameter is used, clear all profiles, when it is not, clear only specific profile.
2. is_sum -> I used this profile to differentiate between total_counters and specific_counters for pid. Default profile id ("_") is used for both counters, so I needed a parameter to differentiate between the two situation. 
Furthermore, the most important change made is that the _ref parameter is now being tracked for each profile. 
Before, _ref parameter was being overlapped every time "clear_stats" was being called and not being tracked for each profile.

hhaim:
Ok, make sense.  The same idea should be implemented in the server for the backward counters

JunsubHan4rmEricsson:
@hhaim yes it will be implemented in a separate thread in the near future

JunsubHan4rmEricsson:
@hhaim let me know if any changes need to be made for this PR

JunsubHan4rmEricsson:
@hhaim let me know if any changes need to be made for this PR

hhaim:
rebased

PR #304, opened 2019-07-15
Added a new test ASTF regression test cases for dynamic profile feature

JunsubHan4rmEricsson:
@hhaim 
This is derived from #300 and #299 
I have fixed the problems you've mentioned

PR #373, opened 2019-11-12
added  back the test case

JunsubHan4rmEricsson:
Hi @hhaim 
This test case should be added back since #355 has been solved. 
It passes regression test from our side. 
Please check it for me.


hhaim:
@JunsubHan4rmEricsson, It seems that I have missed this.

JunsubHan4rmEricsson:
@hhaim we excluded this test case because #355 was not implemented before (which solves issues that existed in #304 with the test case)

JunsubHan4rmEricsson:
@hhaim let me know if any changes need to be made for this PR

hhaim:
rebased

PR #377, opened 2019-11-14
Increase the number of maximum interfaces

jhexan:
We need to run TREX on a setup with more than 24 interfaces. This change increases the maximum number of allowed interfaces to 32.

### DIFF
--- a/scripts/dpdk_setup_ports.py
+++ b/scripts/dpdk_setup_ports.py
@@ -539,8 +539,8 @@ Other network devices
 
         if_list= if_list_remove_sub_if(self.m_cfg_dict[0]['interfaces']);
         l=len(if_list);
-        if l > 24:
-            raise DpdkSetup("Configuration file %s should include interfaces field with maximum 24 elements, got: %s." % (fcfg,l))
+        if l > 32:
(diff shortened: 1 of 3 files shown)


kitguyy:
This is also related to https://github.com/cisco-system-traffic-generator/trex-core/commit/6682313151fd92539ed68a12ed953170867c36f0#diff-5d31b06d242467092df1601dcdc44bd1

hhaim:
@kitguyy could you elaborate the use-case? In which setup, it is important to have more than 24 ports?

kitguyy:
Our device under test has 28 10G interfaces. We could test with multiple TREX boxes, but we’ve had issues with virtualizing using 10G intel. We’re ordering 10G mellanox interfaces but in the meantime we have a physical 28 port traffic generator that we want to use with TREX.

sent from my iphone

kitguyy:
Can we get this merged? Currently we have a custom build we're relying on.
Thanks

bdollma:
I think this is solved in 8c6706d. However, it is worth noting that there is a small diff between the commit and this PR, which changes the 
```c
#define MAX_DPDK_ARGS (30 + TREX_MAX_PORTS * 2)
```
while the commit doesn't.

hhaim:
@jhexan I missed this. We have integrated it is a few commits. Thanks for verifying it ..

PR #379, opened 2019-11-26
Fix invalid escape sequences

fperrin:
Hi,

When running under Python3, we get a number of `DeprecationWarning: invalid escape sequence` warnings. This pull-request fixes these. I tried following the convention of using the raw-string `r""` prefix for regexp strings, or extra backslashes for other strings.

To avoid more of these creeping in, running `pycodestyle --select=W6 .` from the top of the code repository will flag this error.


doc/TRexDataAnalysis.py
```diff
@@ -108,7 +108,8 @@ def analyze_latest_test_results(self):
             test_dates.append(test.latest_result_date)
             test_latest_results.append(test.latest_result)
         self.latest_test_results = pd.DataFrame(
-            {'Date': test_dates, 'Test Name': test_names, 'MPPS\Core (Norm)': test_latest_results},
+            {'Date': test_dates, 'Test Name': test_names,
+             'MPPS/Core (Norm)': test_latest_results},
```
fperrin: I changed this to a forward slash because other locations in that file use a forward slash too (see eg. l150).

fperrin:
Since there is already a self-test for syntax; I could add a test running pycodestyle, assuming an extra dependency on that package is okay.

ybrustin:
Are you running Python with warnings switch in order to produce described results?

fperrin:
Yes. More exactly I use pytest and Python3 to drive TRex, and I thing pytest forces recompilation of Python module files every time it runs.

PR #382, opened 2019-12-04
check_limit modified to cover generate_flow fail case

jsmoon:
Hi, I changed check_limit action to cover generate_flow fail case.
Currently, check_limit reduces the limit number immediately. But when generate_flow returns without a new flow, generate_flow will not retry the failed flow at next time.
So, it results less flows than the specified limit. This change make it up to the specified limit.


hhaim:
rebased

PR #383, opened 2019-12-04
wait_on_traffic for a profile in ASTF mode.

ejaecki:
The major change of this commit is that user wants to use the wait_on_traffic() api of ASTF to wait a profile, related changes are on wait_no_traffic() and wait_for_profile_state(..).
 And additional change is to reduce the duplicated lines by use the has_elapsed() function added in PassiveTimer class.
 Please review and comments
.


hhaim:
rebased.

PR #386, opened 2019-12-05
ASTF multi: fix calculating client port by using profile id

Kangphil:
When client create profile with same IP address, ASTF connection does not established because of the source port conflict in client side.  This issue is caused because client source port is generated by randomly, but, random sequence is same per profile.  And, in case of single profile include multiple ip_gen object start with same client IP address, this issue can happen by same reason.
I modified port calculation function use  porfile id and base_client_id to fix this issue, and I think this commit can almost reduce port conflicts.


hhaim:
see my answer here:
https://github.com/cisco-system-traffic-generator/trex-core/issues/376

PR #388, opened 2019-12-06
fixing import path in some example script

gamo-k:
Some STL example scripts have broken import path and raise ImportError on execution. I have applied the tweak used by other examples and now they can be executed.


hhaim:
Thanks

hhaim:
rebased

PR #396, opened 2019-12-20
recv(bytes,clear=True) function issue in ASTFProgram

ejangle:
When recv() with clear=true is used again in ASTF Program,
There is issue which accumulate recv bytes


hhaim:
rebased

PR #397, opened 2019-12-20
Register the sync method of ASTFClient to the client API

ejaecki:
As a user, I want to get the astf_profile_state of ASTFClient periodically by use a client API, the class method of ASTFClient which is sync(), it is used to update the astf_profile_state but it doesn't have a return value and it's not a client API.
So, I modified the sync() method to the client API with the return value which is astf_profile_state.
 Could you please review and comment?


ejaecki:
@hhaim @Cyborn-Minjae-Lee @egwakim @ejangle @JaeyongYu @jsmoon @JunsubHan4rmEricsson @taewoonyun @yonghwan007 @youngjun1 @Kangphil

hhaim:
rebased

PR #403, opened 2020-01-03
Enable server port sharing among templates in multiple profiles.

jsmoon:
Hi, this PR is related to issue #381 and will enable server port sharing by server IPs.
These commits cover the server implementation only, client implementation will be next.

I prepared the L7 map concept for the UDP case as discussed earlier. This can be enabled by the following format in "server_template".
```
    "assoc": [
        {
            "port": 53,
            "l7_map": {
                "offset": [ 0, 1, 2, 3 ], <-- mandatory to assemble value from payload.
                "mask": [ 127 ],          <-- optional default value is 255 for missing
                                              (== [ 127, 255, 255, 255 ])
                "value": [ 0, 42, 0, 0 ]  <-- optional specified value for the server only mode.
                                              w/o this, server will assemble it from buffer.
            }
        }
    ]
```

Next commits are planned for full functionality.
- implement L7 map concept for the TCP case (TCP payload handling will be added)
- implement ASTFClient support for the L7 map concept.
- add regression test for this feature. (can be separated PR)

Please check the changes and give your feedback.
Thank you.


jsmoon:
Hi Hanoh,
I've checked the failed test cases. At most of them, 'err_no_template' counters are observed.
Since my change limits the server IP is limited to the destination IP pools, other IPs would be dropped to the 'err_no_template' counters.
In my test environment, it also passed without no 'err_no_template' counters.
Is it possible for your test environment?

jsmoon:
Now, server IP is limited when the range is specified explicitly.
```
    "assoc": [
        {
            "port": 53,
            "ip_end": "48.0.255.255",
            "ip_start": "48.0.0.0",
            "l7_map": {
                "offset": [ 0, 1, 2, 3 ], <-- mandatory to assemble value from payload.
                "mask": [ 127 ],          <-- optional default value is 255 for missing
                                              (== [ 127, 255, 255, 255 ])
                "value": [ 0, 42, 0, 0 ]  <-- optional specified value for the server only mode.
                                              w/o this, server will assemble it from buffer.
            }
        }
    ]
```

jsmoon:
Hi, I've finished the planned feature.
Please check the changes and give your comments now.

egwakim:
Hi @hhaim,
It's implementation of #381 , Can review the code?

egwakim:
Hi @hhaim 
Can you review this change?

hhaim:
1.  ***Optimized the common case***

```
CServerTemplateInfo* CServerPortInfo::get_template_info(uint32_t ip, uint8_t* data, uint16_t len) {
    CServerTemplateInfo* temp = get_template_info_by_payload(ip, data, len);
    return temp ? temp: get_template_info_by_ip(ip);
}
```

Can we make this function faster by making a offline calculation that there is no ip range and no payload data, in this case just save the pointer to the template (like it was in the old version)

With the new code, there are two lookups, payloads  and range of IPs. These lookups is done once in a flow but we could make it faster for the common case. 

2. **Pcap utility**
 I would add a utility in the ```ASTFCapInfo``` that in case of a port conflict ( 3 templates on the same port, 80 in this case)  try to calculate the l7 data/mask to differentiate the association by the L7 

So in this example 

````
        profile = ASTFProfile(default_ip_gen=ip_gen, cap_list=[
            ASTFCapInfo(file="../avl/delay_10_http_get_0.pcap", cps=102.0,port=8080),
            ASTFCapInfo(file="../avl/delay_10_http_post_0.pcap", cps=102.0,port=8081),
            ASTFCapInfo(file="../avl/delay_10_https_0.pcap", cps=33.0),
```

This script will do the job and calculate the json automatically 

````
        profile = ASTFProfile(default_ip_gen=ip_gen, cap_list=[
            ASTFCapInfo(file="../avl/delay_10_http_get_0.pcap", cps=102.0),
            ASTFCapInfo(file="../avl/delay_10_http_post_0.pcap", cps=102.0),
            ASTFCapInfo(file="../avl/delay_10_https_0.pcap", cps=33.0),
```

jsmoon:
@Hanoh, Thank you for your comments.
1. Optimized the common case
I added a pointer to the template for the all IP range (default IP range for the compatibility).
2. Pcap utility
According to my short investigation, I think it will take several days.
So, I'd like to implement it with another PR.

hhaim:
1. **UDP** I mean faster than that ..

Instead of this code:


```
uint8_t *l7_data = pkt + ftuple.m_l7_offset;
uint16_t l7_len = ftuple.m_l7_total_len;

CServerTemplateInfo *temp = ctx->get_template_info(dst_port,false,dest_ip, l7_data,l7_len);
CPerProfileCtx *pctx = temp ? temp->get_profile_ctx(): nullptr;
CTcpServerInfo *server_info = temp ? temp->get_server_info(): nullptr;


```


Maybe this:

```
// lookup only by port, in  case there is one template for one port (common case) we have a match. You need to maintain map[port]:*CServerTemplateInfo
CServerTemplateInfo *temp = ctx->get_template_info(dst_port);

if (!temp) { // try with more info 
    uint8_t *l7_data = pkt + ftuple.m_l7_offset;
    uint16_t l7_len = ftuple.m_l7_total_len;
    CServerTemplateInfo *temp = ctx->get_template_info_ext(dst_port,false,dest_ip, l7_data,l7_len);
}

CPerProfileCtx *pctx = temp ? temp->get_profile_ctx(): nullptr;
CTcpServerInfo *server_info = temp ? temp->get_server_info(): nullptr;
```
This way you are offline build a faster map ```map[port]template``` similar to what it was before 

**TCP**:

I don't understand how the code works, maybe I miss something. There is no payload in the first SYN.  In the SYN you need to look only by port/ip-range (**no L7 data**) 

In case you find that this port has more options that need to be resolved by the L7 data you need to **defer** that decision and to associate the template only after you get the L7 data. 
This is a bit complex as you need to set a "dummy template emulation"  (you must have recv in all the programs on the server side) after getting the first data change it to the real template 
for example 
```
1. prog_s.recv(100) -> run the loopup by data 
2. change the emulation server data 
```

important: You need to validate that all the server programs that has a conflict in ports starts with ```recv```  instruction else this defer decision won't work  (Client side)

**The code**
``
uint8_t *l7_data = pkt + ftuple.m_l7_offset;
uint16_t l7_len = ftuple.m_l7_total_len;

CServerTemplateInfo *temp = ctx->get_template_info(dst_port,true,dest_ip, **l7_data,l7_len**);
CPerProfileCtx *pctx = temp ? temp->get_profile_ctx(): nullptr;
CTcpServerInfo *server_info = temp ? temp->get_server_info():
nullptr;
``


Pcap utility: agreed, let's defer it

jsmoon:
**UDP**
In my design, the payload should be checked first when the port shared templates specified the L7 map. Your suggestion may not be worked to the templates when the IP only template exists (as old usage) on the same port.

**TCP**
I deferred the decision using by **interim template info** which contains interim profile context for statistics updates and template_rw reference. The interim profile context is dedicated to the flow only. And I made the server program starts after a payload received in the flow. (see update_tcp_flow method)
I know that it can cause a performance issue when the port is shared by L7 map specified templates. Although there will be possible to improve performance on the case, I'd like to do it after this feature is stabilized.

> important: You need to validate that all the server programs that has a conflict in ports starts with recv instruction else this defer decision won't work (Client side)

Since the server program starts after the template is decided, I thought that there is no need to check it.

> CServerTemplateInfo *temp = ctx->get_template_info(dst_port,true,dest_ip, **l7_data,l7_len**);

OK, I can remove l7_data and l7_len for the SYN packet received. I'll update it by tomorrow.

hhaim:
@jsmoon 
**UDP**  I would not want to incur a performance penalty in the common case that there is no port conflict (one template for one port). In this case there is no need for L7. 
I agree that L7 should be checked first when you didn't find it in the first map (port: template)  

So this code should work and it will work faster. 
```
// lookup only by port, in  case there is one template for one port (common case) we have a match. You need to maintain map[port]:*CServerTemplateInfo
CServerTemplateInfo *temp = ctx->get_template_info(dst_port);

if (!temp) { // try with more info 
    uint8_t *l7_data = pkt + ftuple.m_l7_offset;
    uint16_t l7_len = ftuple.m_l7_total_len;
    CServerTemplateInfo *temp = ctx->get_template_info_ext(dst_port,false,dest_ip, l7_data,l7_len);
}

CPerProfileCtx *pctx = temp ? temp->get_profile_ctx(): nullptr;
CTcpServerInfo *server_info = temp ? temp->get_server_info(): nullptr;
```

**TCP**
Understood.  The following C/S program with port conflict will not work with your code because the server sends data and not waits for data. You need to verify that the first instruction in case of conflict has  a ```recv``` instruction 

```
        prog_c = ASTFProgram()
        prog_c.recv(len(http_req))

        prog_s = ASTFProgram()
        prog_s.send(http_req)
```

hhaim:
Looking into the TCP code, it is not good. 
```
HOT_FUNC bool CFlowTable::update_tcp_flow(CTcpPerThreadCtx * ctx,
                                    CTcpFlow *  flow,
                                    struct rte_mbuf * mbuf,
                                    CSimplePacketParser& parser,
                                    CFlowKeyTuple& tuple,
                                    CFlowKeyFullTuple& ftuple){

   
```
You can't use the L7 before the TCP layer has process it as this could be retransmition or an error, you should use the L7 data after TCP layer has worked and from the right direction 
The code does not look good and will have a very heavy performance penalty for the common case that there is no conflict. It has a duplication of code too.
The defer emulation should be done in the Emulation layer with a dummy program that has "receive" that will update later. 
The flow_table code should have **minimal** change as it is a hot spot for performance

jsmoon:
First of all, `update_tcp_flow` won't work on the common case (if you meant it no L7 map usage case).
```
            if (unlikely(flow->m_pctx->is_on_flow())) {
                if (!update_tcp_flow(ctx,flow,mbuf,parser,tuple,ftuple)) {
                    return false;
                }
            }
```


> You can't use the L7 before the TCP layer has process it as this could be retransmition or an error, you should use the L7 data after TCP layer has worked and from the right direction

1. I don't understand why I can't refer to the L7 data. Since I assume the L7 data itself will not be crashed, I think I can refer it to decide the server template. What's your exact concern about it? My assumption is wrong?

> The code does not look good and will have a very heavy performance penalty for the common case that there is no conflict. It has a duplication of code too.

2. I've already mentioned the performance issue when the L7 map used. But there is a trivial penalty (pctx->is_on_flow() checking added) for the common case.

> It has a duplication of code too.

3. I think it requires some refactoring of the flow_table code for this. I'd like to do it also after this feature is stabilized. Or you can do it as you wish.

> The defer emulation should be done in the Emulation layer with a dummy program that has "receive" that will update later.

4. The emulation starts with flow allocation, but the server program is deferred until the right server program identified. So, I think the dummy program is not needed. Did I miss something?

> The flow_table code should have **minimal** change as it is a hot spot for performance

5. Agreed. Which codes should I change for this?

hhaim:
Let's start with this:

>> You can't use the L7 before the TCP layer has process it as this could be retransmition or an error, you should use the L7 data after TCP layer has worked and from the right direction

>> I don't understand why I can't refer to the L7 data. Since I assume the L7 data itself will not be crashed, I think I can refer it to decide the server template. What's your exact concern about it? My assumption is wrong?

By using the raw packet **before** before the TCP layer you are skipping important TCP logic. Let's take an example  
SYN  
         SYN-ACK
GET -wrong seq (-1)
GET -right seq    (-2)

The  TCP GET (-1) packet is with a **wrong** TCP sequence numbers and in your case you will look into the L7 while the right packet is (-2)

Another example, if there are a few packets that has been sent, each with **one** L7 byte, with TCP you will get all the stream at the end with enogth data 

My suggestion is to create a special emulation instruction called ```recv-defer``` that will receive the L7 data and look into it and update the emu-program or terminate if nothing valid was received 

Corner case is a specific retrtansmition that L7 will get **only the number of bytes** and not the mbuf. There is a need to fix this to save a minimum of 20 bytes per flow 

see https://github.com/cisco-system-traffic-generator/trex-core/blob/master/src/44bsd/tcp_input.cpp#L361

With this design there would be zero performance penalty for backward compatible cases

jsmoon:
> The TCP GET (-1) packet is with a wrong TCP sequence numbers and in your case you will look into the L7 while the right packet is (-2)

Do you mean that TCP GET (-1) packet has **different payload contents** compare to TCP GET (-2)?
Let me know how can it be possible in the network. Is it possible in the general case?

> Another example, if there are a few packets that has been sent, each with one L7 byte, with TCP you will get all the stream at the end with enogth data

Currently, the profile will not be loaded without "value" in this case. It's a kind of restriction on my implementation. **Do you want the restriction to be removed?**

> There is a need to fix this to save a minimum of 20 bytes per flow
> 
> see https://github.com/cisco-system-traffic-generator/trex-core/blob/master/src/44bsd/tcp_input.cpp#L361
> 
> With this design there would be zero performance penalty for backward compatible cases

Currently, I don't understand how comes the zero performance penalty achieved. I think I need to understand the TCP process more.

**Anyway, I'll try to find a way to remove your concern.** I think this takes time also.

hhaim:
>> Do you mean that TCP GET (-1) packet has different payload contents compare to TCP GET (-2)?
Let me know how can it be possible in the network. Is it possible in the general case?

Of course, everything could happen - this is what we need to verify as a TGN.  The L7 data could be different between the packet#1 and packet#2 as packet#1 will be ignored by TCP.  

For doing it right, you need to simulate server application that works on top of  TCP layer. This is how emulation layer works

egwakim:
Hi @hhaim ,
There seems 2 topics(optimize common case, pcap utility support) are under discussing now.
Since it's new feature which backward compatible with old legacy usage, 'pcap utility' can be done later. We can do it another PR. 
Do you agree with that?

Regarding 'optimize common' case, 
Can you explain more details about your idea to optimize it?

Best Regards
Gwangmoon

hhaim:
@egwakim  there are 3 topics 
1. Optimization  and pcap utility is less of an issue. 
2. The TCP topic is more important as it is a functional issue and **not** only a performance issue. 
if you remove the TCP support and improve the UDP path (look into the port map first) it would be possible to continue

hhaim:
@jsmoon 

You should put all the logic in the CEmulAppCmd object and not in the low level tcp/flow_table 
https://github.com/cisco-system-traffic-generator/trex-core/blob/master/src/44bsd/tcp_socket.cpp

The CEmulAppCmd get callbacks of ``sbappend(..mbuf)`` or ``sbappend_bytes(..bytes) ``
By default the retransmission buffers are not saved (just the size). You should add support for merging the buffers to the upper layer and provide the **same** API ``sbappend(..mbuf) `` to be processed by the upper layer (not your hack callback) -- this buffer will be processed by a new command ```rcv_choose``` that will look into the buffer and update the emu program 

Try to remove all the new logic you added from flow_table/tcp 

tcp object should include a generic flag that makes the retransmission **save** mbufs and merge them 

I would try with this task as a separate PR first  and a new instruction that signal for merging retransmition the mbuf instead of the size in this case. 
I would check this in the simulator as there could be many case of errors 

After this commit, let's try to add this PR 

For example, this code should be put in the CEmulAppCmd layer (application layer)
You are trying to change things under the hood..

```
          bool is_flow = process_tcp_packet(ctx,flow,mbuf,lpTcp,ftuple);
            if (unlikely(is_flow && flow->m_pctx->is_on_flow())) {
                return update_tcp_flow(ctx,flow,parser,ftuple);

```

hhaim:
@jsmoon thinking about it more, it is better to provide a *new* api for the application layer 
```sbappend_buffer(*char ,size)``` as with mbuf we might be limited with retransmition with the buffer size

**Comment by jsmoon** [comment]
@hhaim, I think my code collects correctly received data into
flow->m_payload_buf. So, I couldn't understand what you mentioned.
And I'm not sure how I handle your suggestion.
According to my understanding, although there is no application start, the
emulation layer can receive payloads in the TCP layer. I think there is no
callback to program in this situation. So, I think the app can be started
after the right program is identified.
Please correct me if I'm wrong.

P.S. What you mean "your hack callback"?

2020년 1월 22일 (수) 오후 11:02, Hanoh Haim <notifications@github.com>님이 작성:

**Comment by hhaim** [comment]
@jsmoon 
You are mixing a few features in one PR.
Let’s split this PR to a few smaller PR
1. Only UDP with the optimization we talked about 
2. Only TCP retransmission callback that support buffer as an input (General PR not related to this PR)
3. New instruction that read the buffer and choose the right template

**Comment by jsmoon** [comment]
> 1. Only UDP with the optimization we talked about

This PR is now fit for this request.

**Comment by hhaim** [comment]
rebased. 
Please increment the ASTF minor version in the next commit. 

next commit try to add  PR only for the **TCP retransmission buffer support** (not related to this issue)

1. Add new command ```rcv_cmp (string)``` 
2. This command will enable a new TCP flow flag that let it collect the out of order mbufs into the vector (will be allocated only when needed, instead of a size callback). When the retransmission buffer is ready it will call with a new callback ```sbappend_buffer(*char ,size)```
3. Create a simulation test with drop/OOO and check that it works and get the right buffer with ``rcv_cmp (string)```

**Comment by egwakim** [comment]
Hi @hhaim 
Is "rcv_cmp(string)" means that new ASTF program command which will act like l7_map() API?
Can you explain the benefit of adding this new API?

BR
Gwangmoon

**Comment by hhaim** [comment]
@egwakim this PR (``rcv_cmp(string)``) have nothing to do with the #381 it just a prerequisite 
Because it there should be a new API to the L7 application layer, I suggested to add this rcv_cmp to verify the new API using the simulator that has very expensive tests in regards to this. 
This feature is a TCP feature and need to test very carefully. 
Once this PR will be tested and implemented, we could add a **new** API to choose the template based on the data (direct or buffer API)

Make sense?

**Comment by hhaim** [comment]
@jsmoon this PR was rebased already. please create a new PR

**Comment by jsmoon** [comment]
@hhaim O.K. I'll create a new PR for **ASTF minor version update**.

**Comment by egwakim** [comment]
Hi @hhaim ,
We have new API l7_map() now. It can be used for both UDP/TCP case.
For TCP case, you suggested additional light-weighted(?) new API 'rcv_cmp()' to verify TCP feature in simulator. 
rcv_cmp(string) API will be partial set of existing l7_map() API which can cover only sequential string payload only.
Benefit of new "TCP re-transmission buffer support" feature will be 
1) Slightly better performance.
2) Simpler API, but, limited coverage

What is wanted way of handling this case?
We have several options to handle it.
Option#1. Just add test cases to cover current PR with l7_map API - no new rcv_cmp() API
Option#2. Add new rcv_cmp() API with new TCP re-transmission buffer support. 
- Create new issue " TCP re-transmission buffer support"

If you need Option#2, when we should do it?

**Comment by hhaim** [comment]
@egwakim 
We are out of sync. The reason the new ```rcv_cmp(string)``` instruction was suggested is to test the new buffer API (```sbappend_buffer(..buffer)```) - to provide the app layer the buffer in case of OOO or error. Once this will be done as a **separate** PR and a **separate** feature it would be easy to add the **TCP** ```l7_map ``` instruction to support #381 
The retransmition block is a sensitive block in the TCP and need to test it carefully. Without this feature the map_l7 will not work in some cases

**Comment by egwakim** [comment]
Ok, I see.

**Comment by jsmoon** [comment]
@hhaim,
According to my investigation, I think `sbappend_buffer` is not required.
I'll remove `sbappend_bytes` and make `sbappend` keep the mbuf in `CEmulApp` until `rx` command handled. Definitely, `pre_tcp_reass` also keeps the mbuf and the mbuf will be handled properly by `tcp_reass_no_data`.
Only I concerned with this idea that the mbuf lifetime would be longer than before.
My first commit will implement this idea only (without `recv_cmp(string)` implementation).
What do you think?

**Comment by darrentodd** [comment]
I tried the example astf/shared_port.py script here; changing http_req1 and http_req2 (lines 8 and 9) to:

```
      8 http_req1 = b'GET /thisisrequestone HTTP/1.1\r\nHost: 22.0.0.3\r\nConnection: Keep-Alive\r\nUser-Agent: Mozi        lla/4.0 (compatible; MSIE 7.0; Windows NT 5.1; SV1; .NET CLR 1.1.4322; .NET CLR 2.0.50727)\r\nAccept: */*\r\        nAccept-Language: en-us\r\nAccept-Encoding: gzip, deflate, compress\r\n\r\n'
      9 http_req2 = b'GET /thisisrequesttwo HTTP/1.1\r\nHost: 22.0.0.3\r\nConnection: Keep-Alive\r\nUser-Agent: Mozi        lla/4.0 (compatible; MSIE 7.0; Windows NT 5.1; SV1; .NET CLR 1.1.4322; .NET CLR 2.0.50727)\r\nAccept: */*\r\        nAccept-Language: en-us\r\nAccept-Encoding: gzip, deflate, compress\r\n\r\n'

```

Running it up on v2.75 TRex crashed with segmentation fault::

```
Error: signal 11:

*** traceback follows ***

1       0x559800e74f04 ./_t-rex-64(+0x17cf04) [0x559800e74f04]
2       0x7f025307a630 /lib64/libpthread.so.0(+0xf630) [0x7f025307a630]
3       0x559800e26507 CServerIpPayloadInfo::is_server_compatible(CTcpServerInfo*) + 23
4       0x559800e2ad50 CServerPortInfo::insert_template_payload(CTcpServerInfo*, CPerProfileCtx*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) + 3360
5       0x559800e2c25a CServerPortInfo::insert_template_info(CTcpServerInfo*, CPerProfileCtx*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) + 74
6       0x559800e2d449 CTcpPerThreadCtx::append_server_ports(unsigned int) + 393
7       0x559800e41424 CFlowGenListPerThread::load_tcp_profile(unsigned int, bool) + 676
8       0x55980107ac30 TrexAstfDpCore::create_tcp_batch(unsigned int, double) + 1184
9       0x55980107d407 TrexAstfDpCreateTcp::handle(TrexDpCore*) + 23
10      0x559800fc8970 TrexDpCore::handle_cp_msg(TrexCpToDpMsgBase*) + 16
11      0x559800e65512 CFlowGenListPerThread::check_msgs() + 338
12      0x559800fc851b TrexDpCore::idle_state_loop() + 59
13      0x559800fc865b TrexDpCore::start() + 59
14      0x559800e5c4f3 CFlowGenListPerThread::start(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, CPreviewMode&) + 115
15      0x559800dffc15 CGlobalTRex::run_in_core(unsigned char) + 487
16      0x559800e1e6bd ./_t-rex-64(+0x1266bd) [0x559800e1e6bd]
17      0x559801144e5d eal_thread_loop + 477
18      0x7f0253072ea5 /lib64/libpthread.so.0(+0x7ea5) [0x7f0253072ea5]
19      0x7f02520e78cd clone + 109


*** addr2line information follows ***

??:0
??:0
??:0
??:0
??:0
??:0
??:0
??:0
??:0
??:0
??:0
??:0
??:0
??:0
??:0
??:0
??:0
??:0
??:0


./t-rex-64: line 100: 33325 Segmentation fault      (core dumped) ./_$(basename $0) $INPUT_ARGS $EXTRA_INPUT_ARGS

```

**Comment by jsmoon** [comment]
@darrentodd I've just tried to reproduce your case, but failed (no core dumped). Could you give your changed shared_port.py file? And could you elaborate this situation?

**Comment by darrentodd** [comment]
@jsmoon sure of course; file is as:

```
# Example for creating your program by specifying buffers to send, without relaying on pcap file

from trex.astf.api import *


# we can send either Python bytes type as below:
http_req = b'GET /3384 HTTP/1.1\r\nHost: 22.0.0.3\r\nConnection: Keep-Alive\r\nUser-Agent: Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1; SV1; .NET CLR 1.1.4322; .NET CLR 2.0.50727)\r\nAccept: */*\r\nAccept-Language: en-us\r\nAccept-Encoding: gzip, deflate, compress\r\n\r\n'
http_req1 = b'GET /thisisrequestone HTTP/1.1\r\nHost: 22.0.0.3\r\nConnection: Keep-Alive\r\nUser-Agent: Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1; SV1; .NET CLR 1.1.4322; .NET CLR 2.0.50727)\r\nAccept: */*\r\nAccept-Language: en-us\r\nAccept-Encoding: gzip, deflate, compress\r\n\r\n'
http_req2 = b'GET /thisisrequesttwo HTTP/1.1\r\nHost: 22.0.0.3\r\nConnection: Keep-Alive\r\nUser-Agent: Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1; SV1; .NET CLR 1.1.4322; .NET CLR 2.0.50727)\r\nAccept: */*\r\nAccept-Language: en-us\r\nAccept-Encoding: gzip, deflate, compress\r\n\r\n'
# or we can send Python string containing ascii chars, as below:
http_response = 'HTTP/1.1 200 OK\r\nServer: Microsoft-IIS/6.0\r\nContent-Type: text/html\r\nContent-Length: 32000\r\n\r\n<html><pre>**********</pre></html>'

class Prof1():
    def __init__(self):
        pass  # tunables

    def _udp_client_prog(self, req, res):
        prog_c = ASTFProgram(stream=False)
        prog_c.send_msg(req)
        prog_c.recv_msg(1)
        return prog_c

    def _udp_server_prog(self, req, res):
        prog_s = ASTFProgram(stream=False)
        prog_s.recv_msg(1)
        prog_s.send_msg(res)
        return prog_s

    def _tcp_client_prog(self, req, res):
        prog_c = ASTFProgram(stream=True)
        prog_c.send(req)
        prog_c.recv(len(res))
        return prog_c

    def _tcp_server_prog(self, req, res):
        prog_s = ASTFProgram(stream=True)
        prog_s.recv(len(req))
        prog_s.send(res)
        return prog_s

    def create_profile(self, proto='all', temp='all'):
        # UDP client and server commands
        udp_prog_c = self._udp_client_prog(http_req, http_response)
        udp_prog_c1 = self._udp_client_prog(http_req1, http_response)
        udp_prog_c2 = self._udp_client_prog(http_req2, http_response)

        udp_prog_s = self._udp_server_prog(http_req, http_response)

        # TCP client and server commands
        tcp_prog_c = self._tcp_client_prog(http_req, http_response)
        tcp_prog_c1 = self._tcp_client_prog(http_req1, http_response)
        tcp_prog_c2 = self._tcp_client_prog(http_req2, http_response)

        tcp_prog_s = self._tcp_server_prog(http_req, http_response)
        tcp_prog_s1 = self._tcp_server_prog(http_req1, http_response)
        tcp_prog_s2 = self._tcp_server_prog(http_req2, http_response)

        # ip generator
        ip_gen_c = ASTFIPGenDist(ip_range=["16.0.0.0", "16.0.0.255"], distribution="seq")
        ip_gen_s = ASTFIPGenDist(ip_range=["48.0.0.0", "48.0.255.255"], distribution="seq")
        ip_gen = ASTFIPGen(glob=ASTFIPGenGlobal(ip_offset="1.0.0.0"),
                           dist_client=ip_gen_c,
                           dist_server=ip_gen_s)

        # use default port 80
        assoc_by_ip = ASTFAssociationRule(ip_start="48.0.0.0", ip_end="48.0.255.255")
        assoc_by_l7 = ASTFAssociationRule(ip_start="48.0.0.0", ip_end="48.0.255.255",
                                          l7_map={"offset":[0,1,2,3]})

        templates = []
        # UDP templates
        udp_temp_c = ASTFTCPClientTemplate(program=udp_prog_c, ip_gen=ip_gen)
        udp_temp_s = ASTFTCPServerTemplate(program=udp_prog_s, assoc=assoc_by_ip)
        udp_template = ASTFTemplate(client_template=udp_temp_c, server_template=udp_temp_s, tg_name='udp')

        udp_temp_s1 = ASTFTCPServerTemplate(program=udp_prog_s, assoc=assoc_by_l7)
        udp_temp_c1 = ASTFTCPClientTemplate(program=udp_prog_c1, ip_gen=ip_gen)
        udp_template1 = ASTFTemplate(client_template=udp_temp_c1, server_template=udp_temp_s1, tg_name='udp1')

        udp_temp_c2 = ASTFTCPClientTemplate(program=udp_prog_c2, ip_gen=ip_gen)
        udp_template2 = ASTFTemplate(client_template=udp_temp_c2, server_template=udp_temp_s1, tg_name='udp2')

        if proto == 'all' or proto == 'udp':
            if temp == 'all' or temp == 'ip':
                templates += [ udp_template ]
            if temp == 'all' or temp == 'payload':
                templates += [ udp_template1, udp_template2 ]

        # TCP templates
        tcp_temp_c = ASTFTCPClientTemplate(program=tcp_prog_c, ip_gen=ip_gen)
        tcp_temp_s = ASTFTCPServerTemplate(program=tcp_prog_s, assoc=assoc_by_ip)
        tcp_template = ASTFTemplate(client_template=tcp_temp_c, server_template=tcp_temp_s, tg_name='tcp')

        tcp_temp_c1 = ASTFTCPClientTemplate(program=tcp_prog_c1, ip_gen=ip_gen)
        tcp_temp_s1 = ASTFTCPServerTemplate(program=tcp_prog_s1, assoc=assoc_by_l7)
        tcp_template1 = ASTFTemplate(client_template=tcp_temp_c1, server_template=tcp_temp_s1, tg_name='tcp1')

        tcp_temp_c2 = ASTFTCPClientTemplate(program=tcp_prog_c2, ip_gen=ip_gen)
        tcp_temp_s2 = ASTFTCPServerTemplate(program=tcp_prog_s2, assoc=assoc_by_l7)
        tcp_template2 = ASTFTemplate(client_template=tcp_temp_c2, server_template=tcp_temp_s2, tg_name='tcp2')

        if proto == 'all' or proto == 'tcp':
            if temp == 'all' or temp == 'ip':
                templates += [ tcp_template ]
            if temp == 'all' or temp == 'payload':
                templates += [ tcp_template1, tcp_template2 ]

        # profile
        profile = ASTFProfile(default_ip_gen=ip_gen, templates=templates)
        return profile

    def get_profile(self, **kwargs):
        proto = kwargs.get('proto','all').lower()
        temp = kwargs.get('temp','all').lower()
        return self.create_profile(proto, temp)


def register():
    return Prof1()

```

I run up v2.75 in astf mode:

`# ./t-rex-64 -i --astf -c 12 --cfg /etc/trex_cfg_80.yaml `

YAML: 

```
### Config file generated by dpdk_setup_ports.py ###

- version: 2
  interfaces: ['05:00.0', '05:00.1']
  port_info:
      - ip: 172.30.0.16
        default_gw: 172.30.0.100
        vlan: 400
      - ip: 172.31.0.16
        default_gw: 172.31.0.100
        vlan: 401

  platform:
      master_thread_id: 0
      latency_thread_id: 18
      dual_if:
        - socket: 0
          threads: [1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,36,37,38,39,40,41,42,43,44,45,46,47,48,49,50,51,52,53]

```

run up trex-console & 

`trex>start -f astf/shared_port.py -t proto=tcp,temp=payload`

this gives me:

`Last profile _ command failed: Could not create ASTF batch: Server for TCP port 80: already registered in [30000000, 3000ffff]
`

I then reset & rerun the same command

```
trex>reset

Resetting ports [0, 1]:                                      [SUCCESS]

416.77 [ms]

trex>start -f astf/shared_port-dt.py -t proto=tcp,temp=payload

Loading traffic at acquired ports.                           [SUCCESS]


Starting traffic.                                            [SUCCESS]

5.87 [ms]

trex>
trex>
trex>

2020-02-27 12:45:24 - [local][warning]   - Connection lost - Subscriber timeout: no data from TRex server for more than 5 seconds
```

The Connection lost - here the TRex engine has crashed.

jsmoon:
@darrentodd I still cannot reproduce your case. I'll try it more later.
What is the `astf/shared_port-dt.py`? Is it the same as `astf/shared_port.py`?

PR #405, opened 2020-01-10
fix recursive port alloc fail possibility

jsmoon:
Hi,
I found a possibility of recursive port allocation fails during the min port is occupied.


hhaim:
rebased

PR #406, opened 2020-01-10
Update recv(bytes,clear=True) function issue in ASTFProgram

jsmoon:
Hi,
I found that previous `recv(bytes,clear=True) function issue in ASTFProgram` implementation may have different behavior compared to the document and server.
The client clears before recv(). But, the server clears after recv() done.
My change will make the client clear after recv() done also.


jsmoon:
The failure occurred at STL performance test. I think my change does not affect it.

hhaim:
reabsed

PR #407, opened 2020-01-13
fix STL stream vm program size overflow issue by massive value_list

jsmoon:
Hi,
This change will solve the issue when the STL flow_var value_list is massively used.
When the total size of the value_list is over 64K, the unexpected results will occur.
So, I removed the limitation of 64K opcodes size by this change.

Please check my code changes and give your feedback.


egwakim:
Hi @hhaim,
Can you review this change?

hhaim:
rebased

PR #422, opened 2020-01-31
add support for memif PMD from dpdk 19.11

juliusmilan:
memif PMD from dpdk v19.11 without last commit: "net/memif: support zero-copy slave",
does not require changes in dpdk lib code

Note: to specify memif interfaces in trex yaml config, whole --vdev kind of specification is used.
For example I tested with following interfaces specification:
interfaces: ["--vdev=net_memif0,socket=/run/vpp/memif.sock,id=0", "--vdev=net_memif1,socket=/run/vpp/memif.sock,id=1"]
With VPP as DUT.

Signed-off-by: Július Milan <julius.milan@pantheon.tech>


hhaim:
nice! @juliusmilan 
Could you share an example how to configure it from TRex side and NFV side  
Is multi-core supported only for STL ?. I will incorporate it in the documentation

juliusmilan:
Thank you @hhaim
Sure, there is a foursome of arguments to interconnect 2 memifs to communicate: socket, id, role and secret.
Socket, id and secret needs to be the same (for pair of connected memifs), role has to differ.
One of pair needs to have role "master" and second "slave".
Socket is for exchange of protocol control messages. (by default /run/vpp/memif.sock in FD.io VPP)
Id just has to match :) it's unique for every pair of memifs.
And secret is optional string (by default empty) wich has to match.
For example for yaml trex memif interface definition mentioned above:
interfaces: ["--vdev=net_memif0,socket=/run/vpp/memif.sock,role=slave,id=0", "--vdev=net_memif1,socket=/run/vpp/memif.sock,role=slave,id=1"]
to create corresponding interfaces in VPP I used following CLI commands:
vpp# create interface memif id 0 master
vpp# set interface state memif0/0 up
vpp# create interface memif id 1 master
vpp# set interface state memif0/1 up
Then you can check the state by:
vpp# show memif
You should see "connected" if everything works as expected.
You can try to connect 2 memifs also using testpmd of DPDK. Maybe that would be even more straightforward.
Hope this helps. I plan to create a demo as well, will send you an update hopefully soon.
> Is multi-core supported only for STL ?
What do you mean? What is STL?
I know it is possible to create memif with multpiple RX/TX queues, but not sure if that's the

juliusmilan:
Hi @hhaim, I created a demo of usage, you can take a look here:
https://asciinema.org/a/OJjNL9Vv1unL2Dv9BQ5W2z6Hy

Please note that there is use of start --force ... in trex-console yet.
It is caused by missing peace of information in DPDK memif PMD.
I already sent them a patch, will give you an update as soon as they will merge it for me.

hhaim:
@juliusmilan nice!
so to loopback TRex memif ports,  it would be 

```
interfaces: ["--vdev=net_memif0,socket=/run/vpp/memif.sock,role=master,id=0", "--vdev=net_memif1,socket=/run/vpp/memif.sock,role=slave,id=0"]
```

I will update the doc

juliusmilan:
@hhaim yes, you are right, that is valid configuration for loopback. However I tested it and it does not work with current DPDK memif PMD due to some checks.
I am going to send another patch to DPDK to enable it so that your configuration will work.
Will give you an update. By the way my previous patch they already added to dpdk next branch, I will open a PR for you when they will accept it into their master.

Just little note: for even more simple example of loopback you can use the default DPDK memif socket: /tmp/memif.sock as described here: https://doc.dpdk.org/guides/nics/memif.html
/run/vpp/memif.sock is the default one for FD.io VPP (but it will work as well)
So you can use also this:
```
interfaces: ["--vdev=net_memif0,role=master,id=0", "--vdev=net_memif1,role=slave,id=0"]
```
Next please note, that memif can run in polling mode (default), so to achieve best performance, every memif port should obtain one CPU core.

hhaim:
@juliusmilan 
loopback does not work for me, what am I missing?

```
- port_limit    : 2         
  version       : 2
  interfaces    : ["--vdev=net_memif0,role=master,id=0", "--vdev=net_memif1,role=slave,id=0" ] 
  c : 1
  platform :
        master_thread_id  : 0
        latency_thread_id : 5
        dual_if   :
             - socket   : 0
               threads  : [1,2,3,4]
             - socket   : 1
               threads  : [8,9,10,11]
```

```
[csi-kiwi-02]> sudo ./t-rex-64 -i --cfg /etc/trex_memif.yaml -c 1
Starting Scapy server..... Scapy server is started
The ports are bound/configured.
Starting  TRex v2.73 please wait  ... 
EAL: Started without hugepages support, physical addresses not available
**memif_socket_create(): NULL: Failed to setup socket /run/memif.sock: Address already in use**
vdev_probe(): failed to initialize net_memif0 device
EAL: Bus (vdev) probe failed.
Could not find all interfaces (asked for: 2, found: 1).
Killing Scapy server... Scapy server is killed
```

when both master, it works

juliusmilan:
@hhaim The problem is on DPDK side and needs to be fixed in latest DPDK, I already found a fix and going to send them a patch. You are right, loopback does not work yet. I will send you an update as soon as they will accept my patch, I think 1-2 weeks is realistic estimate for them. If you want, I can send you PR with fix even before they will accept it, but I wanted to have a review from them before.

hhaim:
Understood. Please send a PR I can test it

juliusmilan:
Here it is: https://github.com/cisco-system-traffic-generator/trex-core/pull/426
I discovered another minor problem with memif PMD, the memif socket must be removed every time before running trex as:
sudo rm /run/memif.sock
However with PR above, the loopback is working

PR #423, opened 2020-02-03
TCP retransmission assemble buffer support

jsmoon:
@hhaim This PR is the first commit for implementing the `rcv_cmp` concept as discussed at #403.
I didn't implement `rcv_cmp` command itself, but I added a test of the functionality at `gt_tcp.tst43`.
Please review my changes and give your feedback.

### DIFF
--- a/src/44bsd/tcp_input.cpp
+++ b/src/44bsd/tcp_input.cpp
@@ -164,6 +164,7 @@ int CTcpReass::tcp_reass_no_data(CPerProfileCtx * pctx,
 
     flags = (m_blocks[0].m_flags==1) ? TH_FIN :0;
 
+    tp->add_rx_bytes(m_blocks[0].m_len);    // mbuf already updated to rx_data
     sbappend_bytes(&tp->m_socket,
                    &tp->m_socket.so_rcv ,
                    m_blocks[0].m_len);
@@ -197,6 +198,8 @@ int CTcpReass::pre_tcp_reass(CPerProfileCtx * pctx,
 
     assert(ti);
     if (m) {
+        tp->update_rx_data(m, ti->ti_seq, ti->ti_len);
+        /* mbuf updated at seq, no need to keep mbuf any more */
         rte_pktmbuf_free(m);
     }
     uint16_t tg_id = tp->m_flow->m_tg_id;
@@ -376,6 +379,7 @@ inline void TCP_REASS(CPerProfileCtx * pctx,
         uint16_t tg_id = tp->m_flow->m_tg_id;
         INC_STAT(pctx, tg_id, tcps_rcvpack);
         INC_STAT_CNT(pctx, tg_id, tcps_rcvbyte,ti->ti_len);
+        tp->append_rx_data(m, ti->ti_seq, ti->ti_len);
         sbappend(so,
                  &so->so_rcv, m,ti->ti_len); 
         sorwakeup(so); 
@@ -706,6 +710,7 @@ HOT_FUNC int tcp_flow_input(CPerProfileCtx * pctx,
              */
             tcp_pktmbuf_fix_mbuf(m, off,ti->ti_len);
 
+            tp->append_rx_data(m, ti->ti_seq, ti->ti_len);
             sbappend(so,
                      &so->so_rcv, m,ti->ti_len);
             sorwakeup(so);
--- a/src/44bsd/tcp_var.h
+++ b/src/44bsd/tcp_var.h
@@ -95,6 +95,31 @@ struct CTcpPkt {
 
 #define PACKET_TEMPLATE_SIZE (128)
 
+class CTcpRxData {
+public:
+    CTcpRxData(uint32_t seq, uint16_t size) {
+        m_size = 0;
+        m_bytes = 0;
+        if (size) {
+            m_data = (uint8_t*)calloc(1, size);
+        }
+        if (m_data) {
+            m_seq = seq;
+            m_size = size;
+        }
+    }
+    CTcpRxData() : CTcpRxData(0, 0) {}
+    ~CTcpRxData() {
+        free(m_data);
+        m_data = nullptr;
+    }
+
+    uint32_t    m_seq;
+    uint16_t    m_size;
+    uint16_t    m_bytes;  // saved bytes
+    uint8_t*    m_data;
+};
+
 
 struct tcpcb {
 
@@ -213,11 +238,17 @@ struct tcpcb {
     CTcpReass * m_tpc_reass; /* tcp reassembley object, allocated only when needed */
     CTcpFlow  * m_flow;      /* back pointer to flow*/
 
+    CTcpRxData* m_rx_data;
+
 public:
 
     
     tcpcb() {
         m_tuneable_flags = 0;
+        m_rx_data = nullptr;
+    }
+    ~tcpcb() {
+        clear_rx_data();
     }
 
     inline bool is_tso(){
@@ -242,7 +273,73 @@ public:
     inline rte_mbuf_t   * pktmbuf_alloc(uint16_t size){
         return (tcp_pktmbuf_alloc(mbuf_socket,size));
     }
-    
+
+public:
+    uint16_t update_rx_data(struct rte_mbuf* m, uint32_t seq, uint16_t len) {
+        if (m_rx_data && (m_rx_data->m_bytes < m_rx_data->m_size)) {
+            assert(m->nb_segs == 1);
+            uint8_t* p = rte_pktmbuf_mtod(m, uint8_t*);
+            int32 q = m_rx_data->m_seq + m_rx_data->m_size - seq;
+            if (q > 0) {
+                uint16_t cp_len = std::min(q, (int32)len);
+                memcpy(m_rx_data->m_data + m_rx_data->m_size - q, p, cp_len);
+                return cp_len;
+            }
+        }
+        return 0;
+    }
+    void append_rx_data(struct rte_mbuf* m, uint32_t seq, uint16_t len) {
+        if (m_rx_data) {
+            m_rx_data->m_bytes += update_rx_data(m, seq, len);
+        }
+    }
+    void add_rx_bytes(uint16_t len) {
+        if (m_rx_data) {
+            m_rx_data->m_bytes += len;
+        }
+    }
+
+    void clear_rx_data() {
+        if (m_rx_data) {
+            delete m_rx_data;
+            m_rx_data = nullptr;
+        }
+    }
+
+    void setup_rx_data(uint16_t size) {
+        clear_rx_data();
+        // NOTE: rx_data may not be filled fully if m_tpc_reass is not nullptr
+        m_rx_data = new CTcpRxData(rcv_nxt, size);
+    }
+
+    inline uint8_t* get_rx_data() const { return m_rx_data ? m_rx_data->m_data: nullptr; }
+    inline uint32_t get_rx_size() const { return m_rx_data ? m_rx_data->m_bytes: 0; }
+
+#ifdef  TREX_SIM
+    inline uint8_t get_rx_byte(uint16_t off) {
+        if (m_rx_data) {
+            return off < m_rx_data->m_size ? m_rx_data->m_data[off]: 0;
+        }
+        return 0;
+    }
+
+    bool expect_rx_data(std::vector<uint8_t>& data) {
+        uint32_t size = data.size();
+        if (get_rx_size() < size) {
+            fprintf(stdout, "rx_data size %u is less than %u.\n", get_rx_size(), size);
+            return false;
+        }
+        for (int i = 0; i < size; i++) {
+            auto cmp_byte = data[i];
+            auto rx_byte = get_rx_byte(i);
+            if (cmp_byte != rx_byte) {
+                fprintf(stdout, "data different between rx(%x) and cmp(%x).\n", cmp_byte, rx_byte);
+                return false;
+            }
+        }
+        return true;
+    }
+#endif
 };
 
 #define intotcpcb(ip)   ((struct tcpcb *)(ip)->inp_ppcb)
--- a/src/gtest/bp_tcp_gtest.cpp
+++ b/src/gtest/bp_tcp_gtest.cpp
@@ -130,6 +130,36 @@ TEST_F(gt_tcp, tst4) {
 }
 
 
+class CTcpReassMBuf {
+    rte_mempool_t * m_mempool;
+public:
+    CTcpReassMBuf() {
+        m_mempool = utl_rte_mempool_create("big-const", 10, 9000, 32, 0, false);
+    }
+    ~CTcpReassMBuf() {
+        utl_rte_mempool_delete(m_mempool);
+    }
+
+    struct rte_mbuf* alloc_mbuf(size_t len) {
+        assert(len <= m_mempool->elt_size);
+        struct rte_mbuf* m = rte_pktmbuf_alloc(m_mempool);
+        m->data_len = m->pkt_len = len;
+        return m;
+    }
+};
+
+static CTcpReassMBuf tcp_reass_mbuf;
+
+struct rte_mbuf* tcp_reass_mbuf_alloc(size_t len) {
+    return tcp_reass_mbuf.alloc_mbuf(len);
+}
+
+struct rte_mbuf* tcp_reass_mbuf_alloc(size_t len, std::vector<uint8_t>& data) {
+    auto mbuf = tcp_reass_mbuf.alloc_mbuf(len);
+    uint8_t *pkt = rte_pktmbuf_mtod(mbuf, uint8_t *);
+    memcpy(pkt, data.data(), len > data.size() ? data.size() : len);
+    return mbuf;
+}
 
 
 class CTcpReassTest  {
@@ -167,7 +197,7 @@ void CTcpReassTest::add_pkts(vec_tcp_reas_t & lpkt){
         ti.ti_seq  = lpkt[i].m_seq;
         ti.ti_len  = (uint16_t)lpkt[i].m_len;
         ti.ti_flags =lpkt[i].m_flags;
-        m_tcp_res.pre_tcp_reass(&m_ctx,&m_flow.m_tcp,&ti,(struct rte_mbuf *)0);
+        m_tcp_res.pre_tcp_reass(&m_ctx,&m_flow.m_tcp,&ti, tcp_reass_mbuf_alloc(ti.ti_len));
         if (m_verbose) {
             fprintf(stdout," inter %d \n",i);
             m_tcp_res.Dump(stdout);
@@ -176,7 +206,7 @@ void CTcpReassTest::add_pkts(vec_tcp_reas_t & lpkt){
 }
 
 void CTcpReassTest::expect(vec_tcp_reas_t & lpkt,FILE *fd){
-    m_tcp_res.expect(lpkt,fd);
+    EXPECT_TRUE(m_tcp_res.expect(lpkt,fd));
 }
 
 
@@ -1685,39 +1715,38 @@ TEST_F(gt_tcp, tst42) {
     ti.ti_seq = 0x1000+100;
 
     int tiflags=0;
-
-    tiflags = tcp_reass(&ctx,tp, &ti, (struct rte_mbuf *)0); 
+    tiflags = tcp_reass(&ctx,tp, &ti, tcp_reass_mbuf_alloc(ti.ih_len));
 
     ti.ih_len = 100;
     ti.ti_seq +=100 ;
-    tiflags = tcp_reass(&ctx,tp, &ti, (struct rte_mbuf *)0); 
+    tiflags = tcp_reass(&ctx,tp, &ti, tcp_reass_mbuf_alloc(ti.ih_len));
 
     tp->m_tpc_reass->Dump(stdout);
 
     ti.ih_len = 100;
     ti.ti_seq =0x1000 ;
-    tiflags = tcp_reass(&ctx,tp, &ti, (struct rte_mbuf *)0); 
+    tiflags = tcp_reass(&ctx,tp, &ti, tcp_reass_mbuf_alloc(ti.ih_len));
 
 
     ti.ih_len = 100;
     ti.ti_seq =0x1000+1000 ;
-    tiflags = tcp_reass(&ctx,tp, &ti, (struct rte_mbuf *)0); 
+    tiflags = tcp_reass(&ctx,tp, &ti, tcp_reass_mbuf_alloc(ti.ih_len));
 
     tp->m_tpc_reass->Dump(stdout);
 
     ti.ih_len = 100;
     ti.ti_seq =0x1000+2000 ;
-    tiflags = tcp_reass(&ctx,tp, &ti, (struct rte_mbuf *)0); 
+    tiflags = tcp_reass(&ctx,tp, &ti, tcp_reass_mbuf_alloc(ti.ih_len));
 
     tp->m_tpc_reass->Dump(stdout);
 
     ti.ih_len = 100;
     ti.ti_seq =0x1000+3000 ;
-    tiflags = tcp_reass(&ctx,tp, &ti, (struct rte_mbuf *)0); 
+    tiflags = tcp_reass(&ctx,tp, &ti, tcp_reass_mbuf_alloc(ti.ih_len));
 
     ti.ih_len = 100;
     ti.ti_seq =0x1000+4000 ;
-    tiflags = tcp_reass(&ctx,tp, &ti, (struct rte_mbuf *)0); 
+    tiflags = tcp_reass(&ctx,tp, &ti, tcp_reass_mbuf_alloc(ti.ih_len));
 
     ctx.get_tcpstat()->Dump(stdout);
     printf(" tiflags:%x \n",tiflags);
@@ -1755,12 +1784,17 @@ TEST_F(gt_tcp, tst43) {
     app_c->set_flow_ctx(&ctx,&flow);
     flow.set_app(app_c);
 
-
     tcpcb * tp=&flow.m_tcp;
 
     tp->rcv_nxt = 0x1000; /* expect this seq in rcv */
 
     tp->t_state = TCPS_SYN_RECEIVED;
+
+    std::vector<uint8_t> cmp_data;
+    for (int i = 0; i <= UINT8_MAX; i++) {
+        cmp_data.push_back(i);
+    }
+    tp->setup_rx_data(cmp_data.size());   /* enable rx data saving */
 
     struct tcpiphdr ti;
     memset(&ti,0,sizeof(struct tcpiphdr));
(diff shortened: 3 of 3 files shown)


hhaim:
@jsmoon it is in the right direction.
I would keep the reassemble buffer as Pointer to vector of char pointers from *heap* memory and *remove* the mbufs as it could be very big (not limited in 9k)

Just keep vector of pointer of data (char *) per *each* section and in case the tcp buffer is enabled allocate and copy the mbuf to allocated buffer. In case of merge just order the pointers in the vector. When it is ready either allocate one mbuf (if it small) and copy the vector of pointers of just provide it as a contiguous buffer (not mbuf) as I suggested 

This is not the common case so there is no need to over optimization

jsmoon:
@hhaim Do you mean that `m_rx_data` type in `CEmulApp` should be changed from `struct rte_mbuf*` to `std::vector<uint8_t>`?
Could you explain your suggestion in detail? Which codes should be changed in my changes?

hhaim:
@jsmoon read again my suggestion in the original thread.
I’ve suggest to add another callback (sappend) to emu layer with *char interface (not vector)

jsmoon:
@hhaim I think you are saying the following comment.

> thinking about it more, it is better to provide a new api for the application layer
sbappend_buffer(*char ,size) as with mbuf we might be limited with retransmition with the buffer size

I thought I've already implemented similar functionality by `CEmulApp::append_rx_data(m)`. Regarding this implementation, I've already commented in the thread.

And I thought [your previous comment](https://github.com/cisco-system-traffic-generator/trex-core/pull/423#issuecomment-581251462) was something like a different one. So, I asked you to explain your thought more specifically in the codes.

hhaim:
@jsmoon I’m saying the same thing.
The previous comment was on the implementation of the merge (vector of char*,size) per section.
After the merge the API should be contiguous memory with sbappend_buffer(*char ,size) to emu layer.

jsmoon:
@hhaim
Do you mean that I should change the mblock in CTcpReass contains the pointer of copied contents from large size mbuf(9k)?

hhaim:
Yes. Each block should have a pointer to vector<string>. It should be enabled by a field in tcp flow(instruction) Once merge is ready, the callback sappend_buffer should be called
Today sappend_bytes is called instead 

You should *not* save mbuf in the ring even though it is faster

jsmoon:
@hhaim I have some questions for you. The `sbappend_buffer(char*, size)` may be called multiple times including in the `sbappend(m)` also.
I implemented to collect the contents in `CEmulApp`. Do you agree with the location?
What do you prefer how to collect the contents in `CEmulApp`? merge them into contiguous? keep them as the vector of pointers of each content?

P.S. If you read my comment, you could know I've already concerned about the long lifetime of mbuf. If you commented on it then, I could avoid using mbuf.

hhaim:
My main point that you can’t save the mbuf in the vector you must copy for later merge.
You *can* call a few times to sappend(m) and merge the buffers to a temporary mbuf.
I would use up to 2k mbuf and *NOT* the 9k -allocate it dynamically from the same pool

jsmoon:
@hhaim What about using **heap memory** instead of 2k mbuf? What's the benefit of using 2k mbuf?
What does "the vector" mean? If you mean it `CTcpReass::m_mblocks`, I'll remove the mbuf from it.

hhaim:
@jsmoon 
Again, I think you might mixing the callback API (to emails layer ) and the reassembly implementation.
The API could be either ```sbappend_buffer(char*, size)``` - one buffer or **multiple** called to ```sbappend(m)``` (the API exists today)

Once the retransmission buffer is ready you should choose either to call to a new API ```sbappend_buffer(char*, size)``` or the old ```sbappend(m)``` both are good. 

If you choose not to add a new API,  (In the later case) ```m``` should be maximum of 2K (there are many blocks in this chunk) for each call. 

The reassembly object should keep the buffers internally in the *heap* (```vector<string>```) and not save the original mbuf 

Let me know if you it is clear, as it seems I repeating myself

hhaim:
to be more specific with code 

instead 
```
class CTcpReassBlock {

public:
  void Dump(FILE *fd);
  bool operator==(CTcpReassBlock& rhs)const;

  uint32_t m_seq;  
  uint32_t m_len;  /* scale option require 32bit maximum adv window */
  uint32_t m_flags; /* only 1 bit is needed, for align*/
};

typedef std::vector<CTcpReassBlock> vec_tcp_reas_t;
```

Add this pointer to vector 

```
class CTcpReassBlock {

public:
  void Dump(FILE *fd);
  bool operator==(CTcpReassBlock& rhs)const;

  uint32_t m_seq;  
  uint32_t m_len;  /* scale option require 32bit maximum adv window */
  uint32_t m_flags; /* only 1 bit is needed, for align*/
class CTcpReassBlock {

public:
  void Dump(FILE *fd);
  bool operator==(CTcpReassBlock& rhs)const;

  uint32_t m_seq;  
  uint32_t m_len;  /* scale option require 32bit maximum adv window */
  uint32_t m_flags; /* only 1 bit is needed, for align*/
  std::vector<string> * m_buffers
};

typedef std::vector<CTcpReassBlock> vec_tcp_reas_t;
```

This vector should be allocated **only** if  tcp object enable this feature

```struct tcpcb *tp->buffer_retransmit =true``` (new  field)

This new field would be controlled from EMU layer, using a command

Once the buffer is ready, 

instead of this 
```
    sbappend_bytes(&tp->m_socket,
                   &tp->m_socket.so_rcv ,
                   m_blocks[0].m_len);
```

call the merge the internal vector to mbuf and call 

```
       for reassembly buffers {
         convert it to mbuf
        sbappend(so,
                 &so->so_rcv, m,ti->ti_len); 
        sorwakeup(so); 
      }
```
My suggestion is to 
1. add a unit-test for the CTcpReassBlock in case of buffer merge is enabled to verify the buffer is correct
2. add a new EMU instruction to compare the buffer and check the new callback to verify the merge is working. change the MTU to be very low (1 could be valid) and make a test with OOO and drop to verify that it works 

once the above unit-tests works we can continue to the real features you wanted to a new instruction 

```rcv_chooose_template```

**Comment by hhaim** [comment]
1. you can't free the mbuf and keep the memory. You need to copy it!
```
   assert(ti);
    assert(m);
    assert(m->pkt_len == ti->ti_len);

    char* buf = utl_rte_pktmbuf_to_mem(m);  <<<<
    rte_pktmbuf_free(m);                                  <<<

    uint16_t tg_id = tp->m_flow->m_tg_id;
    INC_STAT(pctx, tg_id, tcps_rcvoopack);
@@ -214,7 +214,7 @@ int CTcpReass::pre_tcp_reass(CPerProfileCtx * pctx,
        lpb->m_seq = ti->ti_seq;
        lpb->m_len = (uint32_t)ti->ti_len;
        lpb->m_flags = (ti->ti_flags & TH_FIN) ?1:0;
        lpb->m_mbuf = m;
        lpb->m_buf = buf;                   <<<<<<
        m_active_blocks=1;
        return(0);

```
2. The copy and reassembly should be conditioned by the upper layer. By default it should do the same as today. Upper layer should be able to change mode dynamically (from EMU layer from enable/disable)

**Comment by jsmoon** [comment]
@hhaim
1. Please check the utl_rte_pktmbuf_to_mem routine first.
2. For the future use case (dynamically enabled later), I think
pre_tcp_reass should keep all the reass packets.

2020년 2월 4일 (화) 오후 7:03, Hanoh Haim <notifications@github.com>님이 작성:

**Comment by hhaim** [comment]
@jsmoon 

1. You are right. 
2. We should have the option to support both modes (enable/disable) and dynamic move between the modes.  By default it should be *disabled* as it is today. It should be enabled only for specific duration of the flows controlled by EMU. The reason is **resource**. In case of 40 MFlows with some retransmission you can very fast end up without heap memory!! 
why to save memory if it is not needed by EMU layer. 
The code should work as it is today -- just keep a pointer to the buffer. Allocate only when needed

**Comment by jsmoon** [comment]
2. There is no guarantee to get received payloads when it turn enabled
dynamically. So I think it should be handled statically per flow. The
enabled flow can be disabled. But, disabled flow should not be enabled
basically. If you agree with me, I'll change it disabled by default
tomorrow.

2020년 2월 4일 (화) 오후 7:53, Hanoh Haim <notifications@github.com>님이 작성:

**Comment by hhaim** [comment]
@jsmoon It could be changed dynamically. I mean that 
```CTcpReassBlock``` once allocated **will copy** the mode from the flow and behave like it.
Once it started in a specific mode, it could not change. (up to a point that OOO is finished and object released)
For example, if it started as disabled, it could not be changed to enabled. 

Examples:

1. flow start with disabled mode and does not change it -- current 
2. flow start with enabled mode and does not change it -- new 
3. flow start with disabled, OOO happend, Reasembley object started, EMU change it to enable but it won't get the buffer -- OK
4. flow start with disabled, **no OOO**,  EMU change it to enable, OOO happend.  EMU layer will get the data 

thanks
Hanoh

**Comment by jsmoon** [comment]
What's the purpose of enabling by EMU? Example 3 and 4, for the same EMU
requests, the results are different and unpredictable.

2020년 2월 4일 (화) 오후 8:34, Hanoh Haim <notifications@github.com>님이 작성:

**Comment by hhaim** [comment]
@jsmoon you are mixing predictability and resource saving. 
In case EMU changes it from enabled->disable in the middle of the flow and only then the OOO 
start  there could be resource saving  
Your case is similar.  In your case you care about the first request, but next request or data (big URL) is not cared. 

to sum:

enable->disable : will work 
disable->enable: will not always give to EMU layer the full data

just add a enable/disable flag to each ```CTcpReassBlock```.once the array is allocated copy the flow.flag to each block flag. The flag will not change while the object is alive, but could change if the object is flushed and recreated

**Comment by jsmoon** [comment]
@hhaim My question was that what the EMU can do with the collected incomplete data.

> In your case you care about the first request, but next request or data (big URL) is not cared.

The next request in the flow is in the right template identified by the first request, isn't it?

**Comment by hhaim** [comment]
EMU layer will get either length or buffer and the total size will always be coherent.
The only thing is that in transition (enable-disable) you might keep getting buffers instead of bytes. It would be the responsibility of EMU layer to handle it using the current instruction.
(What make me think that it better to keep the old sappend(m) with multiple calls instead of the sappend(buffer)

**Comment by jsmoon** [comment]
@hhaim 
> EMU layer will get either length or buffer and the total size will always be coherent.

I think I'm aware of the incoherent possibility between them when disabled -> enabled.

> The only thing is that in transition (enable-disable) you might keep getting buffers instead of bytes. It would be the responsibility of EMU layer to handle it using the current instruction.

I think my last commit implements this a bit.

> (What make me think that it better to keep the old sappend(m) with multiple calls instead of the sappend(buffer)

I don't understand what you said in this comment. Do you want to remove the `sbappend_buffer`? **What do you want me to do next in this PR?**

**Comment by hhaim** [comment]
@jsmoon let me start by saying this is very delicate code and need a very careful testing and design 

1. It would be very easy to eat all the memory and I don't want to a crash or taking all OS resource. 
 
You should handle the case that malloc return null and **limit** the number of re-transmit flows/total flow memory allocated from the heap for not getting close to this scenario. This is not common case and we should have a guard.

In case we breach the limit (total heap memory or amount of flows) update a new err counter and drop the packet to make the TCP re transmit as if the packet was dropped. 

wrap the malloc/free with a context (ctx) functions for allocating and free flow memory and make sure it is limited. 
 
so both should be added (limit of flow in this state and handle the case that malloc return zero)

*examples:*

This function will crash in case of no memory 
1.1 

```
// this function was written for testing, should handel the case of no memory 
char * utl_rte_pktmbuf_to_mem(struct rte_mbuf *m){
    char * p=(char *)malloc(m->pkt_len);
    char * op=p;
    uint32_t pkt_len=m->pkt_len;
}
```

1.2. 

```
-- need to check that pointer is valid and save null in case it is not 
                    if (m_reass_buf) {
                        char * p = (char *)realloc(tblocks[ci-1].m_buf, tblocks[ci-1].m_len + cur.m_len-q);
                        assert(p);
```

2. 
I would add assert that mbuf len is eq to ti->ti_len. 
Looking into the code the function ``tcp_pktmbuf_fix_mbuf`` already does that  before but just to make sure. 

```
    if (m_active_blocks==0) {
        /* first one - just add it to the list */
        CTcpReassBlock * lpb=&m_blocks[0];
        lpb->m_seq = ti->ti_seq;
        lpb->m_len = (uint32_t)ti->ti_len; <<<
        lpb->m_flags = (ti->ti_flags & TH_FIN) ?1:0;
        lpb->m_buf = buf;
        m_active_blocks=1;
        return(0);
    }
```


3. lazy allocation:
you are always allocating the memory and in some cases free it immediately after. 
Instead keep the convert to char * only when needed and defer the mbuf free 

* alloc *
```
    char* buf = nullptr;
    if (m_reass_buf) {
        buf = utl_rte_pktmbuf_to_mem(m);
    }
    rte_pktmbuf_free(m);
```

example of immediate free:

* free*

```
    if (ci>MAX_TCP_REASS_BLOCKS) {
        /* drop last segment */
        INC_STAT(pctx, tg_id, tcps_rcvoopackdrop);
        INC_STAT_CNT(pctx, tg_id, tcps_rcvoobytesdrop,tblocks[MAX_TCP_REASS_BLOCKS].m_len);
        free(tblocks[MAX_TCP_REASS_BLOCKS].m_buf);  << free the same object that was just allocated 
    }

```


4. This code is completely off. This will reduce the current TCP performance to the floor. 
EMU layer does not need to allocate memory, the TCP layer already does this


```
CEmulApp::append_rx_data
  void append_rx_data(uint8_t* buf, uint32_t len) {
        if (get_rx_enabled() && (get_rx_data_size() < m_rx_data_wm)) {
            uint16_t cp_size = std::min(m_rx_data_wm - get_rx_data_size(), len);
            m_rx_data.insert(m_rx_data.end(), buf, buf + cp_size);
        }
    }

    void clear_rx_data() { m_rx_data.clear(); }
    uint8_t get_rx_byte(uint32_t off) { return m_rx_data[off]; }

```

I suggest to revert  tcp_socket.h/cpp and *keep* the same API as before  

instead of this
```

    sbappend_buffer(&tp->m_socket,
                    &tp->m_socket.so_rcv,
                    m_blocks[0].m_buf,
                    m_blocks[0].m_len);

```

convert the buffer to m_buf and make a multi - call in case the buffer is too long (>2K) 

convert to this
```
 convert buffer to mbufs 
  loop  m in mbufs {
        sbappend(so,
                 &so->so_rcv, m,ti->ti_len); 
        sorwakeup(so); 
  }
```

This will give to EMU layer the same API. 

More advance optimization is to make sure the allocated m are big enough

**Comment by hhaim** [comment]
@jsmoon I don't see how the last commit align with the comments. 
It is the opposite. Could we have a talk? You moved the logic from TCP to EMU which is exactly the opposite from the proposal.

*High level of my proposal*

* EMU layer should be kept the same !
* TCP layer should work as today (REASS_OFF) by default.
* In case configured (by EMU) a flow could move to REASS_ON mode, which save memory and update the EMU in the same callbacks (with mbuf instead of len)
* The TCP REASS allocation size  (in case it is enabled) should be protected. (number of flows/total byte allocation, etc., malloc return null)
* New counters should be added to TCP layer to count the error cases.

**Comment by jsmoon** [comment]
@hhaim 

> EMU layer should be kept the same !

EMU layer is the same as before if setup_rx_data(size) is not used.

> TCP layer should work as today (REASS_OFF) by default.

Yes it is.

> In case configured (by EMU) a flow could move to REASS_ON mode, which save memory and update the EMU in the same callbacks (with mbuf instead of len)

Now, there is no need to copy intermediate REASS buffer or mbuf.

> The TCP REASS allocation size (in case it is enabled) should be protected. (number of flows/total byte allocation, etc., malloc return null)

Now, there is no memory allocation for TCP REASS.

> New counters should be added to TCP layer to count the error cases.

The only case will be disabled -> enabled transition case. It may not work to collect the full data. But you said it's OK. So, I thought that there is no need to keep contents for the future use and there is no complicated mbuf handing in TCP REASS.

**Comment by hhaim** [comment]
I see diff in both 
```
src/44bsd/tcp_socket.cpp
src/44bsd/tcp_socket.h
```
am I missing somthing?
try to revert those changes ( should have *zero* diff). diff only in tcp layer.

**Comment by jsmoon** [comment]
Since I reverted my previous commits and made a new commit. You can find the all the changes in 31a41fe. There is no need to check other previous commits.

**Comment by hhaim** [comment]
Is this diff in your last commit?
```

void CEmulApp::update_rx_data(struct rte_mbuf* m, uint32_t seq, uint16_t len) {
    if (m_rx_data && (m_cmd_rx_bytes < m_rx_data->m_size)) {
        assert(m->nb_segs == 1);
        uint8_t* p = rte_pktmbuf_mtod(m, uint8_t*);
        int32 q = m_rx_data->m_seq + m_rx_data->m_size - seq;
        if (q > 0) {
            memcpy(m_rx_data->m_data + m_rx_data->m_size - q, p, std::min(q, (int32)len));
        }
    }
}
void CEmulApp::update_rx_data(struct rte_mbuf* m) {
    if (m_rx_data) {
        update_rx_data(m, m_rx_data->m_seq, rte_pktmbuf_pkt_len(m));
    }
}

void CEmulApp::clear_rx_data() {
    if (m_rx_data) {
        delete m_rx_data;
        m_rx_data = nullptr;
    }
}

void CEmulApp::setup_rx_data(uint16_t size) {
    clear_rx_data();
    /* enabled only when there is no received bytes in EMUL */
    if (m_flow && m_cmd_rx_bytes == 0) {
        m_rx_data = new CEmulRxData(m_flow->m_tcp.rcv_nxt, size);
    }
}


```

Did you change 
```
src/44bsd/tcp_socket.cpp
src/44bsd/tcp_socket.h
```

files?

**Comment by hhaim** [comment]
>>EMU layer should be kept the same !

This mean zero diff in both ```src/44bsd/tcp_socket.cpp
src/44bsd/tcp_socket.h```

**Comment by jsmoon** [comment]
I understood the same that the behavior is the same. I don't think you meant that don't change the files.
Do you want me the buffer to the tcpcb?

**Comment by hhaim** [comment]
Yes! the app layer (EMU) should not be changed **at all**, **ZERO** changes. 
The all point of TCP is to give the same API to the application. In our case a stream of mbuf. 
In case of REASS_OFF application will get a stream of sizes (only in case of an issue of OOO).  

The protection against malloc issues is a TCP layer responsibility **NOT** the application layer. 
This is how it works in any OS.

**Comment by hhaim** [comment]
@jsmoon 
Not sure I folow the logic, it is not even close to solve the problem. 
The logic should be integrated into the ```CTcpReassBlock``` like you already started to do a few diff back. There shoule be a queue of blocks and merge logic.  

Let's take an example
```
  +   tp->append_rx_data(m, ti->ti_seq, ti->ti_len);
        sbappend(so,
                 &so->so_rcv, m,ti->ti_len); 
        sorwakeup(so); 

```
This code has a valid buffer and it is in order why do you need to add it to ``ip->append_rx_data`` ???
The Reass logic supprt holes and it seems you dulicate it with another algoritem that does not work 

for example you could get:

 100-200, 500-600,  400-410, 200,400 (merge), 410-500(merge-call-sbappend)

* The allocations (``caloc``) are not protected against overuse 

It seems you are repacing diffs (I think I saw this diff already)  without reading the comments.

**Comment by hhaim** [comment]
This is the suggestions, please verify you understand each bullet 

>> *EMU layer should be kept the same !
     * TCP layer should work as today (REASS_OFF) by default.
     * In case configured (by EMU) a flow could move to REASS_ON mode, which save memory and 
    update the EMU in the same callbacks (with mbuf instead of len)
    * The TCP REASS allocation size (in case it is enabled) should be protected. (number of flows/total 
    byte allocation, etc., malloc return null)
    * New counters should be added to TCP layer to count the error cases.

**Comment by jsmoon** [comment]
@hhaim 

> This is the suggestions, please verify you understand each bullet
> 
> *EMU layer should be kept the same !
> * TCP layer should work as today (REASS_OFF) by default.
> * In case configured (by EMU) a flow could move to REASS_ON mode, which save memory and
> update the EMU in the same callbacks (with mbuf instead of len)
> * The TCP REASS allocation size (in case it is enabled) should be protected. (number of flows/total
> byte allocation, etc., malloc return null)
> * New counters should be added to TCP layer to count the error cases.

Please recheck my reply [here](https://github.com/cisco-system-traffic-generator/trex-core/pull/423#issuecomment-582785007) (except first one, we already talked about it) and comment them each like the first one.

> Not sure I folow the logic, it is not even close to solve the problem.

What do you mean "the problem"? I gave up collecting full data after disabled->enabled transition performed as you said it's OK.
 
> The logic should be integrated into the CTcpReassBlock like you already started to do a few diff back.
> There shoule be a queue of blocks and merge logic.

Since `pre_tcp_reass` copies the specified payload from mbuf in the entry, `CTcpReassBlock` does not need to keep mbuf or copied contents any longer. Although there may be duplicated copy in there, I think there is no problem because the duplicated copy will be rare in normal. I don't think this causes serious performance impact.

> This code has a valid buffer and it is in order why do you need to add it to ip->append_rx_data ???
> The Reass logic supprt holes and it seems you dulicate it with another algoritem that does not work

Frankly speaking, I added `append_rx_data()` and `add_rx_bytes()` calls in a hint of statistics `tcps_rcvbyte` update.

> for example you could get:
> 100-200, 500-600, 400-410, 200,400 (merge), 410-500(merge-call-sbappend)

As far as I know, sbappend is not called by it. sbappend_bytes is called. What am I missing?

> •	The allocations (caloc) are not protected against overuse
> It seems you are repacing diffs (I think I saw this diff already) without reading the comments.

I read it. So I decided to remove malloc usage in `CTcpReassBlock`. And because `CTcpRxData` is limited by the request, I thought it was not an issue. If the allocation (CTcpRxData itself or its m_data by calloc) is failed, this feature will not enabled internally.

**Comment by hhaim** [comment]
>>Now, there is no memory allocation for TCP REASS.

You didn’t solve the problem just move it to a different place. You code is a duplication of the logic in the wrong way that I suspect does not work in the best case.

The memory saving should be part of the Reass class and follow the same logic. 
any ‘malloc’ should be protected against overflow. You can’t reach the case that all the system memory is going for REASS 

You can create another class that save the memory instead of just the seq number and when choose the right object in creation time. But it should be the **same** logic for saving the memory and doing the seq Reass

**Comment by egwakim** [comment]
Hi @hhaim 
There are lots of ping pong discussions between @jsmoon .
It's really hard to understand all your wanted ways line by line.
Can you modify jsmoon's PR for your needs? 
It will be more efficient way.

How about your opinion?

Best Regards
Gwangmoon

**Comment by hhaim** [comment]
@egwakim I’m busy with other stuff right now. This is super sensitive code and should be written with care and many unit-tests to verify we are not missing anything.
I see you point that you want to move to your *real* feature but we can’t skip this or to have compromised quality. I’ve suggested to have a short meeting to explain it better.
@jsmoon already written more complex code, so I’m sure he could do this too
Let me know

**Comment by jsmoon** [comment]
@hhaim 

> You didn’t solve the problem just move it to a different place. You code is a duplication of the logic in the wrong way that I suspect does not work in the best case.

This comment doesn't make any help for me. This comment does not contain any information that explains your opinion for "the problem", "duplication", "the wrong way" and "does not work".
I couldn't agree with your opinion without those proper explanations. I cannot implement the way I cannot agree with.

I asked you many questions to get to understand your thought properly. Most of them were ignored and had no answer. So I've implemented it as I understood regarding this issue.

P.S.
I couldn't understand why you are saying TCP_REASS repeatedly. The purpose of this issue is to get the proper L7 payload in the **enabled** state.
The TCP_REASS is just a routine executed only when it is needed. There is no need to keep a copy in the TCP_REASS block because it can be updated directly to the L7 payload buffer. In the **disabled** state, the TCP_REASS block should not keep the copy in the TCP_REASS block as today. So, TCP_REASS does not need to keep any copy of the L7 payload.

**Comment by egwakim** [comment]
Hi @hhaim ,
The last jsmoon's PR was almost same as your idea today.
Can you please review jsmoon's PR again carefully?

Best Regards
Gwangmoon

**Comment by hhaim** [comment]
@egwakim it is very far from it. In my suggestion there is *ZERO* malloc

In his idea, there is
```
class CTcpRxData {
public:
    CTcpRxData(uint32_t seq, uint16_t size) {
        m_size = 0;
        m_bytes = 0;
        if (size) {
            m_data = (uint8_t*)calloc(1, size);   <<<==
        }
        if (m_data) {
            m_seq = seq;
            m_size = size;
        }
    }

```

In my idea, there is a need for a few line of change, no need to maintain anything

**Comment by egwakim** [comment]
Hi,
It's usage is different, it's used for m_app instance. It's unavoidable.

**Comment by hhaim** [comment]
@egwakim I think it is avoidable. 

see this proposal for my change (it is a 3 lines of code)

```
inline void TCP_REASS(CPerProfileCtx * pctx,
                      struct tcpcb *tp,
                      struct tcpiphdr *ti,
                      struct rte_mbuf *m,
                      struct tcp_socket *so,
                      int &tiflags) { 
    if (ti->ti_seq == tp->rcv_nxt && 
       (tcp_reass_is_exists(tp)==false) && 
        tp->t_state == TCPS_ESTABLISHED) { 
        if (tiflags & TH_PUSH) 
            tp->t_flags |= TF_ACKNOW; 
        else 
            tp->t_flags |= TF_DELACK; 
        tp->rcv_nxt += ti->ti_len; 
        tiflags = ti->ti_flags & TH_FIN;
        uint16_t tg_id = tp->m_flow->m_tg_id;
        INC_STAT(pctx, tg_id, tcps_rcvpack);
        INC_STAT_CNT(pctx, tg_id, tcps_rcvbyte,ti->ti_len);
        sbappend(so,
                 &so->so_rcv, m,ti->ti_len); 
        sorwakeup(so); 
    } else { 
        if tp->reass_enabled {                                                                <<<<
        // as if it was droped                                                                   <<<
        INC_STAT(pctx, tg_id, tcps_rcvoopackdrop);                            <<<
        INC_STAT_CNT(pctx, tg_id, tcps_rcvoobytesdrop,ti->ti_len);   <<
        }else{
           tiflags = tcp_reass(pctx,tp, ti, m); 
           tp->t_flags |= TF_ACKNOW;
        } 
    }
}

```


This is the all commit. 
Once this will be committed.  You can add a code that enable ```reass_enabled ``` for flows that does not have a template and then with a new instruction choose the right template

egwakim:
Hi @hhaim ,
We are discussing different discussion point.

In this TCP flow,
tcp_flow_input() +--> (1)no reass required case -> sbappend() --> (3) m_app instance layer
                 +--> (2)reass required case -> tcp_reass() -> sbappend_bytes() --> m_app instance layer

You are talking about the point (1) and (2), but, we are talking about point (3).
In this previous comment, You mentioned that we should not change tcp_socket.cpp and tcp_socket.h code.
But, m_app should check payload in tcp_socket.cpp
To align this rule, we have to allocate buffer and copy some payload content for m_app.
It's unavoidable to prevent changing code of tcp_socket.cpp.
That's the reason to allocate buffer.

Do you have any other idea to check payload in m_app without changing tcp_socket.cpp code?

Best Regards
Gwangmoon

hhaim:
@egwakim you are mixing PR. We should have on PR for making TCP provide the right data for EMU layer (buffer). This is the name of *THIS* PR.
Let’s have a fix and then discuss how to use this data for the original need.
(BTW I don’t think that you need any allocation even in the *next* PR but let’s discuss it then)

This PR is for TCP layer *only*

jsmoon:
@hhaim I'll close this PR because it could not fulfill your thought.

egwakim:
Hi @hhaim ,
We need to clarify the way forward of this PR.
I assume that you want to 3 steps for the final feature as below.

Step#1. Prevent tcp_reass operation to protect memory issue.
Solution : Drop packets in tcp_reass when tp->reass_enabled option disabled.
Just 3 lines enough? 
Then, how about change it in your side?
We'll cancel this PR. After then, we'll focus on step#2 and step#3.

Regarding step#2 and step#3.
Step#2. Pass payload information to m_app emulation layer
Solution : 
Suggestion#1. Current PR#423 - save payload for m_app emulation layer
It's needed because mbuf can be multiple, we need to merge them into single.

Suggestion#2. Any other suggestions?

Step#3. l7_map handling by using passed payload information

Best Regards
Gwangmoon

hhaim:
@egwakim and @jsmoon 
I think we should have at least 2 PR

**provide TCP data to EMU**

*Feature description*:

Today the TCP layer in some cases does not provide the data (char*) to the EMU (app) layer, only the len (number of bytes). We should have an option to change this dynamically using EMU layer API without consuming all the server data 

*Suggestion*: 

1. in case of reass keep the windows very small and assume the packets were dropped if there are no in order. this way we will provide the EMU layer with the data (not size of bytes).
2. add a enable/disable flag into the tcp flow.  (```reass_enabled```)

This could be implemented using the code I've sent  

>> Just 3 lines enough?

**YES**

**multi-port TCP profile**

*Feature description*:
Ability to choose a profile by the l7 data provided by the TCP layer. 

*Suggestion*: 

1. move (```reass_enabled```) to true in case the flow has more than one port. 
2. assign a EMU layer program with a *new* instruction that will get the mbuf and choose by the data the right template 
3. after choosing the right template change the 
  3.1 ```reass_enabled=false```
  3.2 reassign the right EMU layer 
  3.3 update the layer with the number of bytes received from TCP to make it consistent 

4. In this PR there should **not** be **ANY** new malloc/free in application layer. TCP provide the data as mbuf with size, you can use it. 
There is only corner case that TCP will provide byte by byte and we can either have in EMU layer a constant buffer of 8 bytes or not supporting this. Your call (I think we should no support this)

We can  have another meeting if needed.

egwakim:
Hi @hhaim ,
Ok, then, I'll create 2 tickets("Ticket#1. Provide TCP data to EMU", "Ticket#2. multi-port TCP profile") for it.
Can you merge ticket#1 first?
If you merge Ticket#1, then, we'll contribute Ticket#2 on top of merged Ticket#1.

And let's discuss the detailed design in the Ticket#2.

hhaim:
@egwakim could you open a PR for 1 so we can commit it?

Issue #433, opened 2020-02-18
[Suggestion] Increase number of set_var()


egwakim:
To simulate single voice flow,
Series of voice and silent packets loop need to be repeated.
pcap file cannot be used because there are many parameters for it(voice length, packet size, etc), it requires too many .pcap files.

<voice packets><silent packets><voicepackets><silent packets><voice packets>.
To be able to achieve it,
We need at least 3 loops with 3 set_var/jmp_nz,
Overall outer loop(voice traffic loop + silent traffic loop)

Currently, it's not possible because total number of set_var is limited to 2.
src/44bsd/tcp_socket.h
523 class CEmulApp  {
524 public:
525     enum {
526         apVAR_NUM_SIZE =2
527     };
528 

apVAR_NUM_SIZE=2 is too small number to simulate various traffic scenarios.

Can we increase this number to reasonable number(3 or larger number)?
Or
More complex way is limiting concurrent flows by 1 with https://github.com/cisco-system-traffic-generator/trex-core/issues/402

Best Regards
Gwangmoon

hhaim:
@egwakim I would increase to 5

PR #437, opened 2020-02-20
Increase number of set_var()

yonghwan007:
apVAR_NUM_SIZE=2 is too small number to simulate various traffic scenarios.
This issue was already discussed in https://github.com/cisco-system-traffic-generator/trex-core/issues/433, so I increased apVAR_NUM_SIZE=5


hhaim:
rebased.

Issue #429, opened 2020-02-13
Multi-port TCP profile


egwakim:
Multi-port TCP profile

Feature description:
Ability to choose a profile by the l7 data provided by the TCP layer.

Suggestion:

move (reass_enabled) to true in case the flow has more than one port.

assign a EMU layer program with a new instruction that will get the mbuf and choose by the data the right template

after choosing the right template change the
3.1 reass_enabled=false
3.2 reassign the right EMU layer
3.3 update the layer with the number of bytes received from TCP to make it consistent

In this PR there should not be ANY new malloc/free in application layer. TCP provide the data as mbuf with size, you can use it.
There is only corner case that TCP will provide byte by byte and we can either have in EMU layer a constant buffer of 8 bytes or not supporting this. 

Related PR :
https://github.com/cisco-system-traffic-generator/trex-core/pull/423

Related Issues :
https://github.com/cisco-system-traffic-generator/trex-core/issues/381

egwakim:
Hi @hhaim ,
Before implementation, we have several important design options to clarify.
Can you comment your opinion about the design options?

1. New recvcmp() instruction or re-using l7_map() API
l7_map() can map payload contents with mask and it's already merged into UDP case.
But, suggested recvcmp() have an issue when mapped payload is not located at the beginning of the payload. 
User may call recvcmp() after recv() call, it requires too complex handling logic in trex server side.
I cannot see any reason to use recvcmp() instead of l7_map() API.
Can you confirm it whether recvcmp() is needed or not?

2. Support of dynamic Enable/Disable of l7_map()
Do we need to support dynamic on/off of l7_map?
If we support it, It'll make tcpcb::m_reass_disabled dynamically at run-time.
I assume that it's not needed in general.

3. Support of smaller payload transmission for l7_map()
It's quite corner situation, 
If user send smaller packet which will not contain payload in the single packet.
It's waste of efforts to support such case.
Do we need to support such corner case?

Best Regards
Gwangmoon

hhaim:
@egwakim very good summary. Let me refine it. 

1.  flow table (the same place you did the UDP)

```
if flow_table.lookup (SYN) != nil {
    // if the dstport has more than one possible template 
   if ( more_than_one_port_map[pkt.dst_port]  == true ) {

      flow.m_reass_disabled = true // disable reassembly to get all the data in order 
      flow.EMU = a_program_with_a_one_instruction_recvcmp_choose_template()
    }
}

```

[...]

PR #438, opened 2020-02-21
Added TCP port sharing support by payload

jsmoon:
Hi, this PR is the first implementation of issue #429.

Please check the changes and give your feedback.

### DIFF
--- a/scripts/astf/shared_port.py
+++ b/scripts/astf/shared_port.py
@@ -1,0 +1,119 @@
+# Example for creating your program by specifying buffers to send, without relaying on pcap file
+
+from trex.astf.api import *
+
+
+# we can send either Python bytes type as below:
+http_req = b'GET /3384 HTTP/1.1\r\nHost: 22.0.0.3\r\nConnection: Keep-Alive\r\nUser-Agent: Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1; SV1; .NET CLR 1.1.4322; .NET CLR 2.0.50727)\r\nAccept: */*\r\nAccept-Language: en-us\r\nAccept-Encoding: gzip, deflate, compress\r\n\r\n'
+http_req1 = b'POST /3384 HTTP/1.1\r\nHost: 22.0.0.3\r\nConnection: Keep-Alive\r\nUser-Agent: Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1; SV1; .NET CLR 1.1.4322; .NET CLR 2.0.50727)\r\nAccept: */*\r\nAccept-Language: en-us\r\nAccept-Encoding: gzip, deflate, compress\r\n\r\n'
+http_req2 = b'PUT /3384 HTTP/1.1\r\nHost: 22.0.0.3\r\nConnection: Keep-Alive\r\nUser-Agent: Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1; SV1; .NET CLR 1.1.4322; .NET CLR 2.0.50727)\r\nAccept: */*\r\nAccept-Language: en-us\r\nAccept-Encoding: gzip, deflate, compress\r\n\r\n'
+# or we can send Python string containing ascii chars, as below:
+http_response = 'HTTP/1.1 200 OK\r\nServer: Microsoft-IIS/6.0\r\nContent-Type: text/html\r\nContent-Length: 32000\r\n\r\n<html><pre>**********</pre></html>'
+
+class Prof1():
+    def __init__(self):
+        pass  # tunables
+
+    def _udp_client_prog(self, req, res):
+        prog_c = ASTFProgram(stream=False)
+        prog_c.send_msg(req)
+        prog_c.recv_msg(1)
+        return prog_c
+
+    def _udp_server_prog(self, req, res):
+        prog_s = ASTFProgram(stream=False)
+        prog_s.recv_msg(1)
+        prog_s.send_msg(res)
+        return prog_s
+
+    def _tcp_client_prog(self, req, res):
+        prog_c = ASTFProgram(stream=True)
+        prog_c.send(req)
+        prog_c.recv(len(res))
+        return prog_c
+
+    def _tcp_server_prog(self, req, res):
+        prog_s = ASTFProgram(stream=True)
+        prog_s.recv(len(req))
+        prog_s.send(res)
+        return prog_s
+
+    def create_profile(self, proto='all', temp='all'):
+        # UDP client and server commands
+        udp_prog_c = self._udp_client_prog(http_req, http_response)
+        udp_prog_c1 = self._udp_client_prog(http_req1, http_response)
+        udp_prog_c2 = self._udp_client_prog(http_req2, http_response)
+
+        udp_prog_s = self._udp_server_prog(http_req, http_response)
+
+        # TCP client and server commands
+        tcp_prog_c = self._tcp_client_prog(http_req, http_response)
+        tcp_prog_c1 = self._tcp_client_prog(http_req1, http_response)
+        tcp_prog_c2 = self._tcp_client_prog(http_req2, http_response)
+
+        tcp_prog_s = self._tcp_server_prog(http_req, http_response)
+        tcp_prog_s1 = self._tcp_server_prog(http_req1, http_response)
+        tcp_prog_s2 = self._tcp_server_prog(http_req2, http_response)
+
+        # ip generator
+        ip_gen_c = ASTFIPGenDist(ip_range=["16.0.0.0", "16.0.0.255"], distribution="seq")
+        ip_gen_s = ASTFIPGenDist(ip_range=["48.0.0.0", "48.0.255.255"], distribution="seq")
+        ip_gen = ASTFIPGen(glob=ASTFIPGenGlobal(ip_offset="1.0.0.0"),
+                           dist_client=ip_gen_c,
+                           dist_server=ip_gen_s)
+
+        # use default port 80
+        assoc_by_ip = ASTFAssociationRule(ip_start="48.0.0.0", ip_end="48.0.255.255")
+        assoc_by_l7 = ASTFAssociationRule(ip_start="48.0.0.0", ip_end="48.0.255.255",
+                                          l7_map={"offset":[0,1,2,3]})
+
+        templates = []
+        # UDP templates
+        udp_temp_c = ASTFTCPClientTemplate(program=udp_prog_c, ip_gen=ip_gen)
+        udp_temp_s = ASTFTCPServerTemplate(program=udp_prog_s, assoc=assoc_by_ip)
+        udp_template = ASTFTemplate(client_template=udp_temp_c, server_template=udp_temp_s, tg_name='udp')
+
+        udp_temp_s1 = ASTFTCPServerTemplate(program=udp_prog_s, assoc=assoc_by_l7)
+        udp_temp_c1 = ASTFTCPClientTemplate(program=udp_prog_c1, ip_gen=ip_gen)
+        udp_template1 = ASTFTemplate(client_template=udp_temp_c1, server_template=udp_temp_s1, tg_name='udp1')
+
+        udp_temp_c2 = ASTFTCPClientTemplate(program=udp_prog_c2, ip_gen=ip_gen)
+        udp_template2 = ASTFTemplate(client_template=udp_temp_c2, server_template=udp_temp_s1, tg_name='udp2')
+
+        if proto == 'all' or proto == 'udp':
+            if temp == 'all' or temp == 'ip':
+                templates += [ udp_template ]
+            if temp == 'all' or temp == 'payload':
+                templates += [ udp_template1, udp_template2 ]
+
+        # TCP templates
+        tcp_temp_c = ASTFTCPClientTemplate(program=tcp_prog_c, ip_gen=ip_gen)
+        tcp_temp_s = ASTFTCPServerTemplate(program=tcp_prog_s, assoc=assoc_by_ip)
+        tcp_template = ASTFTemplate(client_template=tcp_temp_c, server_template=tcp_temp_s, tg_name='tcp')
+
+        tcp_temp_c1 = ASTFTCPClientTemplate(program=tcp_prog_c1, ip_gen=ip_gen)
+        tcp_temp_s1 = ASTFTCPServerTemplate(program=tcp_prog_s1, assoc=assoc_by_l7)
+        tcp_template1 = ASTFTemplate(client_template=tcp_temp_c1, server_template=tcp_temp_s1, tg_name='tcp1')
+
+        tcp_temp_c2 = ASTFTCPClientTemplate(program=tcp_prog_c2, ip_gen=ip_gen)
+        tcp_temp_s2 = ASTFTCPServerTemplate(program=tcp_prog_s2, assoc=assoc_by_l7)
+        tcp_template2 = ASTFTemplate(client_template=tcp_temp_c2, server_template=tcp_temp_s2, tg_name='tcp2')
+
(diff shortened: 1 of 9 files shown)


jsmoon:
@hhaim Could you give your comment on the failure also?
I've changed ASTF related code only. But the failure happened at stl_performance_test.

bdollma:
It is a performance issue. The test values (of the test that fails) do not consider for a safe margin, hence once in a while it fails. Should run again.

hhaim:
@jsmoon looks much better. There is no allocation of data memory now. 

Minor comments/questions 

1.  Better not to allocate and free a context ```auto pctx = new CPerProfileCtx();``` for every deferred flow. Could we have one ProfileCtx for all the templates that match the same port and create one program to **choose** the right template. CPerProfileCtx could be taken from the map_l7() per port 

```
CPerProfileCtx* CTcpFlow::create_on_flow_profile() {
    try {
        auto pctx = new CPerProfileCtx();
        pctx->set_on_flow();

        pctx->m_profile_id = m_pctx->m_profile_id;
        pctx->m_ctx = m_pctx->m_ctx;
        pctx->m_template_rw = m_pctx->m_template_rw;

        return pctx;
    } catch (...) {
        return nullptr;
    }
}```

2.  

so this code 
```
    if (likely(temp->has_payload_params())) {
        lptflow->start_identifying_template(pctx);
    } else {
        app->start(true); /* start the application */
    }
 ``   
would be as there is a program 

```
        app->start(true);   /*the program with the new instruction */
```

3. 
```
int CEmulApp::on_bh_rx_bytes(uint32_t rx_bytes,
                            struct rte_mbuf * m){
    set_interrupt(true);
    /* for now do nothing with the mbuf */
    if (m) {
        check_l7_data(m);
    }

```

You are hiding the mbuf free inside the ```check_l7_data``` better to defer the free in this function and check_l7_data should be part of a new instruction. 

4. add a new instruction 

```
 typedef enum { tcTX_BUFFER             =1,   /* send buffer of   CMbufBuffer */
               tcDELAY                 =2,   /* delay some time */
               tcRX_BUFFER             =3,   /* enable/disable Rx counters */
               tcRESET                 =4,   /* explicit reset */
               tcDONT_CLOSE            =5,   /* wait for reset */
               tcCONNECT_WAIT          =6,   /* wait for the connection, should be the first command */
               tcDELAY_RAND            =7,   /* delay random min-max*/
               tcSET_VAR               =8,   /* set value to var */
               tcJMPNZ                 =9,   /* jump in case var in not zero */           
               tcTX_PKT                =10,  /* UDP send packet */ 
               tcRX_PKT                =11,  /* check # pkts */
               tcKEEPALIVE             =12,  /* set keep alive */
               tcCLOSE_PKT             =13,  /* close connection for udp */ 
               tcTX_MODE               =14,  /* TCP send mode, block, none block */ 
               tcSET_TICK_VAR          =15,  /*  set tick var, used with jmp_dp */
               tcJMPDP                 =16,  /* jump in case time passed from var is less than some duration */
               tcNO_CMD                =255  /* explicit reset */

```
tcRX_BUFFER_CHOOSE_L7 that will call ```check_l7_data``` and save the number of bytes recived to *replay* it later 

5. 

```
void CEmulApp::check_l7_data(struct rte_mbuf * m) {
    if (unlikely(m_l7check_enable)) {
        uint8_t* l7_data = rte_pktmbuf_mtod(m, uint8_t*);
        uint16_t l7_len = rte_pktmbuf_data_len(m);

        if (!m_flow->check_template_assoc_by_l7_data(l7_data, l7_len)) {
            return; // should keep mbuf for the later use at tcp_respond_rst().
        }
    }
    rte_pktmbuf_free(m);
}
```

* Again remove rte_pktmbuf_free from this function 
* Call this as part of a new instruction 
* Save aside the number of bytes recived to in the old program  ```CEmulApp::start(bool interrupt)``` will replay the number of bytes recived you would be able to replay it


6. the defer code that will update the new template

```
void HOT_FUNC CFlowTable::process_tcp_packet(CTcpPerThreadCtx * ctx,
                                    CTcpFlow *  flow,
                                    struct rte_mbuf * mbuf,
                                    TCPHeader    * lpTcp,
                                    CFlowKeyFullTuple &ftuple){

    tcp_flow_input(flow->m_pctx,
                   &flow->m_tcp,
                   mbuf,
                   lpTcp,
                   ftuple.m_l7_offset,
                   ftuple.m_l7_total_len
                   );

    flow->check_defer_function();

    if (unlikely(!flow->is_activated() && !flow->is_identifying_template())) {
        /* identifying template has done */
        if (flow->m_template_info == nullptr) {
            /* no matched template found, mbuf should be freed */
            if (ctx->tcp_blackhole == 0) {
                tcp_respond_rst(flow, lpTcp, ftuple);
            }
            rte_pktmbuf_free(mbuf);
            FT_INC_SCNT(m_err_no_template);
            handle_close(ctx,flow,true);
            return;
        }
        flow->update_new_template_assoc_info();
        flow->m_app.start(false);
        assert(flow->is_activated()); /* flow is now activated */
    }

    if (flow->is_can_close()) {
        handle_close(ctx,flow,true);
    }
}
```

better to move this to a COLD function 

instead of 
```
    if (unlikely(!flow->is_activated() && !flow->is_identifying_template())) {
        /* identifying template has done */
        if (flow->m_template_info == nullptr) {
            /* no matched template found, mbuf should be freed */
            if (ctx->tcp_blackhole == 0) {
                tcp_respond_rst(flow, lpTcp, ftuple);
            }
            rte_pktmbuf_free(mbuf);
            FT_INC_SCNT(m_err_no_template);
            handle_close(ctx,flow,true);
            return;
        }
        flow->update_new_template_assoc_info();
        flow->m_app.start(false);
        assert(flow->is_activated()); /* flow is now activated */
    }
```


do this:
```
    if (flow->is_identifying_template()) {
       if !flow->update_template() {
          return 
       }
    }
```


and 

```
  flow->update_new_template_assoc_info();
  flow->m_app.start(false);
```

should be with the number of bytes received already 

```
  flow->update_new_template_assoc_info();
  flow->m_app.start(false);
  flow->m_app.replay_number_bytes_saved() <<<<<< this trigger as if number of bytes recived. the number of bytes should be saved in the app 

```

jsmoon:
@hhaim 

1. Since l7_map per port could be assigned by multiple templates, the statistics should be collected independently before the right template is chosen.

2~6. When I considered the implementation with the new instruction, I could not find the simple way to defer the update_template and sending RST when no template found finally.
Since I thought you agreed the implementation without the new instruction,

> Another approach that might be simpler is to make a function in EMU class that should call associate_template_by_l7 in case it wasn't already set without an explicit instruction

I've implemented the `check_l7_data(m)` which is trying to process the mbuf by appropriate way I think.

6. I agreed to move it to COLD function. But I don't agree to call new replay_number_bytes_saved() function. It is because `m_app.start(false);` already started to process the received number of bytes.

hhaim:
@jsmoon 

1. The statistics that should be saved is **simple** there is no need to allocate context and free it for one packet.

We can keep the information simple in this case (not all the TCP counters). it is one packet and a few bytes that need to be updated (in case of an error we could keep the counters in the parent)

2. I agree that you could have the code hacked, but don't free the mbuf inside the function  ```check_l7_data``. Now that I see the code I think that adding a new instruction could be cleaner 

6. Are you starting a program with the old rx-bytes information? Did you verify that it works and has the same number of bytes. 

I suggest  to do this test:

1. hack the code to force it to defer the template even with **one** template - the L7 will check if there is one and choose it. Then run the functional test to verify that everything is working

jsmoon:
@hhaim 

1. I don't think your idea can keep the proper tg_id statistics counter. Without hack the counters per flow, I cannot update 

2. I don't know why ```check_l7_data``` should not free the mbuf. What's the difference from freeing the mbuf in ```on_bh_rx_bytes()```? If I don't handle the mbuf in ```check_l7_data```, I should write the code like this (to keep the mbuf when there is no template found):
```
    if (m && check_l7_data(m)) {
        rte_pktmbuf_free(m);
    }
```
Do you want this?

3. Yes, the chosen template's program will be started with received payload within TCP layer. In EMU layer, rx_bytes is 0 before it started.

> I suggest to do this test:
> 
> hack the code to force it to defer the template even with one template - the L7 will check if there is one and choose it. Then run the functional test to verify that everything is working

I've added astf/shared_port.py and checked this features are working properly. I think it could cover your request. (I started with -t proto=tcp,temp=payload and checked the all tg statistics)

hhaim:
@jsmoon 

**1**

In regards to (1) there is a **heavy** performance impact due to the template defer and to keep all the counters in sync. You are right. 

  
1. Alloc/free of a context 
2. Copy all the TCP/UDP counters (there are ~ 50-100 64bit  counters + many new/free)
3. Copy the tpid TCP/UDP counters (there are ~ 50-100 64bit counters + many new/free) 

This is for every new flow in the first packet!!

```
void CPerProfileCtx::update_profile_stats(CPerProfileCtx* pctx) {
void CPerProfileCtx::update_tg_id_stats(uint16_t tg_id, CPerProfileCtx* pctx, uint16_t in_tg_id) 
```    

I don't think we can allow this impact. I suggest to keep all (2) and (3) counters in this case on the default context and update a few main counters (connect,disconnect, bytes) 

**2**

```
int CEmulApp::on_bh_rx_bytes(uint32_t rx_bytes,
                            struct rte_mbuf * m){
    set_interrupt(true);
    /* for now do nothing with the mbuf */
    if (m) {
        check_l7_data(m);
    }

    if ( get_rx_enabled() ) {

        /* drain the bytes from the queue */
        m_cmd_rx_bytes+= m_api->rx_drain(m_flow); 
        if (m_state==te_WAIT_RX) {
            check_rx_condition();
        }
    }

    set_interrupt(false);
    if (m) {   <<<<<<<<<<<<< move the free to here.the mbuf is not l7 
       rte_pktmbuf_free(m);
   }
    return(0);
}
```

**3** 
One test is **not** enough for this complex code. My suggestion is to change the code in a way to force the defer code to work even with one template and run all the current test. This only for testing (the code should be reverted **force**).  Everything should work as before except the counters we per torpid until the first packet 


**4** 
I would add TCP counter for the case (defer template)

egwakim:
Hi @hhaim ,
Regarding 1.
Yes, it have performance impact, but, it's impact will be limited when l7_map+tcp+huge number of flows case. It's not our usage.
To be able to cover your suggestion, broad range of code change required because there are lot's of reference in pctx like INC_STAT(pctx ...). 
It's not possible to change it in short-term period because it need to be handled carefully.
I do not want to use internal temporary version to meet the schedule. I  want to use same open source TRex version.

My suggestion is
- Create new ticket for this improvement of item 1.  It'll be handled later.
   (More discussion will be needed for it)
- Merge current structure with suggested item 2,3 and 4.

Is it acceptable for you?

jsmoon:
@hhaim 

I've tested as you suggested.
As you know, there are two kind of fail cases on your suggestion.
1) test_astf_caps and test_astf_cc_caps functional tests cannot be passed due to different caps generated.
2) test case without tx in client program is hanging. (sfr.py, sfr_full.py)
Except them, there is no additional issue found in my test environment.

hhaim:
> Hi @hhaim ,
> Regarding 1.
> Yes, it have performance impact, but, it's impact will be limited when l7_map+tcp+huge number of flows case. It's not our usage.
> To be able to cover your suggestion, broad range of code change required because there are lot's of reference in pctx like INC_STAT(pctx ...).
> It's not possible to change it in short-term period because it need to be handled carefully.
> I do not want to use internal temporary version to meet the schedule. I want to use same open source TRex version.
> 
> My suggestion is
> 
> * Create new ticket for this improvement of item 1.  It'll be handled later.
>   (More discussion will be needed for it)
> * Merge current structure with suggested item 2,3 and 4.
> 
> Is it acceptable for you?

Yes, it is acceptable. My suggestion is to keep the default context and update only the important counters. There is **no** need to chante ```INC_STAT(pctx ...).```

One more functional comments 

1. Add counter for cases that template wasn't found   see below  

2. 

````
void CEmulApp::check_l7_data(struct rte_mbuf * m) {
    if (unlikely(m_l7check_enable)) {
        uint8_t* l7_data = rte_pktmbuf_mtod(m, uint8_t*);
        uint16_t l7_len = rte_pktmbuf_data_len(m);

        if (!m_flow->check_template_assoc_by_l7_data(l7_data, l7_len)) {
            return; // should keep mbuf for the later use at tcp_respond_rst().
        }
    }
}
```


could be 

```
void CEmulApp::check_l7_data(struct rte_mbuf * m) {
    if (unlikely(m_l7check_enable)) {
        uint8_t* l7_data = rte_pktmbuf_mtod(m, uint8_t*);
        uint16_t l7_len = rte_pktmbuf_data_len(m);
        im_flow->check_template_assoc_by_l7_data(l7_data, l7_len)
    }
}
```


3.  In case there is a **faliure** to find a template in the first case, you should **mark** it and 

this code 
````
if (temp) {
        CPerProfileCtx* pctx = temp->get_profile_ctx();
        if (pctx->is_active() || !pctx->get_nc()) {
            m_template_info = temp;
            return true;
        }
```


````
if (temp) {
        CPerProfileCtx* pctx = temp->get_profile_ctx();
        if (pctx->is_active() || !pctx->get_nc()) {
            m_template_info = temp;
            return true;
   }
m_template_info_not_found = true  <<< mark it 

```


and in this function 

```
bool CFlowTable::update_new_template(CTcpPerThreadCtx * ctx,
                                    CTcpFlow *  flow,
                                    TCPHeader    * lpTcp,
                                    CFlowKeyFullTuple &ftuple){
    assert(!flow->is_activated());
    if   ( flow->m_template_info_not_found ){   <<<<<< check if we got a packet with a payload 
        if (ctx->tcp_blackhole == 0) {
            tcp_respond_rst(flow, lpTcp, ftuple);
        }
        FT_INC_SCNT(m_err_no_template);  <<< rename it to defer m_err_defer_no_template
        handle_close(ctx,flow,true);
        return false;   /* flow is not available */
}

    if (flow->m_template_info) {
        flow->update_new_template_assoc_info();
        flow->m_app.start(false);
        assert(flow->is_activated()); /* flow is now activated */
    } else {    /* no matched template found, now mbuf should be freed */
       assert()
    }
    return true;
}

```

The difference  betwean the implementaion is this:
You are assuming that you will get the first packet with a buffer. but there could be cases that you got just an ACK.

jsmoon:
@hhaim please read my code more carefully. Ack only case also would be handled as today.
As you mentioned, I forgot removal the unnecessary return. But it's the same result.
In my code, m_template_info itself contains the result.

hhaim:
>> Ack only case also would be handled as today." 

I don't see how you have marked that you already seen a payload (called ``check_l7_data``) and failed ? there is no marker there 

Could you elaborate? 

The following  code is called for every packet (payload  OOO or ACK) you should validate  that ``check_l7_data`` was called and failed  ( ```m_template_info==nil```)

```
flow->check_defer_function();


    if (unlikely(flow->new_template_identified())) {
        /* identifying new template has been done */
        if (!update_new_template(ctx, flow, &tcphdr, ftuple)) {
            return; /* when update is failed, flow is not available also. */
        }
    }
```

jsmoon:
@hhaim 
I am using **m_app.m_l7chk_enable** (which is controlled by m_app.set_l7_check() method) during received payload checking enabled.
Deferring template can be checked by m_pctx->is_on_flow(). `start_identifying_template()` set it to flow and `m_l7chk_enable` to activate `check_l7_data()`.
Once `check_template_assoc_by_l7_data()` is called, the `check_l7_data()` will be deactivated by clearing `m_l7chk_enable` in there.
If ACK is received only, `check_l7_data()` won't be called. It means `flow->new_template_identified()` is false in this case.

hhaim:
@jsmoon 
Understood. Thanks for the explanation. nice work.  
Let me run our regression on this.

hhaim:
rebased.

PR #445, opened 2020-03-25
rx_stats object added in example

itsstilldre:
Change-Id: I6c7bb61d93e03e12366d4ced3a7253b492927785


itsstilldre:
Hi, this pull request is just to make contact with the author of this documentation. My issue here is that I would like to use the rx_stats, and the get_stream_stats RCP method of T-Rex server, but seems liek the get_stream_stats is not found among the supported commands. Also this example of rx_stats is missing the point: the rx_stats object and the valid response from the server.

Can someone tell me where can I find a good documentation that my team can use? we would like to use the T-Rex traffic generator, using a java SDK. Unfortunately the currently available java SDK is not complete as for our needs, we would like to extend it, but we do not have good source of information about how to use the T-Rex server. Can anyone help me out?

hhaim:
@itsstilldre the API doc was written when we started the project. We didn't update the doc but it can give the idea how things works. the best is to use the Pthon API and dump the JSON-RPC (or use the Console with -v)

hhaim:
rebased

PR #447, opened 2020-03-26
fixed TRexTimeoutError issue in astf_client.py

ejangle:
When Timeout has occurred in wait_on_traffic,
Exception is NameError, not TRexTimeoutError


hhaim:
@ejangle thanks for the commit. 
Ignore the error, it is in the main branch. we are fixing it

hhaim:
rebased

PR #452, opened 2020-04-06
Connection tracking using 5-tuple

klement:
Use 5-tuple instead of 3-tuple for connection tracking to support NAT
with port overloading.

Signed-off-by: Klement Sekera <ksekera@cisco.com>

### DIFF
--- a/src/44bsd/flow_table.cpp
+++ b/src/44bsd/flow_table.cpp
@@ -59,8 +59,10 @@ void CSttFlowTableStats::Dump(FILE *fd){
 
 
 void CFlowKeyTuple::dump(FILE *fd){
-    fprintf(fd,"m_ip       : %lu \n",(ulong)get_ip());
-    fprintf(fd,"m_port     : %lu \n",(ulong)get_port());
+    fprintf(fd,"m_src_ip   : %lu \n",(ulong)get_src_ip());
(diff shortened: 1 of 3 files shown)


hhaim:
@klement are you planing to solve the simulator issues?

klement:
@hhaim I cannot move forward with it because I have no idea how it works, there is no documentation, example block which I don't understand is

`  1297 TEST_F(gt_tcp, tst33) {                                                         
  1298     CFlowKeyTuple key;                                                          
  1299     key.dump(stdout);                                                           
>>1300     key.set_ip(0x12345678);                                                     
>>1301     key.set_port(1025);                                                         
  1302     key.set_proto(06);                                                          
  1303     key.set_ipv4(true);                                                         
  1304     key.dump(stdout);                                                           
>>1305     printf(" %lx \n",(ulong)key.get_as_uint64());                               
  1306 }   `

marked with >> are errors. what does this test do? old code takes just ip and port, but now we have src and dest, what values would be appropriate?

hhaim:
@klement it is just a unit-test. I would remove this as it is not relevant anymore

klement:
@hhaim are all TEST_F things with errors to be removed as well?

hhaim:
@klement only the one that is not relevant. 
You could ask

hhaim:
@klement I've fixed the simulation issues. thanks for this
rebased

PR #454, opened 2020-04-08
remove STLFlowLatencyStats seq difference limit value 100000

jsmoon:
Hi, this PR will remove the limit(100000) of the sequence difference.

Currently, if received pkt_seq difference from exp_seq is over 100000, the latency stats will increase dropped counter even though all the packets are received successfully.
So, I'd like to remove the limitation and the maximum supported difference value will be INT32_MAX in nature.

In addition, to pass the test of test_python_tabs, I removed a tab and some trailing spaces from trex_emu_profile.py

### DIFF
--- a/src/stx/common/trex_latency_counters.cpp
+++ b/src/stx/common/trex_latency_counters.cpp
@@ -268,7 +268,7 @@ RXLatency::check_seq_number_and_update_stats(
     uint32_t pkt_seq = fsp_head->seq;
     uint32_t exp_seq = curr_rfc2544->get_seq();
     if (unlikely(pkt_seq != exp_seq)) {
-        if (pkt_seq < exp_seq) {
+        if ((int32_t)(pkt_seq - exp_seq) < 0) {
             handle_seq_number_smaller_than_expected(
                 curr_rfc2544, pkt_seq, exp_seq);
         } else {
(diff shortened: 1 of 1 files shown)


jsmoon:
@hhaim Please check my changes and give your feedback.

hhaim:
@jsmoon what is the use-case for this? in other words, in which scenario you should have such a big dog in sac number?

jsmoon:
My user used some value lists for thousand IPs and ports in a stream. SUT handles burst packets per flow. So there are a bunch of ooo packets and the sequence number difference is easily over the limit 100000.

hhaim:
@jsmoon why you changed EMU profile python file? It does not seem to be relevant to this commit.

jsmoon:
@hhaim When I performed `run-regression --func`, its result was failed. So I worried about failing when I raised this PR.
Now I can see master already fixed the file. I'll remove it and update this PR again by force.

hhaim:
@jsmoon looks good

jsmoon:
@hhaim I think the failure is not related to my change. Should I trigger the test again?

jsmoon:
@hhaim Would you mind if I ask when you can process my PRs?

hhaim:
@jsmoon working on it. back from holiday

hhaim:
rebased.

PR #455, opened 2020-04-10
remove ASTF DP profile ctx safely

jsmoon:
Hi,
I found that ASTF DP profile CPerProfileCtx entry had remained in some cases.
So, I added code to remove it when the related ASTF DB is cleared.

And I fixed an issue that CGenNodeTXFIF can reference already removed CPerProfileCtx.

@hhaim, please check my code and give your feedback.


hhaim:
@jsmoon I wonder how we didn't find crash in our regression

hhaim:
rebased.

PR #456, opened 2020-04-10
update L7_map template reference when the profile removed

jsmoon:
Hi,
ASTF TCP L7_map used template creates an interim profile ctx which refers to one of the possible profile's resources. It caused a crash after the reference profile removed.
So, I improved the way to select the referenced profile with updating when insertion and removal of server port information happens. When the profile removed, all the flow's interim profile will be updated to proper another one.


hhaim:
rebased.

PR #457, opened 2020-04-10
Added mutex lock for CAstfDB instance access

jsmoon:
Hi,
I've experienced that CAstfDB::m_instances access is not safe between ASTF CP and DP.
In some cases, m_instances.find() failed although the element exist in it.
So, I simply added a mutex lock to access the CAstfDB::m_instance container exclusively.

@hhaim, Please check my changes and give your feedback.

### DIFF
--- a/src/astf/astf_db.cpp
+++ b/src/astf/astf_db.cpp
@@ -39,9 +39,11 @@ inline std::string methodName(const std::string& prettyFunction)
 astf_db_map_t CAstfDB::m_instances;
 
 
-CAstfDB::CAstfDB(){
+void CAstfDB::Create(){
+    if (m_validator && m_topo_mngr) {
+        return;
+    }
     m_client_config_info=0;
-    m_validator=0;
     m_validator = new CAstfJsonValidator();
     if (!m_validator->Create("astf_schema.json")) {
         printf("Could not create ASTF validator using file astf_schema.json\n");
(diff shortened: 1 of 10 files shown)


hhaim:
@jsmoon in general to use mutex in the DP is not a good thing. but in this case it is only @ startup of a new profile, so I don't see an issue

hhaim:
@jsmoon I've started a full regression to verify that there is no performance impact

hhaim:
There is a significant performance degradation in both ASTF and STL 
image

I don't know why, might be because you change the hotspot or maybe a diff reason 

Can't you sync the CP and DP in a different  way, using a message. Only CP add and DP finds only after the CP add/remove. Could you explain why there is a race. 
Another thing in case it is a must, you can replace the mutex with spin-lock which is much lighter

jsmoon:
I simply changed from mutex lock to spinlock.

hhaim:
@jsmoon could you check if we can synchronize adding new profile to the DP usage (which do you need the lock from the first place). Can we change the map only in CP while we sure the DP does not change it? 
adding lock is like a workaround for a design issue. Why we have a race from the first place? could you explain

jsmoon:
My user has created lots of profiles and cleared them extremely and dynamically. The profile's lifetime is short. I think that the possibility of the race was increased in this situation.
He got the crash in CAstfDB::free_instance() and CAstfDB::instance() a few times.
According to my investigation, CAstfDB::m_instances.find(profile_id) did not find the element but I can see the element in gdb. So, I realized there is racing between CP and DP.

> could you check if we can synchronize adding new profile to the DP usage (which do you need the lock from the first place). Can we change the map only in CP while we sure the DP does not change it?

I'll try your idea next week.

jsmoon:
@hhaim I've changed to remove ASTF DB map access in DP side.
But I've left the change of ASTF DB contents in DP side. I thought it could be quite a large scale refactoring required.
Please check my latest commit and give your feedback.

hhaim:
@jsmoon very nice.
CP already knowS the state of each DP and with the new PR there is no need to sync the map.

hhaim:
rebased

elados93:
**Reopening this issue following a bug**

- I have reverted 6b486ff7c6888824f0cd807564961170229f6d60

- Failure on test: "test_astf_prof_profiles" on our setup: trex08

- TRex server crashed with  segfault, output:

```
*** traceback follows ***

1             0x5546d2 ./_t-rex-64() [0x5546d2]
2       0x7f819e3385d0 /lib64/libpthread.so.0(+0xf5d0) [0x7f819e3385d0]
3             0x523391 CSTTCp::UpdateTGNames(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&) + 33
4             0x76374f TrexRpcCmdAstfGetTGNames::_run(Json::Value const&, Json::Value&) + 703
5             0x6a5b8d TrexRpcCommand::run(Json::Value const&, Json::Value&) + 77
6             0x6a048c JsonRpcMethod::_execute(Json::Value&) + 44
7             0x69daed TrexJsonRpcV2ParsedObject::execute(Json::Value&) + 125
8             0x69b9a6 TrexRpcServerReqRes::process_request_raw(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) + 486
9             0x69c228 TrexRpcServerReqRes::process_zipped_request(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) + 456
10            0x69c515 TrexRpcServerReqRes::handle_request(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 453
11            0x69cbf7 TrexRpcServerReqRes::_rpc_thread_cb_int() + 1575
12            0x69d326 TrexRpcServerReqRes::_rpc_thread_cb() + 6
13      0x7f819dc4727f so/x86_64/libstdc++.so.6(+0xba27f) [0x7f819dc4727f]
14      0x7f819e330dd5 /lib64/libpthread.so.0(+0x7dd5) [0x7f819e330dd5]
15      0x7f819d3a602d clone + 109


*** addr2line information follows ***

basic_string.h:6289
??:0
vector.tcc:203
trex_astf_rpc_cmds.cpp:584
trex_rpc_cmd.cpp:84
trex_rpc_jsonrpc_v2_parser.cpp:98
trex_rpc_jsonrpc_v2_parser.cpp:66 (discriminator 4)
gthr-default.h:777
trex_rpc_req_resp_server.cpp:303
trex_rpc_req_resp_server.cpp:223
basic_string.h:211
trex_rpc_req_resp_server.cpp:84
??:0
??:0
??:0

```
- Bug at src/stx/astf/trex_astf_rpc_cmds.cpp line 582:
`            lpstt->UpdateTGNames(CAstfDB::instance(pid->get_dp_profile_id())->get_tg_names());
`Using "get_instance()" instead of "instance()" pass the test but it's not the behavior we want, it will massed up the counters each time creating a new CAstfDB.

PR #461, opened 2020-04-21
DPDK 20.02 update corrected Napatech driver 

bkuhre:
The Napatech driver has been corrected to compile with DPDK 20.02.
The problem was that some ETHER defines had changed names from 19.11 to 20.02.


hhaim:
Wow, that was fast. Thanks

hhaim:
@bkuhre 
rebased. There was a warning that was fixed. Please verify

bkuhre:
That is fine. I did not see that the function was changed from void to returning an int. Thank you for correcting this.

Issue #467, opened 2020-04-26
Reopening "Added mutex lock for CAstfDB instance access" #457


elados93:
see @jsmoon 

Reopening issue [457](https://github.com/cisco-system-traffic-generator/trex-core/pull/457) following a bug

I have reverted 6b486ff

Failure on test: "test_astf_prof_profiles" on our setup: trex08

TRex server crashed with segfault, output:

```
*** traceback follows ***

1             0x5546d2 ./_t-rex-64() [0x5546d2]
2       0x7f819e3385d0 /lib64/libpthread.so.0(+0xf5d0) [0x7f819e3385d0]
3             0x523391 CSTTCp::UpdateTGNames(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&) + 33
4             0x76374f TrexRpcCmdAstfGetTGNames::_run(Json::Value const&, Json::Value&) + 703
5             0x6a5b8d TrexRpcCommand::run(Json::Value const&, Json::Value&) + 77
6             0x6a048c JsonRpcMethod::_execute(Json::Value&) + 44
7             0x69daed TrexJsonRpcV2ParsedObject::execute(Json::Value&) + 125
8             0x69b9a6 TrexRpcServerReqRes::process_request_raw(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) + 486
9             0x69c228 TrexRpcServerReqRes::process_zipped_request(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) + 456
10            0x69c515 TrexRpcServerReqRes::handle_request(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 453
11            0x69cbf7 TrexRpcServerReqRes::_rpc_thread_cb_int() + 1575
12            0x69d326 TrexRpcServerReqRes::_rpc_thread_cb() + 6
13      0x7f819dc4727f so/x86_64/libstdc++.so.6(+0xba27f) [0x7f819dc4727f]
14      0x7f819e330dd5 /lib64/libpthread.so.0(+0x7dd5) [0x7f819e330dd5]
[...]

PR #468, opened 2020-04-27
fix get_tg_names RPC command crash issue

jsmoon:
Hi, Issue #467 will be fixed by this PR.


jsmoon:
@hhaim I think the failed test cases are not related to my changes. They are STL performance test cases.

hhaim:
reabsed.

PR #443, opened 2020-03-11
ASTFProgram can send by size and fill.

jsmoon:
Hi @hhaim, This PR is an implementation of suggested issue #401.

```
prog.send(buf, size=1000, fill='*')
```

Please check my changes and give your opinion.

PR #477, opened 2020-05-25
fix send with fill issue in client

jsmoon:
Hi, this PR is to fix my fault at #443.
Due to missing b64encode, the server cannot get the user's fill data properly.
Now the fill data will be applied to the server.


jsmoon:
@ybrustin I think this test was also failed by the DMZ issue. Could you recheck this one?

ybrustin:
recheck

hhaim:
rebased

Issue #476, opened 2020-05-21
documentation  on multi profile on the same port


hhaim:
@jsmoon could you point me to the documentation of this feature?

jsmoon:
@hhaim I didn't write any documentation under doc/ contents. I just added a simple comment for l7_map usage at ASTFAssociationRule.

hhaim:
@jsmoon thanks. Found the commit and the API doc. I see the example that you added too. 
I think it would better add an explicit example in our manual as this is not a simple mechanism

egwakim:
How about adding document as below?
In the trex_astf.html,

2.3.26 Tutorial : Server port sharing between multiple profiles
Description of port sharing feature.
Example1. 2 profiles using different server ports without port sharing(Use two server ports)
Example2. Same as Example1, but, server port shared with l7_map API(Use single server port)
Explanation of Example2.

PR #482, opened 2020-05-29
Added document for server port sharing

jsmoon:
Hi @hhaim, this PR is prepared for issue #476.

### DIFF
--- a/doc/trex_astf.asciidoc
+++ b/doc/trex_astf.asciidoc
@@ -2015,6 +2015,81 @@ In the end we create a profile with two templates (could be much more).
 <1> pps is a tunable like the ones shown in the previous tutorials. You can add it while using the CLI with -t pps=20. 
 In case it wasn't specified pps=1.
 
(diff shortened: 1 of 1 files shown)


hhaim:
@jsmoon good. It was rebased but I have a follow up questions. 

What would happen in case of (no mask, no value)
```l7_map={"offset":[0,1,2,3],```

1. What is the default mask: is it? [255,255,255,255] 
2. What is the default value?  is it taken from the tempaltes value? Is it "GET " and "POST" in your example
3. What do you mean by server mode? It is always server mode.. As only the server need to differentiate the templates 
>>> "Server Mode
  If you want to use it for the server mode, "value" should be used. The sizes of "offset" and "value" 
  should be the same."

jsmoon:
1. yes.
2. If there is no `"value"` specified by the user, it will be taken from each template's client program send and send_msg `buf`.
3. I meant that is server only mode when `--astf-server-only` is used.

hhaim:
understood.

PR #483, opened 2020-06-01
Enabled tx packet capture for the flow_stats_latency

jsmoon:
Hi, this PR is to capture the updated TX packet by flow_stats_latency.
Currently, the captured packet in TX doesn't contain the final contents when the user specifies the STLFlowLatencyStats. So, the packets are captured in TX and RX can be different in this case.
My user has complained about this situation. During analysis by the packet capture, he was confused that the RX packet contents are different from TX ones.


hhaim:
rebased

Issue #478, opened 2020-05-25
Invalid UDP checksum in flowlatency stat enabled packets


egwakim:
With TRex STL mode, 
If flowlatency stats enabled in UDP packets, UDP checksum value is invalid.

The reason is
If user enable flowlatency stat, 16 bytes of last payload will be replaced into flowlatency header.
It'll cause wrong UDP checksum value in Rx side.
Adding dummy flow_var with fix_chksum_hw() would be possible, but, it's not suitable workaround because user should add dummy flow_var for it.

If user use checksum when composing UDP packet(udp.checksum != 0), checksum need to be updated if user enable flowlatency stats.

Another issue is
In Tx packet capture side, these replaced packet payload(by flowlatency) is not visible, it's visible only Rx side.
Tx and Rx should be same, but, these are different now.

egwakim:
This issue can be 100% reproducible when UDP checksum used with flowlatency stat enabled.

jsmoon:
I think `fix_checksum_ipv4` and `fix_checksum_hw` can be allowed without flow variables in SteamVm. And the flow variables should be needed only for the variable reference instructions.
@hhaim if you agree with my opinion, I'll prepare a PR for this issue.

hhaim:
@jsmoon agreed. I would remove this restriction

PR #484, opened 2020-06-01
fix_checksum_ipv4 and fix_checksum_hw allowed without flow variables …

jsmoon:
Hi, this PR is prepared for issue #478.
This change makes fix_checksum_ipv4 and fix_checksum_hw available even though there is no flow variable in StreamVm.


hhaim:
rebased

PR #489, opened 2020-06-08
STLFlowLatencyStats allowed for header changed IP packet.

### LINKED ISSUE #485: [Issue] FlowLatencyStat does not work when IP header changed
In our testing network,
Original IP header changed in Rx side including IP.TOS field.
In that case, Rx packet counters were all zero in FlowLatencyStat.
From Rx captured packets, 16B FlowLatency header(at the end payload) was not changed.

I assume that it's caused by how FlowLatencyStat works.

Will it possible to fix this issue?

jsmoon:
This PR resolves #485.

### DIFF
--- a/src/flow_stat_parser.cpp
+++ b/src/flow_stat_parser.cpp
@@ -340,15 +340,23 @@ uint8_t CFlowStatParser::get_protocol(){
     return (proto);
 }
 
-bool CFlowStatParser::is_fs_latency(){
-    uint8_t tos = 0;
-    if (m_ipv4) {
-        tos = m_ipv4->getTOS();
+bool CFlowStatParser::is_fs_latency(rte_mbuf_t *m){
+    // in case of flow stats
+    uint32_t ip_id = 0;
+    get_ip_id(ip_id);
+
+    if (ip_id > IP_ID_RESERVE_BASE) {
+        return true;
     }
-    if (m_ipv6) {
-        tos = m_ipv6->getTrafficClass();
-    }
-    return ((tos&0x01)==0x01)?true:false;
+
+    // in case of flow stats latency
+    uint8_t tmp_buf[sizeof(struct flow_stat_payload_header)];
+    struct flow_stat_payload_header *fsp_head =
+        (flow_stat_payload_header *)utl_rte_pktmbuf_get_last_bytes(
+            m, sizeof(struct flow_stat_payload_header), tmp_buf);
+
+    return ((fsp_head->magic == FLOW_STAT_PAYLOAD_MAGIC) &&
+            (fsp_head->hw_id < MAX_FLOW_STATS_PAYLOAD));
 }
 
 uint8_t CFlowStatParser::get_ttl(){
--- a/src/flow_stat_parser.h
+++ b/src/flow_stat_parser.h
@@ -85,7 +85,7 @@ class CFlowStatParser {
     virtual int get_payload_len(uint8_t *p, uint16_t len, uint16_t &payload_len);
     virtual uint16_t get_pkt_size();
     virtual uint8_t get_ttl();
-    bool is_fs_latency();
+    bool is_fs_latency(rte_mbuf_t *m);
     uint8_t get_protocol();
 
     uint8_t *get_l3() {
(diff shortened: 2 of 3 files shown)


hhaim:
@jsmoon 
There are a few issues in this short PR

1.in case ip.id was changed by DUT and it is zero. the MAGIC won't help and it won't be considered 

dur to this
```       if (m_rcv_all || is_flow_stat_payload_id(ip_id)) {```

```
    if (m_rcv_all || (ret2 == 0)) {
      if (m_rcv_all || is_flow_stat_id(ip_id)) {
        uint16_t hw_id = get_hw_id((uint16_t)ip_id);
        if (m_rcv_all || is_flow_stat_payload_id(ip_id)) {
          struct flow_stat_payload_header *fsp_head =
              (flow_stat_payload_header *)utl_rte_pktmbuf_get_last_bytes(
                  m, sizeof(struct flow_stat_payload_header), tmp_buf);
```


So there is no meaning to test TOS nor MAGIC in the filter if the handle function can't be more resilient in this case 

2. The MAGIC is not tested correctly in your case ! in case of a packet with more than one mbuf you need to check the **last mbuf**  see how this code works 
```(flow_stat_payload_header *)utl_rte_pktmbuf_get_last_bytes(```

see my last comments on the thread about the issue. Either fix the handle function and remove the check for ip.id or just check for ip.id in the filter

**Comment by jsmoon** [comment]
@hhaim 
1. There is no relation between TOS+ip_id and MAGIC. In SW mode, `m_rcv_all` is true. It means the `handle_pkt` check the MAGIC always in this case.
2. This mbuf is allocated for RX purposes. And as far as I know, it is a **single** mbuf. So, I thought that it should be safe and would be lighter than using `utl_rte_pktmbuf_get_last_bytes`.

**Comment by hhaim** [comment]
@jsmoon 
1. so the filter should be just TOS || IP.id || MAGIC - there is no need to check if you are in software mode, this function is called only in software mode
2. You are wrong! You should use ```utl_rte_pktmbuf_get_last_bytes```

**Comment by jsmoon** [comment]
@hhaim 
1. OK. I'll remove checking in SW mode.
2. OK. I'll change it also.
Since I'm wrong about the mbuf, I think we need to fix the following logical error also.
```
    uint16_t pkt_size= rte_pktmbuf_data_len(m);
    CFlowStatParser_err_t res=m_parser->parse(p,pkt_size);
```
rte_pktmbuf_data_len() wouldn't return total packet size.

**Comment by hhaim** [comment]
@jsmoon 
There is no issue with ```uint16_t pkt_size= rte_pktmbuf_data_len(m);``` it is by design. The parser is using only the first 64B header and we wanted by sure we are not overflow the **first**  mbuf size which is 2048. The parser is not scatter/gather aware.
To fix this there is a need to move the mbuf to the parser.

**Comment by jsmoon** [comment]
@hhaim I understood. Thanks for your explanation.

**Comment by jsmoon** [comment]
@hhaim should I do something on this change?

**Comment by hhaim** [comment]
Looks good. 
I would convert this 

```

if (ip_id && (tos & TOS_GO_TO_CPU)) {
        return true;    }
```

to

```
if (  ip_id ) {
        return true;   
}
```

jsmoon:
@hhaim
According to your suggestion, please confirm the following issues.
1. Since my user fills random ipv4.id for his packet (I don't know exactly what is the purpose), I thought TOS_GO_TO_CPU is a good filter for normal packets.
And if TOS_GO_TO_CPU is not used here, according to my analysis, there is no usage left for TOS_GO_TO_CPU.
**Are you sure to remove checking TOS_GO_TO_CPU?**
2. I thought about filtering ip_id like as the handle_pkt doing. Simply, I can change it `ip_id > ip_id_base`. But since the `ip_id_base` value is variable, I should get the exact value by "get_platform_api().get_port_stat_info(0, num_counters, cap, **ip_id_base**)". But I think `ip_id > IP_ID_RESERVE_BASE` would cover most cases and lighter than previous value.
**Which one is better? `ip_id > IP_ID_RESERVE_BASE` or `ip_id > ip_id_base`?**

hhaim:
@jsmoon 
1. The ipv4.Id/ipv6.flow is offered by TRex server and it is in reserve range
2. IP_ID_RESERVE_BASE is enough

jsmoon:
Thanks for your confirmation.

hhaim:
looks good.

hhaim:
rebased

PR #494, opened 2020-06-18
fix possible ASTF crash issue derived from #455

jsmoon:
Hi @hhaim, this PR is to fix my faulty code derived from #455.

ASTF DP allows the multiple triggering to stop a profile. The stop_profile_ctx() can call set_profile_stopping() multiple times.
If there is a long lifetime flow in the profile and the duration without `nc` option is short, the profile can live after the duration for a while. Meanwhile, a user requests to stop the profile. It may cause a crash if the CGenNodeTXFIF is allocated for another node like CGenNodeCommand.

I'm sorry for my carelessness in the coding.


jsmoon:
@hhaim, have you checked this PR?

hhaim:
@jsmoon will do today

PR #495, opened 2020-06-18
Use configured or discovered dest mac addresses

dverbeir:
When the generator is used against a DUT that consists of a router (gateway) and a device that performs Direct Server Return, or any setup where traffic on a port is asymmetrical in terms of
layer2 counterpart, the source mac address of incoming packets should not be used as destination for outgoing packets of a flow. Instead the configured 'dest_mac' or discovered (ARP) mac address
of the configured 'default_gw' should be used.

Note1: I don't fully understand all trex modes and I wasn't sure that always using the configured/discovered mac would be appropriate. So I introduced this CGlobalInfo::m_options.is_mac_set() check (mainly to avoid a memcmp(empty_mac)) but, except in the case of client config, the configured/discovered mac will always be used for flow templates. If it's also appropriate for the case of client config, then the flag and check could be removed.

Note2: With this change, the mac is now copied from CGlobalInfo::m_options instead of from the incoming packet that triggers to creation of the new server-side flow. This may have a performance impact as this global config might not be as "hot" as the packet data. Though it will be used at each new flow, but this global data did not seem to be laid out with the same performance focus as some others... I did not have the opportunity to make any performance comparison as my use case only works with the change...

### DIFF
--- a/doc/trex_astf.asciidoc
+++ b/doc/trex_astf.asciidoc
@@ -2809,6 +2809,7 @@ link:cp_astf_docs/index.html[python index]
 |  tunable      | min-max|default| per-template |global| Description  
 |   ip.tos    | uint8 |0 |+|+| ipv4/ipv6 TOS/Class. limitation LSB can't be set as it is used by hardware filter on some NICs 
 |   ip.ttl    | uint8 |0 |+|+| ipv4/ipv6 TTL/TimeToLive. limitation can't be higher than 0x7f as it might used by some hardware filter on some NICs 
+|   ip.dont_use_inbound_mac | bool |0|+|+| server side will use configured src/dest MACs for the port or resolved gateway MAC instead of reflecting incoming MACs
 |   ipv6.src_msb    | string |0 |+|+| default IPv6.src MSB address. see Priority in ipv6 enable
 |   ipv6.dst_msb    | string |0 |+|+| default IPv6.dst MSB address. see Priority in ipv6 enable
 |   ipv6.enable     | bool   |0 |+|+| enable IPv6 for all templates. Priority is given for global (template come next) -- not as other tunable. It means that for mix of ipv4/ipv6 only per template should be used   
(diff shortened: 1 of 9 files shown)


hhaim:
@dverbeir interesting use-case. TRex supports multi sub-interfaces and with Linux namespace it is possible to have different route to the different servers IP (depends on the DUT route configuration), In this case there could be the multiple default gateway to your solution there is an assumption that there is one. 
Anyhow, I think your solution is good in the case of one default gateway per-port. I would move this configuration to RPC API (keep it global) or per profile (better). 
Let me know if you know how to do that?

hhaim:
@dverbeir it would be easier to make it a tunable. You can look, for example, commit.

dverbeir:
> @dverbeir it would be easier to make it a tunable. You can look, for example, commit.
@hhaim You probably forgot to paste the example you intended.

Anyway, regarding moving to tunables, did you mean to move just the configuration (whether the MACs from incoming pkt should be used or those from configuration and/or gw resolution), or the actual src+dst macs to use? I thought that putting the macs in tunables (and using incoming pkt macs if tunable not set) was a good idea for performance reasons, but then it looks like I would need to have an array of dst+src macs, one entry per port. Maybe that's too much.
Which of those approaches (or any other) did you mean?

hhaim:
@dverbeir I mean just replace the CLI configuration with tunable configuration which is more interactive friendly - there is no need to add the dst_mac per port for now as it is another feature.

dverbeir:
> @dverbeir I mean just replace the CLI configuration with tunable configuration which is more interactive friendly - there is no need to add the dst_mac per port for now as it is another feature.



> @dverbeir I mean just replace the CLI configuration with tunable configuration which is more interactive friendly - there is no need to add the dst_mac per port for now as it is another feature.

@hhaim Did you see the PR update (I just included doc update) ? Is that the approach you were thinking of regarding tunables?

hhaim:
@dverbeir looks good. thanks

PR #497, opened 2020-06-25
Fixed send by size causes crash in multicore

jsmoon:
Hi, this PR is to fix the crash in multicore usage.

Although the program buffers are created and removed in one core, it can be referenced in multiple cores at `tcp_build_dpkt_`.
```
            if (unlikely(!rb.need_indirect_mbuf(dlen))){
                /* last one */
                rte_pktmbuf_refcnt_update(mn,1);
                lastm->next =mn;
                bsize = rb.get_mbuf_size();
            }else{
```
Since `rte_mbuf_set_as_core_local(m)` should be used for single-core access, it should be removed from the program buffers.

@hhaim, please check my change and give your opinion.

### DIFF
--- a/src/44bsd/tcp_socket.cpp
+++ b/src/44bsd/tcp_socket.cpp
@@ -934,7 +934,10 @@ int utl_mbuf_buffer_create_and_copy(uint8_t socket,
         uint32_t alloc_size = bsd_umin(blk_size,size);
         auto copy_size = bsd_umin(d_size, alloc_size);
         rte_mbuf_t* m;
-        if (copy_size || !mbuf_fill) {
+        if (mbuf_fill && (alloc_size == blk_size)) {
+            m = mbuf_fill;
+            rte_mbuf_refcnt_update(m, 1);
+        } else {
             m = tcp_pktmbuf_alloc(socket,alloc_size);
             assert(m);
             char *p = (char *)rte_pktmbuf_append(m, alloc_size);
@@ -943,8 +946,7 @@ int utl_mbuf_buffer_create_and_copy(uint8_t socket,
                 d += copy_size;
                 p += copy_size;
                 d_size -= copy_size;
-            } else {
-                rte_mbuf_set_as_core_local(m);
+            } else if (!mbuf_fill) {
                 mbuf_fill = m; // filled data reference mbuf
             }
             for (auto i = 0; i < (alloc_size-copy_size); i++ ) {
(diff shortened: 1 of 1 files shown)


hhaim:
@jsmoon if the buffers are local per core, there is no issue with this code.
I need to look again into the buffer allocation

hhaim:
@jsmoon now I see the issue. the buffers are **shared** and in the past, they were CONST but were changed because of this issue  https://trex-tgn.cisco.com/youtrack/issue/trex-537
However, you added another similar function (with fill capability) base on my original code for buffer **but** with a **few* bugs that should crash immediately .. 

```int utl_mbuf_buffer_create_and_copy(uint8_t socket,
                                    CMbufBuffer * buf,
                                    uint32_t blk_size,
                                    uint8_t *d,
                                    uint32_t d_size,
                                    uint32_t size,
                                    uint8_t *f,
                                    uint32_t f_size,
                                    bool mbuf_const){

```

Looking into this function again,

1. You are right, the buffer should **not** be local -- but this is not the only issue !
2. The buffers can't have a rreference (64->2048), beacuse it is going to be used as a referance in tcp code and it is not alowed to have ref->ref-> attach

suggest to fill it with buffers with constant size of 2048 and revert this 

```
            m = tcp_pktmbuf_alloc_small(socket);
            assert(m);
            rte_mbuf_set_as_core_local(m);
            rte_pktmbuf_attach(m, mbuf_fill);
            rte_pktmbuf_trim(m, blk_size-alloc_size);

```

jsmoon:
@hhaim, Thank you for your review.
Do you know why ref->ref->attach is not allowed? What can be happened in that case?

hhaim:
@jsmoon I don't know what happen in this case. Have a look into DPDK manual for why it is not supported. 
there is indirect and none indirect. there is no such thing as indirect that point to indirect ! 

https://doc.dpdk.org/guides/prog_guide/mbuf_lib.html

>> There are a few things to remember when dealing with indirect buffers. First of all, an indirect buffer is never attached to another indirect buffer. Attempting to attach buffer A to indirect buffer B that is attached to C, makes rte_pktmbuf_attach() automatically attach A to C, effectively cloning B. Secondly, for a buffer to become indirect, its reference counter must be equal to 1, that is, it must not be already referenced by another indirect buffer. Finally, it is not possible to reattach an indirect buffer to the direct buffer (unless it is detached first).

hhaim:
@jsmoon I'm not sure this is what you want. 
using this code ```mbuf_fill``` will be always nil and this is not what you want

jsmoon:
@hhaim, this code seems to be complicated to understand.
`mbuf_fill` will be assigned by allocated buffer when `copy_size` is 0 which means the buffer is filled by fill pattern only. That's what I wanted.

hhaim:
understood.

PR #503, opened 2020-07-02
Add Fix Icmpv6 FE instruction

EgorBlagov:
- Add software based Field Engine instruction to fix ICMPv6 checksum
- Add functional tests
- Add examples

NOTES:

- I didn't get whether we could use `rte_raw_cksum` and corresponding methods, but I've used software FixIpv4 instruction as sample, hence extended `fast_csum`, and implemented ipv6 pseudo header checksum calculation
- stl packet builder scapy cannot find out offset names for some packet types (e.g. `ICMPv6ND_NA`, because they have static field `name` which doesn't match classname, however in `trex_stl_packet_builder_scapy.py` we use this construction  to obtain layer name from offset number
```python
# ...
    for pkt in self.pkt_iter():
            layer_name = pkt.__class__.__name__
            if pkt._offset > offset_int:
                break
# ...
```

I suppose this should be fixed (if possible) in separate PR.
- I'm not sure whether we should use one instruction for ICMP and ICMPv6, but I need ICMPv6 only, that's why I've implemented it alone.

### DIFF
--- a/doc/trex_rpc_server_spec.asciidoc
+++ b/doc/trex_rpc_server_spec.asciidoc
@@ -50,6 +50,10 @@ include::trex_ga.asciidoc[]
 | 1.9    | Egor Blagov (eblagov)
 |
 - add stack name in `get_system_info`
+| 1.10   | Egor Blagov (eblagov)
+|
+- update VM instructions 
+
 
 |=================
 
@@ -1347,6 +1351,19 @@ Fix TCP/UDP and IPv4 headers using hardware assit engine
 | Field       | Type        | Description
 | type        | string      | ''fix_checksum_ipv4''
 | pkt_offset  | uint16      | offset of the field to fix 
+|=================
+
+===== fix_checksum_icmpv6
+
+Fix ICMPv6 layer checksum on top of IPv6 using software
+
+.Object type 'vm - fix_checksum_icmpv6'
+[options="header",cols="1,1,3"]
+|=================
+| Field       | Type        | Description
+| type        | string      | ''fix_checksum_icmpv6''
+| l2_len      | uint16      | len of L2 (e.g. 22 Ether + Dot1Q + Dot1Q)
+| l3_len      | uint16      | len of l3 header (e.g. 40 for IPv6)
 |=================
 
 ===== flow_var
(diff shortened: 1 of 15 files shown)


hhaim:
@EgorBlagov looks good. There is no hardware accelerator for icmv6.cs. 
Could you elaborate for the use-case? why to check ICMPv6 at this speed? if there is a need for protocols EMU could be used

EgorBlagov:
@hhaim 
We have some testcases where need to manually send NS or NA packets to DUT, emulating thousands of ipv6 subinterfaces, can it be achieved with EMU?

hhaim:
@EgorBlagov I figure out that this is a hack ! I suggest to use EMU 
EMU can support scale of millions of clients with statistics of IPv6 ND protocols- you would be able to verify if all the clients finished. You won’t be able to do this using FE

EgorBlagov:
@hhaim 
Do you mean this PR should not be merged, and EMU should be used for all?

Another example testcase is user brings up ipv6 subinterfaces (this was implemented internally in trex-core fork, but we can consider migrating to EMU), waits something and then manually sends packets like NS/NA or DAD. ICMPv6 checksum instruction will cover these cases.

Or you suggest EMU should be extended with custom commands like send DAD from clients, send NA from clients or else?

hhaim:
@EgorBlagov It's always good to have this feature. 

The EMU main point is the **feedback**. It has the protocol implementation (ipv6 ND RFCs)  something that you **can't** do using FE that does **not** have feedback !

For sending DAD, EMU does that, you can add RPC to trigger another DAD, or just remove/add the virtual interface and read the counters if DUT behave correctly (again feedback)

BTW this feature could be used in parallel to EMU, but again, I don't see a reason except for DDOS attack of ICMPv6 but this is not your main use-case right?

EgorBlagov:
@hhaim 
Ok I got it, I certainly will try to ask folks to consider migrating to EMU, since it looks very cool. DDOS attack is not my use-case at all :)

trex bot says tests failed due to performance issue, should I do something about it and where to start? I'm just confirming whether these tests are stable and are sure affected by the feature.

hhaim:
@EgorBlagov the performance degradation is expected. Because you added this new instruction. I will check it internally on all the setup to verify the effect on different platforms

hhaim:
rebased

PR #504, opened 2020-07-08
Increment STL API version

EgorBlagov:
When I submitted PR for icmpv6 fix, I took FE value_list PR as reference, and forgot about STL API version increment, but I think it's requred. PLS correct if I'm wrong.

- FE instruction was added, and newer client will show error with older trex about
  unknown instruction type


hhaim:
@EgorBlagov yes, you are right!

hhaim:
rebased

PR #505, opened 2020-07-08
Remove ASTF DP flow node consumption for the profile duration.

jsmoon:
Hi, this PR removes the DP flow node consumption for profile duration in ASTF mode.

Since the number of DP flow nodes is a constant 8192, maximum number of running profiles is limited by this number. If all profiles are running with duration, they will be half of without duration.
So, I'd like to remove this restriction. My change reuses TCP_TX_FIF node to trigger stopping profile after duration.

### DIFF
--- a/src/bp_sim.h
+++ b/src/bp_sim.h
@@ -74,6 +74,9 @@ limitations under the License.
 #include "stl/trex_stl_fs.h"
 #include "hot_section.h"
 
+/* astf includes */
+#include "astf/trex_astf_dp_core.h"
+
 class CGenNodePCAP;
 
 #define FORCE_NO_INLINE __attribute__ ((noinline))
@@ -661,8 +664,9 @@ class CPerProfileCtx;
 struct CGenNodeTXFIF : public CGenNodeBase {
 public:
     CPerProfileCtx *    m_pctx;
+    double              m_time_stop;
 
-    uint8_t             m_pad_end[104];
+    uint8_t             m_pad_end[96];
 
     /* CACHE_LINE */
 #ifdef __PPC64__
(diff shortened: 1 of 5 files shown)


hhaim:
@jsmoon you is it just to save 4K *128 bytes == 0.5Mbytes?  I don't think it is worth it, just change the default from 8K nodes to 16k and make sure the YAML DP_flow can support higher numbers

jsmoon:
@hhaim there is an another benefit.
Since there is no way to free the duration node before it scheduled, if there is long duration profile issued once, although client requests to stop the profile, the duration node resides in the node scheduler. I think these operation may cause unpredictable and unexpected result during start new profile operation.
Although the server-only mode still has the issue, I thought it is worth for the client-server mode.

hhaim:
@jsmoon but again we are talking about 4K nodes of 128 bytes. Why not just to enlarge it by default and make sure YAML supports works

jsmoon:
@hhaim with the current 8192, client can request 8K profiles without the duration but only 4K profiles with the duration. I think it is better the number of maximum profiles is the same than the different. It's more predictable in the client side.

hhaim:
@jsmoon ok. Are you really getting to this limit and adding 8K profiles in real senarios?  Could you elaborate on the use-case just for understanding. Is profile map to a user/client?

hhaim:
rebased

jsmoon:
My user wants to simulate UE App behavior dynamically. My team is providing a profile for a typical UE App behavior. They request us to add and remove profiles dynamically for their UE App behavior simulation. 1 UE can have several App behavior.

PR #519, opened 2020-08-26
Initial support for GTPU support

srakula-trex:
a.  You can enable the gtp mode using RPC
    b.  Tx/Rx callbacks will initialized based on the ports you choose to enable gtp
    c.  Accordingly encap/decap of GTPU header takes place.
    d.  Currently we dont have validation of incoming GTPU packets.


hhaim:
@srakula-trex 
1. please revert re_mbuf.h/tre_mbuf_core.h those file are DPDK.  Just use the API in the right way. there is no need to change the code

2. code style - please follow https://google.github.io/styleguide/cppguide.html 
2.1 tab -> 4 spaces 
2.2 

```
 if (eth->ether_type == kEtherTypeBeVLAN) 
  {
	
   }  
 ```
 do 

``` 
  if (eth->ether_type == kEtherTypeBeVLAN) {
	
  } 
```


the code is mixed -- this is another example that is OK

```   
if (!CGlobalInfo::m_options.m_ip_cfg[port].is_gtp_enabled()) {
		return nb_pkts;
	}
   
```

some code is not aligned, for example 

```
   for (i=0; i < (nb_pkts -3 ); i++) {
        rte_prefetch0(rte_pktmbuf_mtod(pkts[i + 3], void *));
		rte_mbuf * m = pkts[i];
        iph = rte_pktmbuf_mtod_offset(m, struct rte_ipv4_hdr *, m->l2_len);
	tunnel = (Tunnel *)m->dynfield1;
        if(tunnel)
        {
	  tunnel->Prepend(m,is_ipv4, queue);
        }

	}

```

3. Why there is a YAML file configuration file, don't you have RPC for ASTF?

**Comment by hhaim** [comment]
1) How do you know that you have at least 3 packets? 

```
 uint16_t Tunnel::RxCallback(uint16_t port, uint16_t queue,
		struct rte_mbuf *pkts[], uint16_t nb_pkts, uint16_t max_pkts,
		void *user_param) {

 
     for (i = 0; i < 3 && i < nb_pkts; i++) {
        rte_prefetch0(rte_pktmbuf_mtod(
                pkts[i], void *));
    }

    /*Do batch processing for next set of 3 packets each time */
    for (i = 0; i < (nb_pkts -3 ); i++) {
        rte_prefetch0(rte_pktmbuf_mtod(pkts[i + 3], void *));
     	m = pkts[i];
        t.Adjust(m,queue);
	}

    /*Process the remaining packets*/
    for (; i < nb_pkts ; i++) {
        m = pkts[i];
        t.Adjust(m,queue);
    }
```

2) in case of failure you need to exit, so you need to return an error code
and exit in main code..
```
void Tunnel::InstallRxCallback(uint16_t port, uint16_t queue) {


    if (!CGlobalInfo::m_options.m_ip_cfg[port].is_gtp_enabled()) {
        printf("Tx Tunnel disabled on %u:%u...But doing Rx callback", port, 0);
    }

	if (NULL == rte_eth_add_rx_callback(port, queue, Tunnel::RxCallback, NULL)) {
		printf("failed to install RX callback at %u:%u.", port, queue);
		return;
	}
	printf("installed RX callback at %u:%u.", port, queue);
}
```
3) same for install Tx callback. maybe it is better to have one API to do both and exit in case of an error (with error code)
```
void Tunnel::InstallTxCallback(uint16_t port, uint16_t queue) {
```

4) memory leak, did you test it using Valgrind ? nobody free pre_cooked_hdr  
GTPU::GTPU(uint32_t teid,uint32_t src_ip, uint32_t dst_ip)
{

    Encapsulation *pre_cooked_hdr = new Encapsulation[1];

```
	~GTPU() {  // no free, you need to save Encapsulation *pre_cooked_hdr and free it  
	}
```

5) location of files : move the files to a new folder  src/tunnels
 tunnel.cpp  << split the code Tunnel::xxxe.g.  ```void Tunnel::InstallRxCallback ```to a new file 
  
 6) Didn't get why do you need ```mapOfTeid_v4``` and why do you need YAML information
 don't you have RPC to manage the map (which is per DP) and NOT global? 
      mapOfTeid_v4.insert(pair);

it seems as an old code that not relevant

**Comment by srakula-trex** [comment]
1.  Idea is if i have 3 packets minimum , i would prefetch before processing them . Just to make sure we have cache updated for those packets . If there are less packets, then we don't prefetch.

2.  pre_cooked_hdr - > good catch . will address the change and test it before sending PR.

3 . Took care of updating / moving to new files etc. 

4. Removed un-used old code

**Comment by hhaim** [comment]
1. This is very strange code, I would just remove it. 
 *   this could be negative (nb_pkts -3 )
 *  you are assuming at least 6 packet which is more than you have provided pkts[i + 3] >> this could be a crash 
let's say ```nb_pkts==4``` you will do rte_pktmbuf_mtod(pkts[4]) that does not exits 
```
/*Do batch processing for next set of 3 packets each time */
    for (i = 0; i < (nb_pkts -3 ); i++) {
        rte_prefetch0(rte_pktmbuf_mtod(pkts[i + 3], void *));
     	m = pkts[i];
        t.Adjust(m,queue);
	}
```
2. Did you push a new code?

hhaim:
@srakula-trex I assume we can close this. The diff moved to #521

PR #524, opened 2020-09-09
Use ==/!= operators for literal comparisons

shunghsiyu:
Comparison between a literal and a variable with 'is' and 'is not' is prone to error since 'is' [checks for identity rather than equality](https://adamj.eu/tech/2020/01/21/why-does-python-3-8-syntaxwarning-for-is-literal/), for example

```python3
>>> ''.join([c for c in 'abcd']) is 'abcd'
False
>>> ''.join([c for c in 'abcd']) == 'abcd'
True
```

This sometimes leads to issues that are hard to debug, so change those comparison to ==/!= instead.


hhaim:
@shunghsiyu thanks

PR #540, opened 2020-10-09
 RPC calls for Tunnel Add/Del : Toggle State Client

vineetbvp:
1. Add Tunnel info in clients
    2. Delete Tunnel Info from clients
    3. Activate deactivate clients
    4. Get clients info

### DIFF
--- a/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
@@ -1646,6 +1646,8 @@ class ASTFClient(TRexClient):
 
     # execute 'method' for inserting tunnel info for clients
     def insert_tunnel_in_client_record (self, client = None):
+        '''Add tunnel information for a client'''
+
 
         json_attr = []
 
@@ -1654,12 +1656,12 @@ class ASTFClient(TRexClient):
 
         params = {"attr": json_attr }
 
-        print(self._transmit("add_tunnel_in_client", params))
-
-        #return self._transmit("add_tunnel_in_client", params)
+        return self._transmit("add_tunnel_in_client", params)
 
     # execute 'method' for deleting tunnel info from clients
     def delete_tunnel_from_client_record (self, client_list = None):
+        '''Remove tunnel information from a client'''
+
 
         json_attr = []
 
@@ -1672,6 +1674,7 @@ class ASTFClient(TRexClient):
 
     # execute 'method' for Making  a client active/inactive
     def enable_disable_client(self, thread_id, is_enable = False, client_list = None):
+        '''Knob to toggle state of a client'''
 
         json_attr = []
 
@@ -1686,6 +1689,7 @@ class ASTFClient(TRexClient):
    
      # execute 'method' for getting clients stats [ No of clients Active, Inactive , tunnel enabled]
     def get_clients_stats (self):
+        '''Get Total Clients Status(Active/Inactive) per threads'''
 
         json_attr = []
         params = {"type": 'all',
--- a/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
@@ -1679,7 +1679,7 @@ class ASTFClient(TRexClient):
         json_attr = []
 
         for key in client_list:
-	    json_attr.append({'client_ip' : key})
+            json_attr.append({'client_ip' : key})
 
         params = {"thread_id": thread_id,
                   "is_enable": is_enable,
--- a/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
@@ -1638,3 +1638,73 @@ class ASTFClient(TRexClient):
             self.logger.info(format_text("No profiles found with desired filter.\n", "bold", "magenta"))
 
         text_tables.print_table_with_header(table, header = 'Profile states')
+
+
+
+    ################## GTPU Clients functions ##################
+
+
+    # execute 'method' for inserting tunnel info for clients
+    def insert_tunnel_in_client_record (self, client = None):
+
+        json_attr = []
+
+        for key, value in client.items():
+            json_attr.append({'client_ip' : key, 'sip': value.sip, 'dip' : value.dip, 'teid' : value.teid, "version" :value.version, "thread_id": value.thread_id})
+
+        params = {"attr": json_attr }
+
+        print(self._transmit("add_tunnel_in_client", params))
+
+        #return self._transmit("add_tunnel_in_client", params)
+
+    # execute 'method' for deleting tunnel info from clients
+    def delete_tunnel_from_client_record (self, client_list = None):
+
+        json_attr = []
+
+        for key in client_list:
+            json_attr.append({'client_ip' : key[0], 'thread_id': key[1]})
+
+        params = {"attr": json_attr }
+
+        return self._transmit("delete_tunnel_from_client", params)
+
+    # execute 'method' for Making  a client active/inactive
+    def enable_disable_client(self, thread_id, is_enable = False, client_list = None):
+
+        json_attr = []
+
+        for key in client_list:
+	    json_attr.append({'client_ip' : key})
+
+        params = {"thread_id": thread_id,
+                  "is_enable": is_enable,
+                  "attr": json_attr }
+
+        return self._transmit("enable_disable_client", params)
+   
+     # execute 'method' for getting clients stats [ No of clients Active, Inactive , tunnel enabled]
+    def get_clients_stats (self):
+
+        json_attr = []
+        params = {"type": 'all',
+                  "attr": json_attr }
+
+
+        rc = self._transmit("get_clients_stats", params)
+        records = rc.data()
+
+        return records
+
+
+    @console_api('client', 'ASTF', True)
+    def get_client_stats_line (self, line):
+        '''Display Total Clients Status(Active/Inactive) per threads'''
+
+        records = self.get_clients_stats()
+        for rec in records:
+            self.logger.info(rec)
+
+        return True
+
--- a/scripts/automation/trex_control_plane/interactive/trex/examples/astf/tunnel_many.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/examples/astf/tunnel_many.py
@@ -1,0 +1,21 @@
+from trex.astf.api import *
+
+class Prof1():
+    def __init__(self):
+        pass
+
+    def get_profile(self, **kwargs):
+        # ip generator
+        ip_gen_c = ASTFIPGenDist(ip_range=["11.11.0.1", "11.11.0.10"], distribution="seq")
+        ip_gen_s = ASTFIPGenDist(ip_range=["48.0.0.0", "48.0.255.255"], distribution="seq")
+        ip_gen = ASTFIPGen(glob=ASTFIPGenGlobal(ip_offset="1.0.0.0"),
+                           dist_client=ip_gen_c,
+                           dist_server=ip_gen_s)
+        
+        return ASTFProfile(default_ip_gen=ip_gen,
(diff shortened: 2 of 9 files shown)


scripts/automation/trex_control_plane/interactive/trex/examples/astf/tunnel_many.py
```diff
@@ -0,0 +1,21 @@
+from trex.astf.api import *
```
kvbhaskar, line 1: do we need to this?

scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
```diff
@@ -1638,3 +1638,73 @@ def profiles_line(self, line):
...
+    def insert_tunnel_in_client_record (self, client = None):
+
+        json_attr = []
+
+        for key, value in client.items():
+            json_attr.append({'client_ip' : key, 'sip': value.sip, 'dip' : value.dip, 'teid' : value.teid, "version" :value.version, "thread_id": value.thread_id})
+
+        params = {"attr": json_attr }
+
+        print(self._transmit("add_tunnel_in_client", params))
```
kvbhaskar, line 1657: can u remove the print

scripts/automation/trex_control_plane/interactive/trex/examples/astf/tunnel_regression.py
```diff
@@ -0,0 +1,200 @@
+#!/usr/bin/python
```
kvbhaskar, line 1: how do u plan to run this?
and may be remove it?

vineetbvp: This is under examples. Anyone can see how to use it. Also, can test it after making changes.

scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
vineetbvp: I will remove it and have to replace will return statement.

hhaim:
1. Why there is a print here?
```
 ################## GTPU Clients functions ##################


    # execute 'method' for inserting tunnel info for clients
    def insert_tunnel_in_client_record (self, client = None):

        json_attr = []

        for key, value in client.items():
            json_attr.append({'client_ip' : key, 'sip': value.sip, 'dip' : value.dip, 'teid' : value.teid, "version" :value.version, "thread_id": value.thread_id})

        params = {"attr": json_attr }

        print(self._transmit("add_tunnel_in_client", params))

        #return self._transmit("add_tunnel_in_client", params)

```
2. Please add help string to all Python API  ```insert_tunnel_in_client_record``` and ```delete_tunnel_from_client_record```
In the same format the other API has 

3. Why do you need  ```#include <tunnels/gtp_tunnel.h>``` in tuple_gen.h ?

4. I think it is better to have the tunnel information (json) generic and add a field that specify that this is for  GTPU. I would like to support more tunnel with the same API. The data attached to each client will be opaque  and will be known only for the Tx/Rx callback (convert the pointer to data)

vineetbvp:
Response 
Comment 1. It was left by mistake. Removed it
Comment 2. Added Comment
Comment 3. Not required. removed it

--- a/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
@@ -1646,6 +1646,8 @@ class ASTFClient(TRexClient):
 
     # execute 'method' for inserting tunnel info for clients
     def insert_tunnel_in_client_record (self, client = None):
+        '''Add tunnel information for a client'''
+
 
         json_attr = []
 
@@ -1654,12 +1656,12 @@ class ASTFClient(TRexClient):
 
         params = {"attr": json_attr }
 
-        print(self._transmit("add_tunnel_in_client", params))
-
-        #return self._transmit("add_tunnel_in_client", params)
+        return self._transmit("add_tunnel_in_client", params)
 
     # execute 'method' for deleting tunnel info from clients
     def delete_tunnel_from_client_record (self, client_list = None):
+        '''Remove tunnel information from a client'''
+
 
         json_attr = []
 
@@ -1672,6 +1674,7 @@ class ASTFClient(TRexClient):
 
     # execute 'method' for Making  a client active/inactive
     def enable_disable_client(self, thread_id, is_enable = False, client_list = None):
+        '''Knob to toggle state of a client'''
 
         json_attr = []
 
@@ -1686,6 +1689,7 @@ class ASTFClient(TRexClient):
    
      # execute 'method' for getting clients stats [ No of clients Active, Inactive , tunnel enabled]
     def get_clients_stats (self):
+        '''Get Total Clients Status(Active/Inactive) per threads'''
 
         json_attr = []
         params = {"type": 'all',
diff --git a/src/stx/astf/trex_astf_dp_core.cpp b/src/stx/astf/trex_astf_dp_core.cpp
index 1744873..d97e3e2 100644
--- a/src/stx/astf/trex_astf_dp_core.cpp
+++ b/src/stx/astf/trex_astf_dp_core.cpp
@@ -31,7 +31,7 @@ limitations under the License.
 #include "trex_astf_dp_core.h"
 #include "trex_astf_topo.h"
 #include "trex_client_config.h"
-
+#include <tunnels/gtp_tunnel.h>
 
 using namespace std;
 
diff --git a/src/tuple_gen.h b/src/tuple_gen.h
index 306f9a3..6be94dc 100755
--- a/src/tuple_gen.h
+++ b/src/tuple_gen.h
@@ -40,7 +40,6 @@ limitations under the License.
 #include "trex_client_config.h"
 #include "trex_global.h"
 #include <random>
-#include <tunnels/gtp_tunnel.h>
 
Comment 4.
Json Looks generic. Currently tunnel information is stored as opaque.
To make this API generic, I think changes are required at 2 place.
1. At Json at a parameter type of Tunnel.
2. In below function based on tunnel type call respective constructor to create its object .

What are your thoughts ? 

void TrexAstfDpCore::add_tunnel_for_client(CAstfDB* astf_db, std::vector<client_tunnel_data_t> msg_data) {
    CTupleGeneratorSmart* ctg = astf_db->get_smart_gen(m_thread_id);

    for (auto elem : msg_data) {
        CClientPool *entry = ctg->lookup(elem.client_ip);
        if (entry) {
           void *new_elem = NULL;
           if (elem.version == 4){
               new_elem = (void*) new GTPU(PKT_HTONL(elem.teid),
                                           PKT_HTONL(elem.u1.src_ipv4),
                                           PKT_HTONL(elem.u2.dst_ipv4));
           } else {
               new_elem = (void*) new GTPU(PKT_HTONL(elem.teid),
                                           elem.u1.src_ip,
                                           elem.u2.dst_ip);
           }
           entry->set_tunnel_info_for_clients(elem.client_ip, elem.client_ip, true, new_elem);
        }
    }
}

hhaim:
@vineetbvp you should  push the new code as a PR

hhaim:
@vineetbvp  sorry for the delay, the code is not in the right direction! 
you started with something and then change to something else, the code does not make sense at all 
but I want to start from the high level requirements, 

* You want to add tunnel information per client (GTPU in this case) 
* You want to dynamically enable/disable clients
* You want to get back the tunnel information 
* Enable/Disable of client should not be associate with tunnel and for sure not with GTPU

Instead of adding a new RPC for adding/query the info, I suggest to use the topo infra for tunnels (it is used for GW today)
The tunnel info would be generic with type, each type will have the info

see here 
https://trex-tgn.cisco.com/trex/doc/trex_astf.html#_interactive_mode

You just need to expand the topo information in the GW section 


```
 topo.add_tunnel_info(
        port_id = '0',                 
        src_start = '16.0.0.3',
        src_end = '16.0.0.4',
        type = "gtpu",
       data  ={ .. gtpu info }
      
    )

```

This info will be part of 

```
class ClientCfgEntry {

```

and once a new client will be created the info will be attached. 

The client info is attached to a profile and will have all the info. 
the getter is the same 

```topo show ``` should show the tunnel info 

```

trex>topo show
Virtual interfaces

Port |        MAC        | VLAN |      IPv4       | IPv6
-----+-------------------+------+-----------------+------
0.2  | 12:12:12:12:12:12 |  20  |     5.5.5.5     |   -

Gateways for traffic

Port  |   Range start   |    Range end    |       Dest        |     Resolved
------+-----------------+-----------------+-------------------+------------------
 0.2  |    16.0.0.0     |    16.0.0.1     | 45:45:45:45:45:45 | 45:45:45:45:45:45
 0.2  |    16.0.0.2     |    16.0.0.3     |      2.2.2.2      | 00:0c:29:32:d3:4d
  0   |    16.0.0.4     |   16.0.0.255    |      2.2.2.2      | 00:0c:29:32:d3:4d

Tunnel info 

Port  |   Range start   |    Range end    |       type        |     type | 
------+-----------------+-----------------+-------------------+------------------
 0.2  |    16.0.0.0     |    16.0.0.1     |     GTPU            |  {  ...} 
 0.2  |    16.0.0.2     |    16.0.0.3     |      GTPU      | 
  0   |    16.0.0.4     |   16.0.0.255    |     GTPU      |

```


Most of the code in not relevant and ```topo``` mechanism solves the async issue you have reported 

for example

*  ```thread_id``` should not part of the RPC !! only the server knows how it split the clients to threads, we don't want the Python to be involved 

* ```TrexRpcCmdAddTunnelInClient ``` : not needed, the info should be move to topo

* ```TrexRpcCmdGetClientsStats``` : a few issues. 
** should be part of topo
**  RPC should return json and not string. 
** thread_id is not relevant 

* ```TrexRpcCmdDeleteTunnelFromClient:``` - no need topo is handling the teardown 

* ```TrexRpcCmdEnableDisableClient:``` this is the only RPC function needed! 


*
```    void set_tunnel_info_for_clients(uint32_t        min_ip,
                                   uint32_t        max_ip,
                                   bool            add,
                                   void            *gtpu);
```

the API is wrong, you need to set gtpu object to one clients. Again this is not needed as ```topo``` will solve it

* ```void            *gtpu``` is not freed



Next phase you can add a support for update tunnel info dynamically ( not just disable/enable)


* This code is not clear, should use TCDListNode of client (lookup is fast, just add it to a link-list and remove it)


* remove this
```
  std::list<uint32_t> m_active_clients;
    std::list<uint32_t>::iterator m_cur_act_itr;    
```
```
void CClientPool::set_clients_active(uint32_t  min_ip,
                                     uint32_t  max_ip,
                                     bool      activate)
{
    assert(max_ip >= min_ip);
    uint32_t total_ip  = max_ip - min_ip + 1;
    uint32_t i = min_ip - m_ip_info[0]->get_ip();

    uint32_t count = 0;
    
    while(count < total_ip) {
        if (activate) {
            if (m_ip_info[i]->is_active() == false) {
               if (m_ip_info[i]->get_tunnel_info()) {
                  m_ip_info[i]->set_is_active(activate);
                  m_active_clients.push_back(i);
                  if(m_active_clients.size() == 1)
                     m_cur_act_itr = m_active_clients.begin();
               }
            }
        } else {
            m_ip_info[i]->set_is_active(activate);
            if (m_active_clients.size() > 0 ) {
                if (*m_cur_act_itr == i){
                    m_cur_act_itr++;
                }
                m_active_clients.remove(i);
            }
        }
        i+=1;
        count+=1;
    }
}

```


and there is no need to this line 

```
    uint32_t GenerateTuple(CTupleBase & tuple) {

      CIpInfoBase* ip_info;
      uint32_t idx;

      if ( CGlobalInfo::m_options.is_gtpu_enabled() == 0) {  <<<<<<<<<<<<<<<<<<<<<<<<<<<<<< 
          idx = generate_ip();
          CIpInfoBase* ip_info = m_ip_info[idx];
          ip_info->generate_tuple(tuple);
      }
      else {
        if (m_cur_act_itr == m_active_clients.end())
            m_cur_act_itr = m_active_clients.begin();
        idx = *m_cur_act_itr;
        m_cur_act_itr++;
        ip_info = m_ip_info[idx];
        ip_info->generate_tuple(tuple);
        tuple.setTunHandle(ip_info->get_tunnel_info());
      }

```

The code should be the same for all not relevant for GTPU 
If you assume that most of the client is enabled there is no need for link -list 


There are more issues but I've stopped here

vineetbvp:
Needs to refactor the code and push again. Discarding this pull request.

Issue #537, opened 2020-10-07
TRex crash with SIGBUS after vfio-pci driver update


egwakim:
Hi,
After Linux kernel security update(CVE-2020-12888 : https://people.canonical.com/~ubuntu-security/cve/2020/CVE-2020-12888.html), TRex crashed with SIGBUS error during startup.

Since CVS-2020-12888 patch blocks disabled PCI memory access for the security reason, old DPDK driver will crash with SIGBUS.
There is simple patch on DPDK vfio-pci driver here : https://patchwork.dpdk.org/patch/71962/
And this DPDK patch applied from DPDK 20.08 version.

The simple and quick way is just applying DPDK vfio-pci driver, it's just simple change with small number of lines.
The long term way is upgrading DPDK version to 20.08 version.

Can you look into this issue?

Best Regards
Gwangmoon

egwakim:
Hi @hhaim ,
What is opinion about it?
Shall we patch it right now?
Or Do we need to update the DPDK version?

hhaim:
@egwakim thanks for finding this. 
Better just to apply this patch using PR (it is very short and simple). In the next DPDK upgrade, it will be solved without this patch.

PR #542, opened 2020-10-12
TRex crash with SIGBUS after vfio-pci driver update

yonghwan007:
This is patch on DPDK vfio-pci driver.
This issue was discussed in https://github.com/cisco-system-traffic-generator/trex-core/issues/537


hhaim:
reabsed

PR #531, opened 2020-09-29
hugepage usage reduction for unused socket memory

jsmoon:
Hi, this PR implements the suggestion from #523.
@hhaim Please check my change and give your feedback.

PR #543, opened 2020-10-13
use master thread's NUMA node for messaging socket_id.

jsmoon:
@hhaim, from the PR #531, when only socket: 1 (NUMA node 1) is used, T-Rex can't start by assertion fail.
It's because CMessagingManager tries to allocate from dedicated NUMA node 0 that is not configured in this case.
My change uses master thread's NUMA node that should be configured.

P.S. In addition, from the PR #531, T-Rex also failed to start when master thread's NUMA node is not the same as interface's NUMA node. Since this usage is not recommended, I think I don't need to support it. What do you think about it?

### DIFF
--- a/src/msg_manager.cpp
+++ b/src/msg_manager.cpp
@@ -32,6 +32,7 @@ bool CMessagingManager::Create(uint8_t num_dp_threads,std::string a_name){
     assert(m_cp_to_dp==0);
     m_cp_to_dp = new CNodeRing[num_dp_threads] ;
     m_dp_to_cp = new CNodeRing[num_dp_threads];
+    int master_socket_id = rte_lcore_to_socket_id(CGlobalInfo::m_socket.get_master_phy_id());
     int i;
     for (i=0; i<num_dp_threads; i++) {
         CNodeRing * lp;
@@ -39,11 +40,11 @@ bool CMessagingManager::Create(uint8_t num_dp_threads,std::string a_name){
 
         lp=getRingCpToDp(i);
         sprintf(name,"%s_to_%d",(char *)a_name.c_str(),i);
-        assert(lp->Create(std::string(name),1024,0)==true);
+        assert(lp->Create(std::string(name),1024,master_socket_id)==true);
 
         lp=getRingDpToCp(i);
         sprintf(name,"%s_from_%d",(char *)a_name.c_str(),i);
-        assert(lp->Create(std::string(name),1024,0)==true);
+        assert(lp->Create(std::string(name),1024,master_socket_id)==true);
 
     }
     assert(m_dp_to_cp);


hhaim:
@jsmoon ```rte_get_master_lcore()``` make sure it is implemented in simulation. 

>>when master thread's NUMA node is not the same as interface's NUMA 
I think it is better to verify what the ```dpdk_setup_port.py``` is doing in this case, I think you are right and master will be in NUMA 1, so there should not be a problem. Worth verify that in case all the ports in NUMA 1 the master is configure in NUMA 1 too else exit with falure.  (```CPlatformSocketInfoConfig```)

hhaim:
@jsmoon 

1. Better not to use ```ifndef TREX_SIM```
2. Try to query ```CPlatformSocketInfoConfig``` object and verify that master is the same NUMA as the interfaces (in case it is not zero) and use the master socket (from this object) in the rings

This way you will have verification that if someone mess with the configuration file, you will give an error

jsmoon:
@hhaim, This PR is to make NUMA 1 only configuration possible. And currently the mismatched NUMA configuration gives error results also. The errors are varied by the combination of master thread's NUMA, interface's NUMA and socket configuration.
I think your comment is an improvement and more discussion should be done before implementing it. So I'd like to do it later by another PR.

hhaim:
@jsmoon we could create a separate PR for the warning

My suggestion in this  PR is just a cosmetic 

1. add this function to ```CPlatformSocketInfo```
```  CGlobalInfo::m_socket.get_master_socket()``` that will return the socket of master core 

The impelemtation would be this 

```rte_lcore_to_socket_id(  the_master_thread_id) ```

This way all the code that manipulate sockets and threads would ecapsolate into the same place

**Comment by jsmoon** [comment]
@hhaim I think using `rte_lcore_to_socket_id` in trex_platform.cpp is not appropriate for implementing `CGlobalInfo::m_socket.get_master_socket()`. It requires `#include "bp_sim.h"` to use DPDK function.

What about this simple change?
```
@@ -32,7 +32,7 @@ bool CMessagingManager::Create(uint8_t num_dp_threads,std::string a_name){
     assert(m_cp_to_dp==0);
     m_cp_to_dp = new CNodeRing[num_dp_threads] ;
     m_dp_to_cp = new CNodeRing[num_dp_threads];
-    int master_socket_id = rte_lcore_to_socket_id(rte_get_master_lcore());
+    int master_socket_id = rte_lcore_to_socket_id(CGlobalInfo::m_socket.get_master_phy_id());
     int i;
     for (i=0; i<num_dp_threads; i++) {
         CNodeRing * lp;
```

Which one do you want? implementing `get_master_socket()`? or using `get_master_phy_id()`?

hhaim:
@jsmoon I see the issue. The ```CGlobalInfo:: m_socket``` use the information from the configuration file (built by the script that query the kernel with the information)
Let's keep the code you proposed.

hhaim:
rebased

PR #535, opened 2020-10-05
add --new-memory option to build with new DPDK memory subsystem.

jsmoon:
Hi, this PR is an implementation of issue #480.

I added generated dpdk_config.h to override RTE_EAL_NUMA_AWARE_HUGEPAGES configuration in rte_config.h dynamically.
```
cd linux_dpdk/
./b configure --new-memory
./b build
```
@hhaim what do you think about this change?

PR #546, opened 2020-10-15
fix --new-memory libnuma link error

jsmoon:
@hhaim, this PR fixes numa library link error from #535.
When I use sysroot build for different target, I met the linker error messages.
So, I changed the '-lnuma' location to the end.


hhaim:
@jsmoon this is the right way to do it.

hhaim:
rebased

PR #547, opened 2020-10-16
fix crash by no flow variable in StreamVm from #484

jsmoon:
@hhaim, #484 enabled fix_checksum_hw without flow variables for flow latency stats.
Recently, I received crash issue when there is no flow variable.
There are some assert code to check bss_size should not be 0. I'd missed the code at change #484.

```
assert: ../../../../trex-core/src/stx/stl/trex_stl_stream_vm.h:2039 StreamVmDp::StreamVmDp(uint8_t*, uint16_t, uint8_t*, uint32_t, uint16_t, uint16_t, bool, bool) Assertion 'bss_size' failed.
```


hhaim:
rebased

PR #549, opened 2020-10-23
multiple profiles share client IP info for port allocation in a DP core

jsmoon:
@hhaim, this PR is another trial for #376.
Recently I observed `err_dct` counter when I use the same client IP and server IP + server port (by l7_map) in the different profiles.
To avoid this situation, I made the info for the same client IP shared by multiple profiles in the same thread.
I located the shared IP info at CFlowGenListPerThread and it would be resolved by CTupleGeneratorSmart which instance is bounded to profile per thread.

### DIFF
--- a/src/bp_sim.cpp
+++ b/src/bp_sim.cpp
@@ -2955,6 +2955,8 @@ bool CFlowGenListPerThread::Create(uint32_t           thread_id,
     m_node_gen.m_socket_id =socket_id;
     m_flow_id_to_node_lookup.Create();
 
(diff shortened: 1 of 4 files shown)


hhaim:
rebased

hhaim:
@jsmoon have a look into this https://github.com/cisco-system-traffic-generator/trex-core/commit/d7c1099a6eb70703a28ebce7d1334e5cce786db2

It supports topo for all profiles (shared) with your current commit it make more sense now 

Please review

Issue #310, opened 2019-07-22
TRex crash with '--arp-refresh-period 0' option


egwakim:
According to the TRex help, 'arp-refresh-period 0' will 'never send' GARP, but, TRex aborted before starting TRex.

--arp-refresh-period       : Period in seconds between sending of gratuitous ARP for our addresses. Value of 0 means 'never send'

Reproduction>
./t-rex-64 -i --arp-refresh-period 0

assert: ../../../../trex-core/src/trex_modes.h:476 CDpdkModeBase* CDpdkMode::get_mode() Assertion 'm_dpdk_obj' failed.

*** traceback follows ***

1             0x540ada __assert_fail + 650
2             0x52204d ./_t-rex-64() [0x52204d]
3             0x537bf9 ./_t-rex-64() [0x537bf9]
4             0x537e08 ./_t-rex-64() [0x537e08]
5             0x538fd1 main_test(int, char**) + 113
6       0x7f053f469b97 __libc_start_main + 231
7             0x4dd8e9 _start + 41

*** addr2line information follows ***

ostream:561
trex_modes.h:476
main_dpdk.cpp:944 (discriminator 2)
main_dpdk.cpp:1025
main_dpdk.cpp:6378
??:0
??:?

PR #551, opened 2020-10-23
Enable trex for Intel I211 Gigabit NIC

### LINKED ISSUE #550: Enable trex for Intel I211 NIC
Intel I211 NIC has only two combined queues as shown below.
```bash
$ sudo ethtool -l enp1s0
Channel parameters for enp1s0:
Pre-set maximums:
RX:             0
TX:             0
Other:          1
Combined:       2
Current hardware settings:
RX:             0
TX:             0
Other:          1
Combined:       2
```

If you run trex on it you'd get the following error.

```bash
$ sudo ./t-rex-64 -f cap2/dns.yaml -d 10 -m 1000
The ports are bound/configured.
Starting  TRex v2.84 please wait  ...
 set driver name net_e1000_igb
 driver capability  : TCP_UDP_OFFLOAD  TSO
 set dpdk queues mode to DROP_QUE_FILTER
 Number of ports found: 2
zmq publisher at: tcp://*:4500
Ethdev port_id=0 nb_tx_queues=4 > 2
EAL: Error - exiting with code: 1
  Cause: Cannot configure device: err=-22, port=0
```

Which is caused by the following code.
```c++
    // One q for each core allowed to send on this port + 1 for latency q (Used in stateless) + 1 for RX core.
    m_max_queues_per_port = m_cores_to_dual_ports + 2;
```

However, if we don't need features like latency metric and arp refresh, the `m_max_queues_per_port` should be 1, and trex could run on I211.

WqyJh:
TRex crash during port configuration because the `m_max_queues_per_port`
was set to `m_cores_to_dual_ports + 2`, which would have a minimum
number of 3, greater than 2 queues of I211.

TRex also crash when `--arp-refresh-period 0` was specified because one
of the `get_dpdk_mode()` was called before choosing a dpdk mode.

Now TRex was able to run with I211 without RX Core.

Fix #550 #310

### DIFF
--- a/src/main_dpdk.cpp
+++ b/src/main_dpdk.cpp
@@ -973,8 +973,7 @@ COLD_FUNC static int parse_options(int argc, char *argv[], bool first_time ) {
     }
 
     if (po->preview.get_is_rx_check_enable() ||  po->is_latency_enabled() || CGlobalInfo::is_learn_mode()
-        || (CGlobalInfo::m_options.m_arp_ref_per != 0)
-        || (!get_dpdk_mode()->is_hardware_filter_needed()) ) {
+        || (CGlobalInfo::m_options.m_arp_ref_per != 0)) {
         po->set_rx_enabled();
     }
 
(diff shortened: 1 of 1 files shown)


hhaim:
See the answer on the issue

Issue #538, opened 2020-10-08
Traffic rate mismatch issue


egwakim:
Hi,
In high loaded situation.
TRex traffic timing mismatch with actual time.
If temporary traffic burst exist during traffic generation, in spite of average load is not high, traffic will not be generated in wanted traffic ratio.
Virtual time in DP scheduler will be differ from real actual time.

It's quite crucial problem when user want to send traffic with exact speed.
User cannot send traffic with wanted traffic rate if eventual traffic burst exist during traffic generation.

There could be several solutions for it,
One solution would be recovering it when CPU is idle, process delayed packets and recover virtual time to actual time.
It'll be reasonable approach when the traffic burst is just temporary.

The disadvantage of this approach is,
In the long overloaded situation, such as high load with long duration time without idle time, we cannot recover it or it'll cause big time leap when CPU is idle again.

What do you think about this issue?
Do you have any suggestion about it?

Best Regards
Gwangmoon

hhaim:
@egwakim you are right, In case the speed asked can't be provided there is no compensation and this is by design, it is not a bug. There is a conflict between the speed and total of bytes and we try to send the total bytes asked 
This happen in extrem cases that the user asked for more that it is possible.

Take this example 

* 10gbps port 
* 0..10sec  burst of 20gbps  (stream) total bytes 10*20g/8
* 10..20sec  : 0 gbps   total bytes 0 

In 0..10 seconds the time will be stretched to 20 seconds sending **only 10gbps**  because this is what it possible to do (the asked total bytes is OK (10*20g/8)
**then** there would be a wait for 10 seconds (it won't compensate for the burst in 0..10) 
The average speed would be 6gbps instead of 10gbps 

I don't see why we should conpensate as the user ask for a time with 0 gbps ..

egwakim:
Hi @hhaim ,
I'm mentioning about the issue when 'short-term' burst case.
For example,
If we have 10 gbps port, sending traffic with average 1gbps and multiple traffic profiles runs up and down during transmission.
But, traffic rate burst in very short-term period like several mili-seconds(The worst case peak throughput 2gbps, CPU load 100%), this profile's time will be delayed in spite of average traffic ratio is under 10% and CPU is idle after burst.

In this case, user want to send traffic with exact traffic rate, but, user can not send traffic in spite of both port capacity and average CPU capacity is quite enough.

hhaim:
[...]

PR #553, opened 2020-10-26
add sched_max_stretch configuration for #538

jsmoon:
@hhaim this PR is for #538 discussion.
New "get_global_cfg" and "set_global_cfg" RPC commands are added.

### DIFF
--- a/scripts/automation/trex_control_plane/interactive/trex/common/trex_client.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/common/trex_client.py
@@ -1960,6 +1960,40 @@ class TRexClient(object):
 
 
     @client_api('command', True)
+    def set_global_cfg (self, params):
+        """
+            Change global configuration parameters.
+
+            :parameters:
+                params: dictionary
+                    dictionary of global configuration parameters. Available parameters:
+                    "sched_max_stretch", type = double, default =0.0, units = usec, scheduler's maximum time for stretch in case it is a high value there won't be a scheduler compensation on burst and time will not be stretch. 0.0 means use the default internal value (~100usec). value could not be lower than 100usec will be considered as 100usec.
+
+            :raises:
+                + :exc:`TRexError`
+
+        """
+
+        self.ctx.logger.pre_cmd("Changing global configuration parameters")
+
+        rc = self._transmit("set_global_cfg", params = params)
+
+        self.ctx.logger.post_cmd(rc)
+
+        if not rc:
+            raise TRexError(rc)
+
+
+    @client_api('command', True)
+    def get_global_cfg (self):
+        rc = self._transmit("get_global_cfg")
+        if not rc:
+            raise TRexError(rc)
+
+        return rc.data()
+
+
+    @client_api('command', True)
     def push_packets (self, pkts, ports = None, force = False, ipg_usec = 0):
         """
             Pushes a list of packets to the server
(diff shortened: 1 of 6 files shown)


hhaim:
@jsmoon almost

1. 

This original code 
```
 while (state!=scTERMINATE) {

         switch (state) {
         case scINIT:
            cur_time = now_sec();
            {
                dsec_t dt = cur_time - n_time ;

                if (dt > BURST_OFFSET_DTIME) {
                    state = scSTRECH;
                } else if (dt > 0) {
                    state = scWORK;
                } else {
                    state = scWAIT;
                }

            }

```

should be replaced to be 

```
 while (state!=scTERMINATE) {

         switch (state) {
         case scINIT:
            cur_time = now_sec();
            {
                dsec_t dt = cur_time - n_time ;

                if (dt > **m_burst_size**)  {  <<<<<<<<<<<<<<<<<<<
                    state = scSTRECH;
                } else if (dt > 0) {
                    state = scWORK;
                } else {
                    state = scWAIT;
                }

            }

```


Your change in ```handle_time_strech``` is not good, as in the scheduler ```   if (dt > BURST_OFFSET_DTIME) {``` this will be true, it will get to ```handle_time_strech``` and it will do nothing, it will be a waste of loops to get in and out of STRECH state without doing anything 

 it is better **not** to get to it at all (and remove this function change )

```
oid CNodeGenerator::handle_time_strech(CGenNode * &node,
                                        dsec_t &cur_time,
                                        dsec_t &n_time,
                                        dsec_t &offset,
                                        CFlowGenListPerThread *thread) {
```



2. Update the help string 

```
    def set_global_cfg (self, params):
        """
            Change global configuration parameters.

            :parameters:
                params: dictionary
                    dictionary of global configuration parameters. Available parameters:


                    "sched_max_stretch", type = double, default =0.0, units = usec, scheduler's maximum time for stretch in case it is a high value there won't be a scheduler compensation on burst and time will not be stretch. 0.0 means use the default internal value (~100usec). value could not be lower than 100usec will be considered as 100usec. 

            :raises:
                + :exc:`TRexError`

        """
```
2. We should not set ```m_burst_offset_dtime``` smaller than  BURST_OFFSET_DTIME (~ 100usec )
else the scheduler will be very **not efficient** 

```
   #define BURST_OFFSET_DTIME    (100.0/1000000)
```

so this code should protect against too very low value s 

```
    if (CGlobalInfo::m_burst_offset_dtime > 1.0 ) {
        if (CGlobalInfo::m_options.preview.getVMode() > 3) {
            std::cout << "burst_size = " << m_burst_size << std::endl;
        }
          m_burst_size = BURST_OFFSET_DTIME;

       if CGlobalInfo::m_burst_offset_dtime>BURST_OFFSET_DTIME {
          m_burst_size = CGlobalInfo::m_burst_offset_dtime;
       }
}

```

jsmoon:
1. `handle_time_strech` performs additional `handle_maintenance` to prevent watchdog. So, if `m_burst_size` is large, it may cause watchdog.
2. I'll apply it.
3. Since the `handle_time_strech` is performed only when `dt > BURST_OFFSET_DTIME`, too low value is protected implicitly.

hhaim:
@jsmoon understood, we have been mixing two things what about this?

```            

..
            cur_time = now_sec();
            {
                dsec_t dt = cur_time - n_time ;

                // real time clock  (not streched)
                dsec_t dtr = cur_time - m_last_sync_time_sec ;
                if ( dtr > SYNC_TIME_OUT){
                    handle_maintenance_job()
                }

                if (dt > m_burst_size ) {
                    state = scSTRECH;
                } else if (dt > 0) {
                    state = scWORK;
                } else {
                    state = scWAIT;
                }
            }



handle_maintenance_job() {
                    handle_maintenance(thread);
                    node = m_p_queue.top();
                   n_time = node->m_time + offset;

}


```

The problem with your patch that in case you go to ```handle_time_strech ``` and **not** fixing the time_offset, you will get to this function again and again (every 15usec) without doing anything,

This way we disconnect the relation between stretching the time and watchdog handling 


and ```handle_time_strech ``` will do only explicit stretch and not watchdog 

```
void CNodeGenerator::handle_time_strech(CGenNode * &node,
                                        dsec_t &cur_time,
                                        dsec_t &n_time,
                                        dsec_t &offset,
                                        CFlowGenListPerThread *thread) {


    /* fix the time offset */
    dsec_t dt = cur_time - n_time;
    offset += dt;
    /* set new offset */
    m_scheduler_offset = offset;
}
```

hhaim:
@jsmoon, looks good. 
thanks

hhaim:
rebased.

PR #554, opened 2020-10-27
secure updating DP profile ctx for stats.

jsmoon:
Hi, this PR is to secure DP profile ctx when build has been finished.
Current updating method (direct call of `ctx->get_profile_ctx()`) should secure the safe condition to avoid racing condition.
It may cause DP profile ctx is not updated until all profiles are in the stable state.
This change make it sure after the profile build finished from DP using by a new DP to CP message, `TrexDpCoreProfileCtx`.

### DIFF
--- a/src/44bsd/sim_cs_tcp.cpp
+++ b/src/44bsd/sim_cs_tcp.cpp
@@ -784,7 +784,8 @@ void CClientServerTcp::dump_counters(){
     stt_cp.m_init=true;
     stt_cp.Add(TCP_CLIENT_SIDE,&m_c_ctx);
     stt_cp.Add(TCP_SERVER_SIDE,&m_s_ctx);
-    stt_cp.update_profile_ctx();
+    stt_cp.AddProfileCtx(TCP_CLIENT_SIDE,DEFAULT_PROFILE_CTX(&m_c_ctx));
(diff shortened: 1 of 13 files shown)


hhaim:
@jsmoon could you elaborate? There was an old way to protect the counters, did you remove it? Did you find a crash in the old way?

hhaim:
It is better to do this operation from the CP not from DP. 


```
bool
TrexDpCoreProfileCtx::handle(void) {
    get_stx()->add_dp_profile_ctx(m_profile_id, m_client_pctx, m_server_pctx);
    get_stx()->dp_core_finished(m_thread_id, m_profile_id);
    return true;
}
```

jsmoon:
@hhaim, my user added short duration profiles continuously. In this situation, some profiles did not get updated statistics for a while.

The old way is to get DP profile ctx safely from CP side. CP couldn't get the safe period for a few seconds. So I remove the code to wait safe period. Instead, getting DP profile ctx from DP side is added.

`TrexDpCoreProfileCtx` is DP to CP message. So, `TrexDpCoreProfileCtx::handle` is called from CP.

hhaim:
@jsmoon understood. looks good.

hhaim:
rebased

PR #555, opened 2020-10-27
Added ip_header_offset yaml configuration for STF IP tunneling support

jhexan:
The idea here is to allow the user to configure what the inner/subscriber IP layer offset is so that pcaps with IP tunneling are properly processed. The other option would have been to add various tunneling parsing support, but that approach is more complex and perhaps overkill for STF (since the mode is no longer actively being developed).

### DIFF
--- a/doc/trex_book.asciidoc
+++ b/doc/trex_book.asciidoc
@@ -1963,6 +1963,7 @@ see xref:timer_w[Timer Wheel section]
        multi_flow_enabled: true     <8>
        flows_dirs: [0, 1]           <9>
        keep_src_port: true          <10>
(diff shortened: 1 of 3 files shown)


hhaim:
@jhexan is it enough to have the **same** tunnel-id for the same template? 
How do you add more tunnels-id, is it by adding multiple templates with different tunnel-id.
I think it is better to invest on ASTF for the long run

jhexan:
@hhaim Yes, different templates can have different tunnel characteristics, but in our use case we just need the traffic to be tunneled. We are using both STF and ASTF for different purposes.

hhaim:
rebased

PR #561, opened 2020-11-03
Bugfix/scapy server after update

EgorBlagov:
Fixing Scapy Server issues

After scapy update some property names, field types and behaviors has changed, it caused errors in Scapy Server


hhaim:
rebased.

Issue #560, opened 2020-11-03
ASTF Tuneable issue


jsmoon:
@hhaim, according to my examination, some tuneable have issues.
1. mismatching between document and implementation: `ip.dont_use_inbound_mac` and `tcp.no_delay_counter` don't work for per-template. Please see CFlowTemplate::server_update_mac() and count_and_check_no_delay() each.
2. `tcp.no_delay_counter` does not reset by `reset_tuneables()`.
3. `scheduler.accurate = False` has the same effect on `scheduler.accurate = True`.

I'd like to discuss how they could be handled.

hhaim:
@jsmoon thanks for looking into this. BTW who is doing this excellent QA?
1. Fix the doc or add the support 
2. Need to reset, it is a bug
3. Don't remember what I wanted to do there, I think there was two versions in my head (for ramp-up algo) and I ended up with one  (not accurate, update every 1 sec) , so better to remove this from doc and code as there was not a need for accurate (timer is working per 20usec, but rampup is updated every 1 sec)

jsmoon:
@hhaim, I'd like to discuss one more thing.
As I added the document at #562, the current implementation is that only the first profile's global tuneable can be applied. But, the global IP tunable is available for each profile.
I think about making the TCP global tunable also available for each profile. It can run multiple clients and servers with different TCP parameters.
What do you think about it?

hhaim:
@jsmoon I don't see a reason why not to do it. Multiple profile should behave as any profile.

jsmoon:
@hhaim I found additional issues during source code analysis.

1. `tcp.keep*` tunable default values are not matched to implementation.
   for example, the `tcp_keepinit` variable has an initial value of 10. But it is the counter of TCP slow timeout (= 500ms). So, the default value of `tcp.keepinit` would be 5, but the document is 10.

2. `tcp.delay_ack_msec` will not update proper `tcp_slow_fast_ratio` value.
   `tcp_slow_fast_ratio` is calculated by `tcp_fast_ticks` timer tick counter instead of msec value.
   ```
        tcp_fast_ticks =  tw_time_msec_to_ticks(tune->m_tcp_delay_ack_msec);
        tcp_slow_fast_ratio = _update_slow_fast_ratio(tcp_fast_ticks);
   ```
   When `tcp.delay_ack_msec` is 20ms, `tcp_slow_fast_ratio` will be 1 and the TCP keepalive timer will expire in short time.

[...]

PR #573, opened 2020-11-17
ASTF global tunable works per profile

jsmoon:
Hi, this PR is an implementation for per-profile global tunable discussion at #560.
`CTcpPerThreadCtx` tunable variables are moved to new `CTunableCtx` class.
As added to the document, the `tcp.blackhole` tunable will be used from any profile in an epoch.


hhaim:
@jsmoon nice.

hhaim:
rebased

Issue #572, opened 2020-11-16
ASTFProgram recv issues


jsmoon:
Hi,
my user wants to transfer TCP traffic more than 4Gbytes on a flow.
Since `recv` has the 32bit value limit, I made it called in the loop using by `jmp_nz`.
To make it work properly, `recv` should be used with `clear=True` due to the watermarking mechanism effect in the loop.

But, after the **non-block send** is used, another issue has been observed.
The **non-block send** triggers additionally received bytes when `recv` is called.
`clear=True` can remove them unexpectedly.
```
    inline void check_rx_condition(){
        if (m_cmd_rx_bytes>= m_cmd_rx_bytes_wm) {
            if (get_rx_clear()){
                m_cmd_rx_bytes=0;
                set_rx_clear(false);
            }
            EMUL_LOG(0, "ON_RX [%d]- NEXT \n",m_debug_id);
            next();
        }   
    }
```
The next `recv` cannot receive expected bytes. This makes user ASTF programming very tough.

@hhaim can you give me your opinion regarding this issue?

hhaim:
@jsmoon have a look into the http_eflow3.py profile. 

```
 if r[1]==0:
          prog_c.recv(r[0])
        else:
            prog_c.set_var("var1",r[1]); # 
            prog_c.set_label("a:");
            prog_c.recv(r[0],True)
            prog_c.jmp_nz("var1","a:") # dec var "var2". in case it is *not* zero jump a: 
            if r[2]:
               prog_c.recv(bsize*r[2])
```

Maybe it would be better to change rx to 64bit

jsmoon:
@hhaim I will change it to 64bit. Do I need to change the `recv_msg` also?

Anyway, I think the watermark mechanism issue in the loop needs to be solved also.
From the `recv` watermark mechanism and the loop, the user should fulfill the following restrictions.
```
    #prog_s.set_send_blocking(block=False)    # --  (1)
    prog_s.send('#'*first_size)
    prog_s.set_var("var1", loop)
    prog_s.set_label("a1:")
    prog_s.send('*'*size)                     # -- (2)
    prog_s.jmp_nz("var1", "a1:")
    prog_s.send('$'*last_size)

    prog_c.recv(first_size, clear=True)       # -- (3)
    prog_c.set_var("var1", loop)
    prog_c.set_label("a1:")
    prog_c.recv(size, clear=True)             # -- (4)
    prog_c.jmp_nz("var1", "a1:")
[...]

PR #575, opened 2020-11-18
ASTF recv/recv_msg available in loop without clear

jsmoon:
Hi, this PR is another implementation for #572.
Comparing to #574, I removed the stacked watermark mechanism during receive on the server.
There is no change needed in the client.

If you select this PR, PR #574 will be discarded.

### DIFF
--- a/scripts/astf/ftp_multiple_flows_sim.py
+++ b/scripts/astf/ftp_multiple_flows_sim.py
@@ -40,20 +40,18 @@ class ftp_sim():
         if tinc != 0:
             prog_c.delay_rand(0,tinc*1000*1000)  # delay from 0 to random(tinc) seconds
         prog_c.send(self.ftp_get)
-
-        prog_c.set_var("var1",loop); #
-        prog_c.set_label("a:");
-        prog_c.recv(self.max_buf_size,True)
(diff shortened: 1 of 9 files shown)


hhaim:
@jsmoon looks good. Need to fix the documentation  and the udp. rcv_msg too (which is in uint32_t packets)

for example 
```
    def recv(self, bytes,clear=False):
        """
        recv (bytes)

        :parameters:
                  bytes  : uint32_t  <<<<<<<<<<<<<<<< this need to be fixed to uint64
                   wait until the rx bytes watermark is reached on flow counter.  

                  clear  : bool
                     when reach the watermark clear the flow counter 
        """
```

2. I would fix the profiles http_eflowx.py 

This code could be simplified as there is no limit of uint32 bits  

    def calc_loops (self,buffer,loops):
        max_mul = int(round(0xffffffff/buffer)/4)
        div = loops/max_mul;
        if div<1.0:
            return (loops*buffer,0,0)

        res = (max_mul*buffer,int(div),loops-(int(div)*max_mul))
        expected = buffer*loops
        assert(expected==res[0]*res[1]+buffer*res[2])
        return (res)

jsmoon:
@hhaim I'll prepare for your suggestion.

hhaim:
rebased

Issue #569, opened 2020-11-12
STL profile unexpected delayed start possibility


jsmoon:
@hhaim, my user observed a delayed start in STL mode.
He started a profile by start_traffic with 1-sec duration. But it seemed to start sending packets after several seconds. 
So, I checked the source and found a possibility.

This code is from DP start_traffic.
```
    /* update cur time */
    if ( CGlobalInfo::is_realtime() ){
        m_core->m_cur_time_sec = now_sec() + schd_offset;
    }
```
Since `CGlobalInfo::is_realtime()` is always true, these streams will be scheduled real-time(`now_sec()`) basis.
When DP scheduler time is stretched, the streams will be dispatched and performed after the stretched time passed.

```
DP scheduler dispatch time               now_sec()
                  | <-- stretched time --> |
                  v                        v
------------------*-------------------------*--------------------------
                                            ^
                                            |
                         new stream scheduled here
```

Do you have any idea to avoid this issue?

hhaim:
@jsmoon what happen when you disable "stretched time "? does it solve the issue?

jsmoon:
@hhaim definitely it could solve this issue. My user can avoid this issue by disabling it.
Since I found this possibility by source code analysis, I'd like to get a more nice solution.
 
Is there any reason to update `m_cur_time_sec` in here?

hhaim:
In regards to this:
```   m_core->m_cur_time_sec = now_sec() + schd_offset;```
There is a reason. ```m_cur_time_sec``` is used to give an offset between DP cores (pipeline)
So when traffic is started it is possible to offset the timer in such way that it won't start at the same time on all the cores. 
It is not related to stretch time (BTW when this code have been written there was only **one** profile at a time so when we started the streams stretch was zero) 

In your case you see the issue when you have: 

1. one running profile that somehow stretch the time
2. start another profile

**Comment by jsmoon** [comment]
Yes, the possibility exists in multiple profiles.

But, I don't think this code gives the offset between DP cores due to the constant `schd_offset` value.
```
    double schd_offset = get_dpdk_mode()->dp_rx_queues()?
        SCHD_OFFSET_DTIME_RX_ENABLED:
        SCHD_OFFSET_DTIME;
```

As you mentioned, if it is not related to the stretch time, I think the stretch time effect needs to be removed.
In ASTF, there is no update of `m_cur_time_sec` while the new profile is added.

hhaim:
[...]

PR #576, opened 2020-11-19
remove STL profile delayed start possibility

jsmoon:
Hi, this PR is an implementation for #569.
I tried to keep backward compatibility.

### DIFF
--- a/src/stx/stl/trex_stl_dp_core.cpp
+++ b/src/stx/stl/trex_stl_dp_core.cpp
@@ -1304,7 +1304,7 @@ TrexStatelessDpCore::add_stream(PerPortProfile * profile,
         node->m_state =CGenNodeStateless::ss_INACTIVE;
     }
 
-    if ( unlikely(start_at_ts) ) {
+    if ( likely(start_at_ts) ) {
         node->m_time = start_at_ts + stream->get_start_delay_sec();
     } else {
         node->m_time = m_core->m_cur_time_sec + stream->get_start_delay_sec();
@@ -1487,7 +1487,14 @@ TrexStatelessDpCore::start_traffic(TrexStreamsCompiledObj *obj,
 
     /* update cur time */
     if ( CGlobalInfo::is_realtime() ){
-        m_core->m_cur_time_sec = now_sec() + schd_offset;
+        double cur_time = m_core->m_cur_time_sec;
+        if (m_state != TrexStatelessDpCore::STATE_TRANSMITTING) {
+            cur_time = now_sec();
+            m_core->m_cur_time_sec = cur_time + schd_offset;
+        }
+        if (!start_at_ts) {
+            start_at_ts = cur_time + schd_offset;
+        }
     }
 
     /* no nodes in the list */


hhaim:
rebased. 
Don't you want the same in ASTF?

jsmoon:
@hhaim No same issue in ASTF.

hhaim:
@jsmoon I've found an issue on slow setup
```
        if (!start_at_ts) {
            start_at_ts = cur_time + schd_offset;
        }
```

This code change the timing in case of single profile with start of two ports (on the same core).
The second time the start_ts is not current time it is current +schd_offset. 
I don't think that start_at_ts is useful and not sure why you override it.
removing this code fix the issue on those setups

jsmoon:
@hhaim since `add_stream` refers the `start_at_ts` like as following.
```
    if ( likely(start_at_ts) ) {
        node->m_time = start_at_ts + stream->get_start_delay_sec();
    } else {
        node->m_time = m_core->m_cur_time_sec + stream->get_start_delay_sec();
    }
```
I don't think it breaks the backward compatibility.
As far as I understood, `start_at_ts` came from the user's request.
So, I thought I can use it when the user does not specify it.

hhaim:
You are right in case ```m_state != TrexStatelessDpCore::STATE_TRANSMITTING```
The issue is in case ```m_state == TrexStatelessDpCore::STATE_TRANSMITTING`` which happen on the second profile on the same core (port 1 start after port 0 start), there is no need to have multi-profile to get to this issue. 

look into this example the duration of the test is 0.1 sec which make the test fail 
```
transmit | cur_time | offset | node-time 
=================================
  false  |   7      |  0.2   |  7+0.2
  true   |   8      |  0.2   |  8 (your code it is 8+0.2) 
```
In the second line the node-time would be 7 and with your code it would be cur_time + offset 

In the old code, we have updated the cur_time with now() and this is not the case anymore.

jsmoon:
@hhaim I missed the case.
I think the following code change would remove this issue.
```
--- a/src/stx/stl/trex_stl_dp_core.cpp
+++ b/src/stx/stl/trex_stl_dp_core.cpp
@@ -1490,7 +1490,7 @@ TrexStatelessDpCore::start_traffic(TrexStreamsCompiledObj *obj,
         double cur_time = m_core->m_cur_time_sec;
         if (m_state != TrexStatelessDpCore::STATE_TRANSMITTING) {
             cur_time = now_sec();
-            m_core->m_cur_time_sec = cur_time + schd_offset;
+            m_core->m_cur_time_sec = cur_time;
         }
         if (!start_at_ts) {
             start_at_ts = cur_time + schd_offset;
@@ -1530,6 +1530,9 @@ TrexStatelessDpCore::start_traffic(TrexStreamsCompiledObj *obj,
 
 
     if ( duration > 0.0 ){
+        if (start_at_ts > m_core->m_cur_time_sec) {
+            duration += (start_at_ts - m_core->m_cur_time_sec);
+        }
         add_port_duration( duration ,obj->get_port_id(), profile_id, event_id );
     }
 
```
Since `add_port_duration` refers to `m_core->m_cur_time_sec` internally, the adjustment based on node-time(start_at_ts) is needed to keep the exact duration.

jsmoon:
@hhaim I'm sorry for your inconvenience.
I know you already solved this issue.
Since you removed setting `start_at_ts`, I think you need to revert `likely(start_at_ts)` to `unlikely(start_at_ts)` also.

Anyway, what do you think about the change cf31f92?

PR #584, opened 2020-11-30
STF tunneling ip variation (max_ip_tunnels)

jhexan:
I've added a new configuration, max_ip_tunnels, to control the number of unique tunnels by changing the outer IP addresses based on the inner source IP. To calculate the outer IPs I'm using:

tunnel IP = template tunnel IP + (current inner source IP % max_ip_tunnels)

### DIFF
--- a/doc/trex_book.asciidoc
+++ b/doc/trex_book.asciidoc
@@ -1964,6 +1964,7 @@ see xref:timer_w[Timer Wheel section]
        flows_dirs: [0, 1]           <9>
        keep_src_port: true          <10>
        ip_header_offset: 50         <11>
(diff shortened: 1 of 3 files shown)


hhaim:
@jhexan looks good. I see that you are still using the old STF mode. Do you have a plan to move to ASTF?

hhaim:
rebased

jhexan:
We’re using both modes, and still have use cases for STF.

sent from my iphone

hhaim:
So what do you do in case of tunnel and ASTF?

jhexan:
That’s the next thing I’ll be taking a look at :)

PR #587, opened 2020-12-02
add an option to disable mlx4 or mlx5 driver only and fix mlx5 configuration file

aleks-kozyrev:
Fix compilation issues with MLX_OFED 5.1 and higher.

### DIFF
--- a/linux_dpdk/ws_main.py
+++ b/linux_dpdk/ws_main.py
@@ -202,6 +202,123 @@ def missing_pkg_msg(fedora, ubuntu):
         msg += unknown_install
     return msg
 
+@conf
+def configure_mlx5 (ctx):
+    ctx.start_msg('Configuring MLX5 autoconf')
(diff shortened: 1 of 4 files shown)


hhaim:
@aleks-kozyrev 
1) it is better to make it optional to add mlx4 driver (e.g. --no-mlx4)
2) I would keep this or make it optional too (we tune it to be build on specific distro and mlx5 header) so it is needed by default 
#ifndef HAVE_IBV_DEVX_EVENT
#define HAVE_IBV_DEVX_EVENT
#endif

#ifndef HAVE_IBV_VAR
#define HAVE_IBV_VAR
#endif

#ifndef HAVE_MLX5DV_DR
#define HAVE_MLX5DV_DR
#endif

hhaim:
@aleks-kozyrev looks good. 
I've removed the auto generated header from our git. 
It works fine with an installed OFED system but does not work on system without OFED. I've fixed it. I'm running our full regression now.

hhaim:
rebased

Issue #585, opened 2020-12-01
ASTF Tunable min-max values in the document


jsmoon:
Hi, 
https://trex-tgn.cisco.com/trex/doc/trex_astf.html#_tunables_reference_a_id_tunables_a
The following min-max values are different from the source code.

tunable | min-max (document) | min-max (source code)
-- | -- | --
tcp.rxbufsize | 1K-1M | 1K-1G
tcp.txbufsize | 1K-1M | 1K-1G
tcp.delay_ack_msec | 20-500 | 20-50000

I'd like to synch up the document by the source code.
But `tcp.delay_ack_msec` is used for TCP timer tick also, I think that the 50 sec from the source code is too large.

@hhaim, would you give your opinion about this mismatch?

hhaim:
@jsmoon 50sec for delay_ack is for scale test of 50M active flows. In this case the timer take 100% of the timer so there is a need to make them longer

jsmoon:
@hhaim, I'll change it in the document also.

PR #588, opened 2020-12-02
update document ASTF Tunable min-max values

jsmoon:
@hhaim this PR is the change for #585.


hhaim:
rebased

PR #595, opened 2020-12-10
Small tunneling IP bug fix

jhexan:
Didn't convert the tunnel IPs form network to host order before incrementing them, which resulted in flipped addresses.


hhaim:
@jhexan please build a new clean PR based on master

PR #596, opened 2020-12-11
fix ASTFProgram command parameter type mismatch

jsmoon:
Hi, I found the ASTFProgram API document is mismatched to the implementation.
For example, the document is ...
image
The implemented code is
```
uint32_t CAstfDB::get_delay_ticks(uint32_t program_index, uint32_t cmd_index) {
    Json::Value cmd;

    try {
        cmd = m_val["program_list"][program_index]["commands"][cmd_index];
    } catch(std::exception &e) {
        assert(0);
    }

    assert (cmd["name"] == "delay");

    return tw_time_usec_to_ticks(cmd["usec"].asInt()); // <-- int32_t
}
```

I think changing implementation is better to match them.


hhaim:
rebased

PR #580, opened 2020-11-24
Fix wait_for_peer_close is not working.

jsmoon:
Hi, this PR is a change for #578.

@hhaim please check my changes and give your feedback.

PR #603, opened 2020-12-17
Fix issue that non-blocking send wakes up following blocked commands

jsmoon:
Hi, this PR would amend the previous fix at PR #580.
It had fixed the issue for the `wait_for_peer_close` only.
But there is the same issue on the `recv`, `delay` and `delay_rand` also.


hhaim:
rebased

Issue #606, opened 2020-12-24
ASTF traffic duration is shorter than expected


jsmoon:
Hi, I recently found that my ASTF traffic duration is much shorter than expected.

I have programmed to send UDP traffic 9 packets per second for 120 seconds.
The client sends the 3 packets and receives 3 packets at first. And then it sends 1080 packets for 120 seconds.
The server receives the first 3 packets and sends 3 packets. And then it waits for 120 seconds and receives all packets.
The send rate and waiting are implemented by the `delay()` command.
The send rate uses `delay(111100)` and `delay(111200)`. The waiting uses `delay(120*1000*1000)`.

As you can see in the following result, the client's traffic duration is less than expected.
```
     m_traffic_duration |        116.12 sec |           120 sec | measured traffic duration 
                      - |                   |                   |                         
                    TCP |                   |                   |                         
                      - |                   |                   |                         
                      - |                   |                   |                         
                    UDP |                   |                   |                         
                      - |                   |                   |                         
           udps_accepts |                 0 |                 1 | connections accepted    
          udps_connects |                 1 |                 0 | connections established 
            udps_closed |                 1 |                 1 | conn. closed (includes drops) 
           udps_sndbyte |           1474596 |                36 | data bytes transmitted  
            udps_sndpkt |              1083 |                 3 | data packets transmitted 
           udps_rcvbyte |                36 |           1474596 | data bytes received     
            udps_rcvpkt |                 3 |              1083 | data packets received   
                      - |                   |                   |                         
```
I think this error rate (over 3.23%) is **quite large for my traffic simulation**.

According to my analysis, this issue is caused by the level 1 timer tick unit.
The level 0 timer tick unit is 0.02 msec. It covers 0 ~ 40.96 msec by 2048 wheel size.
And the level 1 timer tick unit is 5.12 msec. It means that the maximum error rate could be 12.5%.

I need a more precise action for the `delay()` timer.
I think there are 2 approaches to solve this issue.
1. lower the level 1 timer error rate using the larger `level1_div` value. (8 -> 16, 32, 64, ...)
2. remove the level 1 timer error rate by starting the level 0 timer with `m_ticks_left`.

**I prefer solution 2** because solution 1 still has quite a large error rate.
But the solution 2 will cause one more timer action. The most timer callbacks will be performed by the level 0 timer.

@hhaim, what do you think about this issue? Do you have any idea to solve this issue?

hhaim:
[...]

PR #607, opened 2021-01-04
Improved ASTF delay command accuracy

jsmoon:
Hi, this PR is a change for issue #606.
No other timer works except `delay` command are not affected by this change.
I think that this approach is better than changing `timer_start`.


hhaim:
@jsmoon this is not good for the same reason, TCP counters can't work with that and it misses the all points. 
With this idea, you won't be able to reach the scale of flows (e.g. 30M active flows). Each timer_start should have exactly **one** tick until callback.   Suggest to fix it in the timer_wheel with the random number, it should be one or two lines of code.  
I suspect that we can add this random to any timer_start without noticeable impact

jsmoon:
@hhaim I just touched the `delay` command only. I couldn't understand why my change affects TCP counters and how `delay` command has a relationship with the scale of flows. My change is the same as the app's accumulation of delay.

> Suggest to fix it in the timer_wheel with the random number, it should be one or two lines of code.
I suspect that we can add this random to any timer_start without noticeable impact

I have no idea of this suggestion on how can solve the programming traffic duration issue.

hhaim:
To reach 50M active flows there are few things that need to be "tune" 
1. TCP delay-ack timer should be in range of a few seconds instead of 50msec 
2. keepalive timer should be enlarge 
3. Application should have a very long delay (e.g. 200 sec) 
With your fix it will be impossible to reach 50M active flows 

>> The idea is to calculate the error and add one in the probability of the range. Example:
Let's say you you need 16.1 ticks, but choose 16 (error 0.1/1) each time you ask 16.1 you throw a random coin, 1 out of 10 you will choose 17 instead of 16. In many timer starts it will give exact accumulation/average on many starts of with the same error -- the same example you gave 
we can try adding this by default . I've provided a link to the place to add this.

hhaim:
@jsmoon if it is too complicate I can provide the patch, let me know

jsmoon:
@hhaim do you mean that long delay command is used to accomplish 50M active flows? So the delay timer performance may affect the scale of flows? I understood what you had been worried about now.
But I worried about the single delay error also (not only rate but also absolute value). If I understood correctly, your suggestion doesn't help in that case.
So, if you don't mind, I'd like to add an `accurate` option to the `delay` command. I think users can select accuracy or performance with this approach. What do you think about this?

hhaim:
@jsmoon You gave an example that can be solved using a simple change, if this is a real requirement let's fix it as I suggested. 
Regarding a single, accurate timer, can you give an example when it is important? Why if someone would ask for 100 sec delay with 20 usec resolution? I still don't get it.  our main use-case is scale. 
now even if I be conviced that it is important, I think you should add it to the timer_w and not as you did.

let's take a 100sec timer. in your case it would have 5000 ticks while with it could be implemented using ~3 ticks with the same accuracy (level2->level1->level0) so it should be part of timer_w code and not instruction

jsmoon:
@hhaim

> You gave an example that can be solved using a simple change, if this is a real requirement let's fix it as I suggested.

I think the `delay_rand` command can do a similar operation. I couldn't expect any random operation for the `delay` command.

> Regarding a single, accurate timer, can you give an example when it is important?

I had mentioned before, I **implemented the requested traffic duration** by the `delay` command. It is not related to the scale use-case. It is just a timing requirement at the app level. My user sees the traffic duration difference **as a value, not a rate**.
If there is another appropriate timer for it, I can change my app with it. If you know it, please let me know it.

>  Why if someone would ask for 100 sec delay with 20 usec resolution? I still don't get it. our main use-case is scale.
> now even if I be conviced that it is important, I think you should add it to the timer_w and not as you did.
>
> let's take a 100sec timer. in your case it would have 5000 ticks while with it could be implemented using ~3 ticks with the same accuracy (level2->level1->level0) so it should be part of timer_w code and not instruction

I don't need the 20 usec resolution for a single timer. I've taken only the simple code change to add accuracy to the `delay` command.
I didn't care about the performance when the accuracy is chosen. If you agreed to add accuracy to the `delay` command, I could improve my implementation at timer_w.

hhaim:
@jsmoon 
1) ```delay_rand``` can't do the same thing! read again my suggestion, it should be part of the timer_w level and not part of the app level. 
2)  your ```simple code change``` is not efficient by a few order of magnitude (x5000 slower in the example I gave). If it is really required  (I don't think anyone requires this accuracy ) add ```(level2->level1->level0) ```  idea to the timer_h level which will limit the number of invocation to the number of levels of the timers and not by time/max_duration_level_0 which could be very high.

I know that in your example that you have one timer and one flow, it seems simple to solve it like that but we try to think about the common use-case and not just one use-case

Both suggestions are orthogonal . both should be implemented in timer_w level, the first one is very simple, I would start from it

jsmoon:
@hhaim 
1. I read it again. I think your suggestion accumulates the error from each delay request and it added to the following delay request. Do I understand correctly? And, if so, I agree to do it.

request seq. | request ticks | real ticks | leftover ticks
---- | ---- | ---- | ----
1 | 5555 | 86*64 = 5504 | (5555 + 0) - 5504 = 51
2 | 5555 | 87*64 = 5568 | (5555 + 51) - 5568 = 38
3 | 5555 | 87*64 = 5568 | (5555 + 38) - 5568 = 25
4 | 5555 | 87*64 = 5568 | (5555 + 25) - 5568 = 12
5 | 5555 | 86*64 = 5504 | (5555 + 12) - 5504 = 63

2. I agree that my code was not efficient for a long delay. Actually, I didn't like to touch timer_w in this PR.

And I have a question for you. What is your opinion on my `delay` usage that implements the traffic duration?

jsmoon:
@hhaim, this change is under my understanding as I mentioned in my previous comment.
If I am wrong, please fix the point in the code.

hhaim:
@jsmoon, almost. There is no need for memory and I suggest to do it all the timer (no need to change the API and add accurate). 
in your example if you ask: 5555 it is 55/64 probability for choosing tick 87. 
so you ran a random number (0-32K)*55/64 if it is bigger than 16K you add 1 and instead of 86 you put 86+1 
The amortize of this random (+0 or +1) over number of start would be exactly 5555 (which is 5504 + 55/64)

hhaim:
@jsmoon let's start with random(), it should not use the system call which used a global variable across all cores !! this has a performance issue, see what we did in STL FE implementation (we kept the seed per core). 

Next the change should located in **one** place, should be something like that

```
  int add=0
   rand_num = residue (you example it was 55) *local_rand_per_core()/residure_max (e.g. in your example it was 64)
  if rand_num >32K {
    add=1
  }

 new_ticks +=add (this for level 1-3)
```

jsmoon:
@hhaim 
> it should not use the system call which used a global variable across all cores !!

I located the random number generator at `CTcpPerThreadCtx::m_timer_w`. As far as I know, the `CTcpPerThreadCtx` is instantiated per core. What point in my implementation is the global variable?

>  this has a performance issue, see what we did in STL FE implementation (we kept the seed per core)

I have found the random generator in the IpPool random selector and tried to implement it in a similar way.
I tried to find what you said, in STL FE implementation, I'm not sure what you have mentioned. Please let me know it as a codebase.
P.S.
I found another random number generator in `CTcpPerThreadCtx::m_rand` used by `delay_rand` command. What about changing it with this?

> new_ticks +=add (this for level 1-3)

If this is for level 1-3, the escalation for long-timer, `level3->level2->level1`, won't work properly. I need a long-timer error value that is less than a second. And this approach may make a second level difference.

hhaim:
This is the code of the fast random, seed should be part of the thread  
```
//https://software.intel.com/en-us/articles/fast-random-number-generator-on-the-intel-pentiumr-4-processor/

//Used to seed the generator.
inline void fast_srand(uint32_t &g_seed, int seed ){
  g_seed = seed;
}


//fastrand routine returns one integer, similar output value range as C lib.

inline int fastrand(uint32_t &g_seed)
{
  g_seed = (214013*g_seed+2531011);
  return (g_seed>>16)&0x7FFF;
}         
```

My comment was on this fix
>> @jsmoon, almost. There is no need for memory and I suggest to do it all the time (no need to change the API and add accurate).
in your example if you ask: 5555 it is 55/64 probability for choosing tick 87.
so you ran a random number (0-32K)*55/64 if it is bigger than 16K you add 1 and instead of 86 you put 86+1
The amortize of this random (+0 or +1) over number of start would be exactly 5555 (which is 5504 + 55/64)

Let's do one fix at a time

jsmoon:
@hhaim I've changed random generator code using `fastrand()`.
And I removed the level up operation as you suggested.
But I remained the codes that are aware of the timer spread effect.

ybrustin:
recheck

Issue #604, opened 2020-12-17
ASTF server send first issue


jsmoon:
Hi, this issue can be minor.
When I send first at the server-side, the first byte of the request would not be delivered.
```
           tcps_sndbyte |               128 |               249 | data bytes sent by application 
        tcps_sndbyte_ok |               128 |               248 | data bytes sent by tcp  
```
![image](https://user-images.githubusercontent.com/4134762/102458438-9933bf00-4087-11eb-9ea3-4d8f50375507.png)
![image](https://user-images.githubusercontent.com/4134762/102458669-eca60d00-4087-11eb-9a58-40b9cca76268.png)
As you can see in the picture, the first byte 'G' of "GET ..." was not included in the first data packet.

According to my investigation, during connection establishment (SYN -> SYN,ACK -> ACK), each ACK increases the acknowledge sequence number 1. If there is some data to send in the buffer where the last ACK received, the server treats the first byte already received by the client. (at src/44bsd/tcp_input.cpp)
```
1261         if (acked > so->so_snd.sb_cc) {
1262             INC_STAT_CNT(pctx, tg_id, tcps_rcvackbyte,so->so_snd.sb_cc);
1263             INC_STAT_CNT(pctx, tg_id, tcps_rcvackbyte_of,(acked-so->so_snd.sb_cc));
1264             tp->snd_wnd -= so->so_snd.sb_cc;
1265             so->so_snd.sbdrop_all(so);
1266             ourfinisacked = 1;
1267         } else {
1268             INC_STAT_CNT(pctx, tg_id, tcps_rcvackbyte,acked);
1269             so->so_snd.sbdrop(so,acked);  // <-- acked 1 byte here if there is some data (so->so_snd.sb_cc > 0)
1270             tp->snd_wnd -= acked;
1271             ourfinisacked = 0;
1272         }
```

I think this could happen because the `send()` from an application program is performed before the connection established. (at src/44bsd/flow_table.cpp)
```
 910     if (unlikely(temp->has_payload_params())) {
 911         lptflow->start_identifying_template(pctx, payload_info);
 912     } else {
 913         app->start(true); /* start the application */
 914     }       
 915 
 916     /* start listen */
 917     tcp_listen(lptflow->m_pctx,&lptflow->m_tcp);
```

@hhaim, do you know why the `app->start()` is called before `tcp_listen()`?

hhaim:
@jsmoon can you provide the profile?  
[...]

PR #609, opened 2021-01-08
Avoid ASTF server send first issue #604

jsmoon:
Hi, this PR is a change for issue #604.
Now, the first send doesn't lose the first byte by injecting the accept operation implicitly.


hhaim:
rebased

hhaim:
reabsed

PR #610, opened 2021-01-13
API to update/Add tunnel information for clients

vineetbvp:


hhaim:
@vineetbvp you should create a new PR based on the new code in the master branch. 
Your branch is has an old code and it has conflicts with master branch.
I couldn't review it, but suggest not to make the client context (RPC and messaging) a GTPU specific, try to make RPC-JSON and context generic as possible to be possible to add more tunnels for example 
So set_tunnel_context  for example, should be 
```{ 'gtpu' : {...} } ```

etc

vineetbvp:
@hhaim 
I have tried to make it generic. Please see the whole change set, and then you would understand what I am trying to say.
I have closed the current PR. I will post the new one soon.

vineetbvp:
@hhaim 
I understood, what you are saying. Python side code (RPC) I will make it generic. That part is missing.

hhaim:
@vineetbvp looks better. can you create a clean PR base on latest code base?

PR #617, opened 2021-01-25
Extend support for mpls tunneling

jhexan:
When mpls headers are encountered we now rely on ip_header_offset to point to the IP layer. If ip_header_offset is not configured we increment m_non_ip and return as before.


jhexan:
@hhaim any updates on this? Not sure what the failure above is, all the unit tests passed on my end

hhaim:
@jhexan there was an issue with the branch. it was fixed. it will be merge soon

hhaim:
rebased

PR #618, opened 2021-02-01
ASTF prevents new server flow after stopped

jsmoon:
Hi, I had met a crash case due to a flow with a removed profile.
I found the reason and this PR would fix it.

@hhaim, please check my change and give your feedback.

### DIFF
--- a/src/44bsd/flow_table.cpp
+++ b/src/44bsd/flow_table.cpp
@@ -489,7 +489,7 @@ CUdpFlow * CFlowTable::alloc_flow_udp(CPerProfileCtx * pctx,
     flow->m_c_template_idx = template_id;
     flow->m_template.set_tuple(src,dst,src_port,dst_port,vlan,IPHeader::Protocol::UDP,tun_handle,is_ipv6);
     flow->init();
-    flow->m_pctx->m_flow_cnt++;
+    flow->m_pctx->append_flow(flow);
     return(flow);
 }
 
@@ -512,14 +512,14 @@ CTcpFlow * CFlowTable::alloc_flow(CPerProfileCtx * pctx,
     flow->m_c_template_idx = template_id;
     flow->m_template.set_tuple(src,dst,src_port,dst_port,vlan,IPHeader::Protocol::TCP,tun_handle,is_ipv6);
     flow->init();
-    flow->m_pctx->m_flow_cnt++;
+    flow->m_pctx->append_flow(flow);
     return(flow);
 }
 
 HOT_FUNC void CFlowTable::free_flow(CFlowBase * flow){
     assert(flow);
     CPerProfileCtx* pctx = flow->m_pctx;
-    flow->m_pctx->m_flow_cnt--;
+    flow->m_pctx->remove_flow(flow);
     flow->m_pctx->on_flow_close();
 
     if ( flow->is_udp() ){
--- a/src/44bsd/tcp_subr.cpp
+++ b/src/44bsd/tcp_subr.cpp
@@ -399,8 +399,8 @@ void CTcpFlow::start_identifying_template(CPerProfileCtx* pctx, CServerIpPayload
     pctx->copy_tunable_values(this->m_pctx);
 
     // to collect TCP statistics
-    this->m_pctx->m_flow_cnt--;
-    pctx->m_flow_cnt = 1;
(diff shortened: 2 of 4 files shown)


hhaim:
@jsmoon please rebase on latest master

jsmoon:
Hi, my original PR causes side effects when a profile runs on multiple cores.
I missed the graceful-stop by the duration should keep the server flow generation after stopped because other core clients can be generating new flows.
So, I changed this PR to clean up unexpected server flows.
Since the unexpected flows would rarely happen by abnormal packets, I think this change has no impact on the normal situation.
@hhaim, please check my changes again and give your feedback.

hhaim:
@jsmoon you can't iterate over all the flows, you could have 100Mflows. Add a tag to each flow 

This will trigger a WD 
```
int CCloseHash<KEY>::detach_if(void *userdata,
                                 close_hash_on_detach_cb_t cb, void *pdata, close_hash_predicate_cb_t p){
    int detach_cnt = 0;

    for(int i=0;i<m_size;i++){
        CCloseHashRec * lpRec= &m_tbl[i];
        TCGenDListIterator iter(lpRec->m_list);
        hashEntry_t *lp;
        int f_itr=0;
        for (;(lp=(hashEntry_t *)iter.node());){
            iter++;
            if (p(pdata, lp)){
                lp->detach();
                m_numEntries--;
                cb(userdata,(void *)lp);
                detach_cnt++;
            }
            f_itr++;
        }
    }

    return detach_cnt;
}
```

jsmoon:
@hhaim I have no idea how the tag on each flow can help to find flows of a specific profile faster.
Could you elaborate on the idea?

jsmoon:
@hhaim I added a flow list for surplus flows created after the profile stopped.
Since the number of surplus flows is small, I think the impact would be negligible.

hhaim:
@jsmoon it is not fundamentally better. Why you are adding the flow if the profile_ctx is not active from the first place?
Iterating over flows is not a good idea especially this code that liniary iterate to find one flow. to remove ```n``` flow  it would be ```n^2```

```
void remove_flow(CFlowBase* flow) {
        m_flow_cnt--;
        if (!is_active()) {
            auto it = find(m_flows.begin(), m_flows.end(), flow);
            if (it != m_flows.end()) {
                m_flows.erase(it);
            }
        }
    }
```

Could you describe your issue again? 
My idea with the tag is to add some tag information that you could by fast query know if this flow is associated with active profile. 
them let the flow age out and kill it if the tag related to inactive profile context. 
Doing a big batch of work is not recommended

jsmoon:
@hhaim let me explain this issue at first.
The crash issue occurred by the flow which was generated after a DP core report finished TX to CP.
It can happen only from the *extra abnormal packets*.
ASTF server program already closed the open flow and reported the profile finished but a cleanup request from CP not yet delivered. During this period, a new packet can make a new flow to the *inactive* profile.
If the packet is *normal* from the other DP core client, the cleanup request would hold until regarding client flow is closed. This is normal.
But if the packet is *abnormal* from an unknown source, CP requests to cleanup profile to DP core, and the flow is left after the profile's AstfDB instance removed. *This flow causes crashes.*

As I mentioned before, the number of surplus flows would be small. So I think the linear search would be enough.
If you worried about the linear search, I can change it to binary search using `std::set`.
Please let me know what you want.

jsmoon:
@hhaim, I changed the linear search to binary search using `std::set`.

> Doing a big batch of work is not recommended

The surplus flows are small and it's not a big batch.

> My idea with the tag is to add some tag information that you could by fast query know if this flow is associated with active profile.
them let the flow age out and kill it if the tag related to inactive profile context.

To add the tag to the flow, the profile should find the flow at first. I don't know how the tag gets a benefit in this situation.

hhaim:
@jsmoon one question, why to open a flow from the first place in this situation. You are opening the flow and add them to a list and then close them. in this state (profile_ctx is not active) just don't open the flows.

jsmoon:
@hhaim my first PR(cb75ede) has been done as you said. But the regression test was failed. Please see [my comment](https://github.com/cisco-system-traffic-generator/trex-core/pull/618#issuecomment-772326670) about it.

jsmoon:
PR #625 would be a better solution. I'll give up this change.

PR #625, opened 2021-02-05
ASTF stops DP core servers after all clients stopped.

jsmoon:
Hi, this PR is another solution for PR #618.

This change separates the DP core report of TX finished.
CP will send the stop request finally after all DP core clients stopped.
This makes DP core servers confirm that there would be no surplus flows.
After the server stopped, open flows are not allowed anymore.

@hhaim, please look at this change also. If you like this one, I will give up PR #618.


hhaim:
@jsmoon looks good.

hhaim:
rebased

PR #627, opened 2021-02-07
Multi Pcap Support for the same Port

akhprasa:
I have made a change to the trex_astf_profile.py file to bring about the support for Multi-pcap for the same port. 
Please review and let me know.


hhaim:
@akhprasa not sure I understand what you did? what do you want to achive?

akhprasa:
@hhaim We needed to support multiple pcap's input for the same port. I have made the changes for that.

hhaim:
@akhprasa I don't see how this even close to solve the problem.

PR #641, opened 2021-02-26
Bump m_rw_mbuf_size to 255 if udp/tcp offsets are larger than 128

jhexan:
We ran into this issue by trying to configure vxlan IPv6 tunnels that add up to UDP/TCP offset >128


hhaim:
rebased

PR #643, opened 2021-03-02
temp changes

kphaye:


hhaim:
@kphaye I assume you merge the code in a wrong way, why the session_id was removed?
I don't see the python code and the rx counters.

Issue #639, opened 2021-02-26
TCP slow timer does not work as expected.


jsmoon:
Hi, I realized the TCP slow timer does not work as expected 500 msec.
It does work every 600 msec.

When I added the following debugging code in `tcp_timers()` routine,
```
printf("%p: %u - %u, %u, %.3f\n", tp, tp->t_rxtshift, tp->t_timer[TCPT_REXMT], tp->t_timer[TCPT_KEEP], now_sec());
```
the result showed like these.
```
0x7f51a0210740: 1 - 2, 5, 32.555
0x7f51a0210740: 2 - 2, 3, 33.756       // <-- 2 slow timer ticks (5 --> 3) elapsed 1201 msec, i.e. 600 msec / 1 tick.
0x7f51a0210740: 3 - 4, 1, 34.955
```

And I found the following comment in the ` CTcpTunableCtx::convert_slow_sec_to_ticks(uint16_t sec)`.
```
    // tcp_slow_fast_ratio + 1 for the slow_tcp bug (tick starts at 0, ratio at 1)
    float sec_to_ticks = 1000.0f / ((tcp_slow_fast_ratio + 1) * TCP_TIMER_TICK_FAST_MS);
```
This workaround generates different results to the default values.
For example,
TCPTV_KEEP_INIT is **10 ticks**, but the tunable `tcp.keepinit = 5` (default value) will use **8 ticks**.

I think this is a minor issue, but it is needed to be fixed.
@hhaim what do you think of this issue?

hhaim:
@jsmoon 

I think this function is not correct anymore as fast tick is tunable and not define. you can change the fast timer. 
```
    float sec_to_ticks = 1000.0f / ((tcp_slow_fast_ratio + 1) * TCP_TIMER_TICK_FAST_MS);
    return int(round(float(sec) * sec_to_ticks));

```

jsmoon:
@hhaim, my simple fix is like this:
```
diff --git a/src/44bsd/tcp_subr.cpp b/src/44bsd/tcp_subr.cpp
index 094e5df5..274abf26 100644
--- a/src/44bsd/tcp_subr.cpp
+++ b/src/44bsd/tcp_subr.cpp
@@ -884,8 +884,7 @@ void CTcpPerThreadCtx::init_sch_rampup(profile_id_t profile_id){
 }
 
 int CTcpTunableCtx::convert_slow_sec_to_ticks(uint16_t sec) {
[...]

PR #644, opened 2021-03-02
use TCP slow timer tick as defined

jsmoon:
Hi, this PR is for issue #639.

From this change, TCP slow timer tick would be as defined TCP_TIMER_TICK_SLOW_MS, 500ms.

@hhaim please give your feedback.


hhaim:
rebased

PR #651, opened 2021-03-10
Add an 'ignore-528-issue' switch to dpdk_setup_port.py to allow other

frmenguy:
ports, on a i40e driven board, to remain bound to a kernel driver.

Signed-off-by: Pierrick Louin <pierrick.louin@orange.com>


hhaim:
rebased

PR #658, opened 2021-03-11
multi tag changes

kphaye:


hhaim:
@kphaye please rebase on latest and fix the merge issues.

hhaim:
@kphaye there was a crash in our DMZ due to a bad commit, we are working to solve this. 
Please fix the merge to latest and try again

hhaim:
image
This change should not be part of you change, please rebase to latest

kphaye:
made the change manually . For some reason git rebase or git merge is not showing any delta for me

hhaim:
@kphaye  the diff looks ok now. you can continue with the counters and RPC now

kphaye:
As part of new commit we are achieving following 

1- As part of latency object we would be getting multi tag option set . 
2- if multi tag is set we would like to capture per vlan tag per port stats 
3- So as part of handle packet parsing we are checking if vlan header is present and retrieve the vlan tag. 
4- Now create a static map with key as hw-id + vlan tag and maintain stats in that location so that RPC can fetch the stats accordingly .

hhaim:
@kphaye 
There are two main issues here but we can continue as if this is an example and handle it later 

1. multi-tag stats should be dynamic allocated only if needed. with your diff you allocating 4K*1K (1KLatency object) which could be 4GB allocated at startup no matter if you have use this feature or not. 
2. in our first design we wanted to support  100K clients per pgid so it would be much bigger 4GB*100K 
3. The counters object should be much smaller 
4. The parser should be enable only if this feature was enabled --> there is a way to move information using messaging to enable/disable things in rx-core


In any case please bare this in mind, I suggest to continue and add RPC command to retrieve the counters now (add a new RPC like the one of latency)

hhaim:
recheck

hhaim:
too old, close for now

PR #663, opened 2021-03-12
fix trex-console exception on template_group

jsmoon:
Hi, this minor change is to avoid an exception when `template_group` command used in `trex-console`.
```
trex>template_group 
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
....
..../automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py", line 1679, in template_group_line
    pid_input = opts.profiles
AttributeError: 'Namespace' object has no attribute 'profiles'
Shutting down RPC client

```

### DIFF
--- a/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
@@ -1673,7 +1673,8 @@ class ASTFClient(TRexClient):
 
         opts = parser.parse_args(shlex.split(line))
 
-        if not opts:
+        if not opts or not opts.command:
+            parser.print_help()
             return
 
         pid_input = opts.profiles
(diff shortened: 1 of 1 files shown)


jsmoon:
@hhaim the test failure seems to be caused by the test environment.
```
  File "/usr/lib64/python2.7/socket.py", line 476, in readline
    data = self._sock.recv(self._rbufsize)
socket.timeout: timed out
Build step 'Execute shell' marked build as failure
```
Could you look into this?

bdollma:
This happens only when the user tries to run **template_group** command without an argument, which is invalid but very good catch.

hhaim:
rebased

PR #673, opened 2021-04-01
update Dhcp Profile To Support MAC & DHCP Options config.

rzrabi:


hhaim:
recheck

hhaim:
recheck

hhaim:
recheck

hhaim:
recheck

hhaim:
recheck

hhaim:
recheck

scripts/emu/simple_dhcp.py
```diff
@@ -32,17 +32,37 @@ def create_profile(self, ns_size, clients_size):
                                       plugs   = {'arp': {},
                                                  'icmp': {},
                                                  'igmp': {}, 
-                                                 'dhcp': {}
-                                                },
+                                                 'dhcp': {
+                                                     'options' : {
```
hhaim, line 36: @rzrabi better to have 2 options
1. backward compatibility -- no options provided just ```DHCP: {}`` in case option wan't provided by user
2. in case user provide an option to build the options object.

hhaim:
suggest to simplify the code. argparse should add the option directly to an object (in this example o) and then add the option directly (if it is not empty)

```import os

dhcp={"dhcp":{}}
o={}

b  = None
c  = None
if b:
  o["b"] =b
if c:
  o["c"] =c

if len(o):
  dhcp["dhcp"]["option"]=o

```

hhaim:
rebased

PR #674, opened 2021-04-05
change the number of ARP requests for gw address resolution

jsmoon:
Hi, this PR is to change the number of ARP requests for GW address resolution during `t-rex-64` startup.
The previous implementation has the hard-coded number 10.
My user wants to start DUT and T-Rex machine simultaneously to save startup time.
Sometimes DUT cannot reply to the ARP request in time. If the GW address resolution is failed at startup, the user should additional ARP requests for the test cases.
So I added a new command-line option `--limit-arp-request <count>`. If the value is 0, it will go until the GW address is resolved.

@hhaim what do you think about this change? Please review it and give your feedback.

### DIFF
--- a/src/main_dpdk.cpp
+++ b/src/main_dpdk.cpp
@@ -196,6 +196,7 @@ enum {
        OPT_LRO_OFFLOAD_DISABLE,
        OPT_CLOSE,
        OPT_ARP_REF_PER,
+       OPT_LIMT_ARP_REQ,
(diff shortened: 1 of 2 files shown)


hhaim:
@jsmoon looks good. I would use count=0. The arp resolve at startup relates to the old STF mode. with new ASTF/STL interactive it can be asked via API and for IPv6 too.

hhaim:
rebased

Issue #680, opened 2021-04-15
ASTF traffic duration with nc option


jsmoon:
Hi,

I have been used the ASTF profile of TCP traffic. The profile started with duration and nc.
But, when the `profile duration` is very short, sometimes the TCP flow cannot be connected within the duration.
Since the `m_traffic_duration` in the statistics is settled after the connection established, sometimes there is no traffic duration.
My user complaint about it. The delayed connection establishment can be natural when the DUT traffic path is in an idle state.

I understand that the `profile duration` is for the generation of new flows.
But when nc option is used, I think the `duration` would be traffic-related because it will stop the traffic.
So, I think that the profile duration could be synced with the traffic duration when both options(duration and nc) are specified.

![image](https://user-images.githubusercontent.com/4134762/114808054-7e2d4d80-9de2-11eb-9594-0c7b675b3eaf.png)

@hhaim what do you think about this?

hhaim:
@jsmoon the all point of ``--nc`` mode is to stop the traffic just after the ```traffic_duration``` (stop new flow generation and stop **active flow** in the middle).  The reason for that is that you can't predict what would be a actual flow duration due to the drop rate, in some cases flow that has normally a duration of 1 sec without drop could take 100 sec to finish due to the drop. ``` --nc``` mode solves this by terminating just after the duration but the price is a statistic that is not accurate. 

I did not understand your suggestion but maybe the solution is to have a new mode that provides **two durations**. 

1. Duration for generating new flows 
2. Duration for waiting for active flows to terminate (in ``--nc`` mode this is zero while without it is unlimited)

jsmoon:
@hhaim my suggestion is to separate the action of `duration` to `duration for new flow generation` and `duration for the traffic of active flows` when `--nc` is given.
The `--nc` mode will use the `duration` from the first connection establishment, not from the first flow generation.

hhaim:
@jsmoon understood your suggestion now. I think it is not a good solution, what would happen if there is an issue with establishing the flows, how log would you wait? better to have two durations as I suggested earlier and just give enough time for at least one flow to be established

jsmoon:
> what would happen if there is an issue with establishing the flows, how log would you wait?

My code will extend the duration during there are active flows. If there is no flow, the profile will be stopped immediately. i.e. maximum delay will be the first establishment time. In my investigation, it will be about **5 seconds** when all the establishments are failed. @hhaim, isn't it acceptable to you?

[...]

PR #681, opened 2021-04-19
ASTF traffic duration with nc option (#680)

jsmoon:
Hi, this change is for the implementation of #680.

When `--nc` option is given, the new `m_traffic_duration` in `CGenNodeTXFIF` is used to check the `traffic duration` from `CPerProfileCtx`.
The connection established flow will be closed by `--nc` when the traffic duration is over the given `duration` value.

@hhaim, please review this change and give your feedback.

### DIFF
--- a/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
@@ -651,7 +651,7 @@ class ASTFClient(TRexClient):
                 self.logger.info(format_text("Cannot remove a profile: %s is not state IDLE and state LOADED.\n" % profile_id, "bold", "magenta"))
 
     @client_api('command', True)
-    def start(self, mult = 1, duration = -1, nc = False, block = True, latency_pps = 0, ipv6 = False, pid_input = DEFAULT_PROFILE_ID, client_mask = 0xffffffff):
+    def start(self, mult = 1, duration = -1, nc = False, block = True, latency_pps = 0, ipv6 = False, pid_input = DEFAULT_PROFILE_ID, client_mask = 0xffffffff, e_duration = 0, t_duration = 0):
         """
             Start the traffic on loaded profile. Procedure is async.
 
@@ -681,6 +681,16 @@ class ASTFClient(TRexClient):
                 pid_input: string
                     Input profile ID
 
+                e_duration: float
+                    Maximum time to wait for one flow to be established.
+                    Stop the flow generation when this time is over. Stop immediately with nc.
+                    Disabled by default. Enabled by positive values.
+
+                t_duration: float
+                    Maximum time to wait for all the flow to terminate gracefully after duration.
+                    Stop immediately (overrides nc pararmeter) when this time is over.
+                    Disabled by default. Enabled by non-zero values.
+
             :raises:
                 + :exc:`TRexError`
         """
@@ -694,6 +704,8 @@ class ASTFClient(TRexClient):
             'latency_pps': latency_pps,
             'ipv6': ipv6,
             'client_mask': client_mask,
+            'e_duration': e_duration,
+            't_duration': t_duration,
             }
 
         self.ctx.logger.pre_cmd('Starting traffic.')
@@ -1330,6 +1342,8 @@ class ASTFClient(TRexClient):
                                          parsing_opts.FILE_PATH,
                                          parsing_opts.MULTIPLIER_NUM,
                                          parsing_opts.DURATION,
+                                         parsing_opts.ESTABLISH_DURATION,
+                                         parsing_opts.TERMINATE_DURATION,
                                          parsing_opts.ARGPARSE_TUNABLES,
                                          parsing_opts.ASTF_NC,
                                          parsing_opts.ASTF_LATENCY,
@@ -1378,7 +1392,7 @@ class ASTFClient(TRexClient):
             elif opts.servers_only:
                 kw['client_mask'] = 0
 
-            self.start(opts.mult, opts.duration, opts.nc, False, opts.latency_pps, opts.ipv6, pid_input = profile_id, **kw)
+            self.start(opts.mult, opts.duration, opts.nc, False, opts.latency_pps, opts.ipv6, pid_input = profile_id, e_duration = opts.e_duration, t_duration = opts.t_duration, **kw)
 
         return True
 
--- a/scripts/automation/trex_control_plane/interactive/trex/utils/parsing_opts.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/utils/parsing_opts.py
@@ -696,6 +696,24 @@ class OPTIONS_DB_ARGS:
          'default': -1.0,
          'help': "Set duration time for job."})
 
+    ESTABLISH_DURATION = ArgumentPack(
+        ['--e_duration'],
+        {'action': "store",
+         'metavar': 'TIME',
+         'dest': 'e_duration',
+         'type': match_time_unit,
+         'default': 0.0,
+         'help': "Set time limit for the first flow establishment."})
+
+    TERMINATE_DURATION = ArgumentPack(
+        ['--t_duration'],
+        {'action': "store",
+         'metavar': 'TIME',
+         'dest': 't_duration',
+         'type': match_time_unit,
+         'default': 0.0,
+         'help': "Set time limit waiting for all the flow to terminate gracefully."})
+
     TIMEOUT = ArgumentPack(
         ['-t'],
         {'action': "store",
--- a/src/bp_sim.h
+++ b/src/bp_sim.h
@@ -667,9 +667,11 @@ struct CGenNodeTXFIF : public CGenNodeBase {
 public:
     CPerProfileCtx *    m_pctx;
     double              m_time_stop;
+    double              m_time_established;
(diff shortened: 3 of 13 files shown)


jsmoon:
@hhaim this update keeps the backward compatible. When `-d` and `--nc` are used, it would do the same behavior as before.
When new options `e_duration` and `t_duration` are used, they would operate independently.

| `-d` | `--nc` | `--e_duration` | `--t_duration` | description
| --- | --- | --- | --- | ---
| any | any | - | - | the same behavior as before
| - | - | 5 | - | if there is no establishment in 5 seconds, the profile will stop immediately.
| - | - | - | 10 | if traffic duration is over 10 seconds, the profile will stop immediately. -- if the first establishment time is 10 seconds, the profile stops after 10 + 10 seconds from the profile start. -- if there is no establishment forever, the profile won't stop forever. Only the client can stop the profile by request.
| - | - | 5 | 10 | if no establishment in 5 seconds, the profile will stop immediately. this profile will stop within maximum 5 + 10 seconds.
| 5 | - | - | 10 | after 5 seconds, the profile will stop if there is no active flow. -- if the flows have a long lifetime, the profile will stop after establishment time + 10 seconds.
| 5 | true | - | 10 | after 5 seconds, the profile will stop immediately. (`--t_duration` has no effect in this case)
| 20 | true | - | 10 | after establishment time + 10 seconds, the profile will stop immediately. -- if establishment time is over 10 seconds, the profile will stop immediately after 20 seconds. (maximum 20 seconds)
| 20 | true | 5 | 10 | after establishment time + 10 seconds, the profile will stop immediately when the establishment time is less than 5 seconds. (maximum 15 seconds, `-d` and `--nc` have no effect in this case)

Please review the code and give your feedback.

hhaim:
@jsmoon very nice table. I think there is a confusion

```
  e_duration <-> traffic_duration <-> t_duration 

e_duration - maximum time to waits for one flow to be established
traffic_duration - generate new flows
t_duration - maximum time to waits for all the flow to terminate gracefully 

```

t_duration always start **after** ```e_duration``` + ```traffic_duration```. 
This is the **maximum** duration to waits for all the flows to tear down. it waits for ```active-flows == 0```

In case ```e_duration``` is provided, this is the duration to wait for the first flow to be established, this is again the maximum duration, if the first flow was establish earlier, the traffic_duration will start at this point. 

so let take your last example:

20 | true | 5 | 10 | after establishment time + 10 seconds, the profile will stop immediately when the establishment time is less than 5 seconds. (maximum 15 seconds, -d and --nc have no effect in this case)
-- | -- | -- | -- | --


==> this is what should happen instead 

After establishment time + 20 seconds (traffic_duration which generate new flows), the profile will get into tear-down state which start a timer of 10 sec and waits for all the flows to teardown. if after 10 sec there are still ```active-flow>0 ``` the traffic will stop immediately and clean those flows. Maximum theoretical duration is 5+20+10 =35 sec

jsmoon:
@hhaim I agree. there is confusion. Especially, I used `t_duration` for the same meaning of `traffic duration`.
And I have been used the following terminologies as the following meaning.
- duration: the time to generate new flows from the profile begin.
- traffic duration: the time which is reported as `m_traffic_duration` in the statistics.
- first establishment time: the time that the first connection is established in the flow. it is the base time to calculate the `m_traffic_duration`,

Since the `first establishment time` is not predictable, `e_duration` should be treated as timeout value as you mentioned.

Actually, my users want to get `m_traffic_duration` for their model's traffic time. The `duration` is not appropriate to it. There is a way needed to specify the `traffic duration`.

> After establishment time + 20 seconds (traffic_duration which generate new flows), the profile will get into tear-down state which start a timer of 10 sec and waits for all the flows to teardown. if after 10 sec there are still active-flow>0 the traffic will stop immediately and clean those flows. Maximum theoretical duration is 5+20+10 =35 sec

1. the duration (as you said, traffic_duration) which generates new flows implies the `first establishment time`. So, you cannot add it to the maximum duration. i.e. it could be 30 sec.
2. if `t_duration` is used to wait until active flows are closed, I couldn't get the appropriate value for it. It's because the first establishment time is not predictable. I need a **fixed value** to guarantee the `traffic duration`. In my point of view, it's not useful that is waiting for active flows to close. (after flow generation has done)

jsmoon:
@hhaim what about this? based on my understanding of your suggestion, I added my solution for my requirement (red color).

image

hhaim:
@jsmoon  yes, I missed the statistics and the traffic generation duration (```duration```)

>> Since the first establishment time is not predictable, e_duration should be treated as timeout value as you mentioned.

agreed. so in my example maximum theoretical duration is ```20+10 =30 sec```

>> the duration (as you said, traffic_duration) which generates new flows implies the first establishment time. So, you cannot add it to the maximum duration. i.e. it could be 30 sec.

agreed. 

>> if t_duration is used to wait until active flows are closed, I couldn't get the appropriate value for it. It's because the first establishment time is not predictable. I need a fixed value to guarantee the traffic duration. In my point of view, it's not useful that is waiting for active flows to close. (after flow generation has done)

```t_duration``` is the time-out after your stop generating new flow. This is the reason I suggest to give this flexibility to the user. Today ``--nc``` set this value to zero and ```normal``` to infinite time. The user should control the timeout duration. This has very high variation depending on the setup and the scenario.

hhaim:
image
@jsmoon 
I don't know why I can't see this in github, so I copy it from the email. 

Looks good. I think ```m_traffic_duration``` in case of ```--nc``` could be lower than ```-d``` as it stops immediately after the ```duration``` stoped and does not add the ```first_flow_establish_time``` 

```t_duration``` is a flexibility to limit the duration you waits for all the flow to terminate

jsmoon:
@hhaim I'm sorry for that. I removed it because I have realized my requirement can be accomplished without `-d` and `--nc` options.

image

I think I can get my solution by the combination of `--e_duration` and `--t_duration`.
With `-d` and `--nc`, it is changed like this.

image

I updated the table also based on above concept.

| `-d` | `--nc` | `--e_duration` | `--t_duration` | description
| --- | --- | --- | --- | ---
| any | any | - | - | the same behavior as before
| - | - | 5 | - | if there is no establishment in 5 seconds, the profile will stop immediately.
| - | - | - | 10 | if traffic duration is over 10 seconds, the profile will stop immediately. -- if the first establishment time is 10 seconds, the profile stops after 10 + 10 seconds from the profile start. -- if there is no establishment forever, the profile won't stop forever. Only the client can stop the profile by request.
| - | - | 5 | 10 | if no establishment in 5 seconds, the profile will stop immediately. otherwise, this profile will stop within maximum 5 + 10 seconds.
| 20 | - | - | 10 | after 20 seconds, the profile will stop new flow generation. and the profile will stop immediately after additional 10 seconds. (maximum 20 + 10 seconds)
| 20 | true | - | 10 | after 20 seconds, the profile will stop immediately. (`--t_duration` has no effect in this case)
| 20 | - | 5 | 10 | if no establishment in 5 seconds, the profile will stop immediately. -- after 20 seconds, the profile will stop new flow generation. and the profile will stop immediately after additional 10 seconds. (maximum 20 + 10 seconds)

jsmoon:
@hhaim according to your comment, I think you may prefer the following change.

`-d` | `--nc` | `--e_duration` | `--t_duration | description
-- | -- | -- | -- | --
20 | - | - | 10 | after 20 seconds, the profile will stop new flow generation. and the profile will stop immediately after additional 10 seconds. (maximum 20 + 10 seconds)
20 | true | - | 10 | after 20 seconds, the profile will stop immediately. (--t_duration has no effect in this case)
20 | - | 5 | 10 | if no establishment in 5 seconds, the profile will stop immediately. -- after 20 seconds, the profile will stop new flow generation. and the profile will stop immediately after additional 10 seconds. (maximum 20 + 10 seconds)

==>

`-d` | `--nc` | `--e_duration` | `--t_duration | description
-- | -- | -- | -- | --
20 | **any** | - | 10 | after 20 seconds, the profile will stop new flow generation. and the profile will stop immediately after additional 10 seconds. (maximum 20 + 10 seconds)
20 | **any** | 5 | 10 | if no establishment in 5 seconds, the profile will stop immediately. -- after 20 seconds, the profile will stop new flow generation. and the profile will stop immediately after additional 10 seconds. (maximum 20 + 10 seconds)

I'll prepare the PR update based on your response.

hhaim:
@jsmoon the chart is clear, but not the table. 

* ```--t_duration ``` duration is explicit and should take effect in all cases.  Without providing it the ```--nc``` should set the ```-t_duration``` to zero and  normal mode to infinity (backward compatible) 

* What is the meaning in the table to a row without a number for  ```-d``` ? there is always a duration 

let's take this example 
 >>20 | - | 5 | 10 | if no establishment in 5 seconds, the profile will stop immediately. -- after 20 seconds, the profile will stop new flow generation. and the profile will stop immediately after additional 10 seconds. (maximum 20 + 10 seconds)
-- | -- | -- | -- | --

it will stop after 10 sec if **not** all the flows will terminate but it could be a lower value if all flows terminated

jsmoon:
> What is the meaning in the table to a row without a number for -d ? there is always a duration

@hhaim it means `-d` is not specified. i.e. it is treated as `infinite`.

hhaim:
@jsmoon agreed.

hhaim:
rebased

PR #682, opened 2021-04-19
coverity mem leaks fix

darshanoag:
Fixing the resource leaks reported by static analysis tool

### DIFF
--- a/src/hdrh/hdr_histogram_log.c
+++ b/src/hdrh/hdr_histogram_log.c
@@ -1179,6 +1179,7 @@ int hdr_log_encode(struct hdr_histogram* histogram, char** encoded_histogram)
 
 cleanup:
     free(compressed_histogram);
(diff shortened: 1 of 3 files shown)


src/hdrh/hdr_histogram_log.c
```diff
@@ -1179,6 +1179,7 @@ int hdr_log_encode(struct hdr_histogram* histogram, char** encoded_histogram)
 
 cleanup:
     free(compressed_histogram);
+    free(encoded_histogram);
```
jsmoon, line 1182: encoded_histogram --> encoded_histogram_tmp
and this line should be moved to line 1175.

hhaim: @jsmoon is right

hhaim:
rebased

Issue #685, opened 2021-04-21
Name correction is needed in response json of RPC "get_global_stats"


leomaatEric:
There are two obvious naming mistakes in the counters returned from "get_global_stats" RPC invocation. as below shows, counter names of "m_total_nat_no_fid" and "m_total_nat_open" hold extra whitespace at the end of name string, and actually name "m_total_nat_open" recently changed in newer release from "m_total_nat_open "(one extra whitespace) to "m_total_nat_open   "(three extra whitespace), this makes my client app fails to parse the result of this RPC invocation.
```
{
	"id": 2051759702,
	"jsonrpc": "2.0",
	"result": {
		.......
		"m_total_clients": 0,
		"m_total_nat_active ": 0,
		"m_total_nat_learn_error": 0,
		"m_total_nat_no_fid ": 0,
		"m_total_nat_open   ": 0,
		"m_total_nat_syn_wait": 0,
		"m_total_nat_time_out": 0,
		"m_total_nat_time_out_wait_ack": 0,
		.......
	}
}
```

hhaim:
@leomaatEric I see 2 year old commit SHA 39a62671 that created this and it was not changed since. 
Is it a new issue?

leomaatEric:
@hhaim  sorry for misleading, it's not changed in the current or recent trex-core implementation.

My client used to receive async event to get the global stats, which has this format:
```
{
	"name": "trex-global",
	"type": 0,
	"data": {
		"ts": {
			"value": 9552318146756110,
			"freq": 2999995883
		},
		"m_cpu_util": 0.000625,
		"m_total_nat_no_fid": 0,
		"m_total_nat_open": 0,
                ......
	}
}
```
you can see these two counters( "m_total_nat_no_fid" and "m_total_nat_open") has no whitespace in the counter names.

Recently I changed my client implementation to call RPC "get_global_stats" to replace the async event, all counters are same except these two who have extra whitespaces in the counter names
```
{
	"id": 2051759702,
	"jsonrpc": "2.0",
	"result": {
		.......
		"m_total_nat_no_fid ": 0,
		"m_total_nat_open   ": 0,
		.......
	}
}
```
So I'm thinking that the counter names should be consistent no matter fetching it via async event or RPC, of cause I can change my client to adapt the name inconsistence, but I guess it could be a problem in trex-core code.

hhaim:
[...]

PR #687, opened 2021-04-23
Name correction for json response of RPC get_global_stats

leomaatEric:
Hi,

This is fix for #685, Only one code source is impacted in this commit.

@hhaim, please review my changes and give your comments.


hhaim:
rebased

Issue #683, opened 2021-04-19
[New feature] LAG/LACP support in TRex


egwakim:
Hi,
We need to use LAG/LACP with TRex.
Fortunately, LAG/LACP is supported in DPDK as bonding PMD.
See https://doc.dpdk.org/guides/prog_guide/link_bonding_poll_mode_drv_lib.html

Bonding device can be supported like memif case.
https://trex-tgn.cisco.com/trex/doc/trex_manual.html#_memif_interfaces

DPDK configuration change and additional changes for parsing .yaml file would be needed.

Will it be feasible in TRex?

Best Regards
Gwangmoon

hhaim:
@egwakim I don't see an issue to add it. We have a full scalable software solution now for both ASTF and STL for redirecting flows, we did that for AWS and Azure. 
vdev is supported in ```trex_cfg`` file too see Azure example 
https://github.com/cisco-system-traffic-generator/trex-core/wiki/Azure-with-netvsc-DPDK-driver-(v2.89)

I would look into a referance commit that added the netvsc dpdk driver example and add the same for bonding interface.

egwakim:
Thanks for comment,
Then, we'll try to contribute it.

PR #691, opened 2021-05-04
Add bonding dpdk driver

jsmoon:
Hi, this change is for issue #683.

- added `net_bonding` driver
- to get proper device information inherited from the slave devices, the initial configuration is called before the `bonding` device start.
- In mode 4 (802.3ad), it is necessary to do TX burst and RX burst at least every 100ms. added device flush to `wait_x_sec` and `idle_state_loop`. 
- DEV_TX_OFFLOAD_MULTI_SEGS is enabled by TREX_PATCH
 
To use the bonding device, add --vdev=net_bondingX... to the interfaces in trex_cfg.yaml.
```
  interfaces: [ '--vdev=net_bonding0,mode=4,slave=d8:00.0,slave=d8:00.1', 'dummy' ]
```

### DIFF
--- a/doc/trex_book.asciidoc
+++ b/doc/trex_book.asciidoc
@@ -813,6 +813,25 @@ Consider the following if using dummy ports for a single interface TRex generato
 | | ASTF | functional | full TCP stack
 |=================
 
+
+=== bonding interfaces
+
+It provides similar capabilities found in Linux bonding driver. Link Aggregation 802.3AD is supported in mode 4. See link:https://doc.dpdk.org/guides/prog_guide/link_bonding_poll_mode_drv_lib.html[here] for the details.
+
+Configuration example:
+
+[source,bash]
+----
+    ...
+    interfaces: [ '--vdev=net_bonding0,mode=0,slave=d8:00.0,slave=d8:00.1', 'dummy' ]
+    ...
+----
+
+The bonded device inherits attributes from the first slave device. All slave devices should have same speed and duplex.
+
+NOTE: When mode 4 is used, more settle time could be needed. Use -w command line option to give enough time for it.
+
+
 === Configuring for running with router (or other L3 device) as DUT
 
 You can follow link:trex_config_guide.html[this presentation] for an example of how to configure the router as a DUT.
--- a/linux_dpdk/ws_main.py
+++ b/linux_dpdk/ws_main.py
@@ -1337,6 +1337,21 @@ dpdk_src_x86_64 = SrcGroup(dir='src/dpdk/',
                  'drivers/net/netvsc/hn_nvs.c',
                  'drivers/net/netvsc/hn_vf.c',
 
+                 #ip_frag
+                 'lib/librte_ip_frag/rte_ipv4_fragmentation.c',
+                 'lib/librte_ip_frag/rte_ipv6_fragmentation.c',
+                 'lib/librte_ip_frag/rte_ipv4_reassembly.c',
+                 'lib/librte_ip_frag/rte_ipv6_reassembly.c',
+                 'lib/librte_ip_frag/rte_ip_frag_common.c',
+                 'lib/librte_ip_frag/ip_frag_internal.c',
+
+                 #bonding
+                 'drivers/net/bonding/rte_eth_bond_api.c',
+                 'drivers/net/bonding/rte_eth_bond_pmd.c',
+                 'drivers/net/bonding/rte_eth_bond_flow.c',
+                 'drivers/net/bonding/rte_eth_bond_args.c',
+                 'drivers/net/bonding/rte_eth_bond_8023ad.c',
+                 'drivers/net/bonding/rte_eth_bond_alb.c',
 
                  ])
 
@@ -2025,6 +2040,7 @@ dpdk_includes_path =''' ../src/
                         ../src/dpdk/lib/librte_port/
                         ../src/dpdk/lib/librte_ring/
                         ../src/dpdk/lib/librte_timer/
+                        ../src/dpdk/lib/librte_ip_frag/
                         ../src/dpdk/
                         
                         ../src/dpdk/lib/librte_security/
--- a/src/dpdk/drivers/net/bonding/rte_eth_bond_pmd.c
+++ b/src/dpdk/drivers/net/bonding/rte_eth_bond_pmd.c
@@ -346,6 +346,9 @@ rx_burst_8023ad(void *queue, struct rte_mbuf **bufs, uint16_t nb_pkts,
 				   !rte_is_same_ether_addr(bond_mac,
 						       &hdr->d_addr)) ||
 				  (!allmulti &&
+#ifdef  TREX_PATCH
+				   !rte_is_broadcast_ether_addr(&hdr->d_addr) &&
+#endif
 				   rte_is_multicast_ether_addr(&hdr->d_addr)))))) {
 
 				if (hdr->ether_type == ether_type_slow_be) {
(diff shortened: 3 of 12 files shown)


jsmoon:
@hhaim I'm not sure why this is happening.
```
+ python /scratch/kill_trex.py
  File "/scratch/kill_trex.py", line 8, in <module>
    if not c.is_trex_daemon_running():
  File "/scratch/trex/v2.81/automation/trex_control_plane/stf/trex_stf_lib/trex_client.py", line 1073, in is_trex_daemon_running
    return self.master_daemon.is_trex_daemon_running()
  File "/scratch/trex/v2.81/external_libs/jsonrpclib-pelix-0.4.1/jsonrpclib/jsonrpc.py", line 764, in __call__
    return self.__send(self.__name, kwargs)
  File "/scratch/trex/v2.81/external_libs/jsonrpclib-pelix-0.4.1/jsonrpclib/jsonrpc.py", line 643, in _request
    response = self._run_request(request)
  File "/scratch/trex/v2.81/external_libs/jsonrpclib-pelix-0.4.1/jsonrpclib/jsonrpc.py", line 676, in _run_request
    verbose=self.__verbose
  File "/usr/lib64/python2.7/xmlrpclib.py", line 1273, in request
    return self.single_request(host, handler, request_body, verbose)
  File "/scratch/trex/v2.81/external_libs/jsonrpclib-pelix-0.4.1/jsonrpclib/jsonrpc.py", line 377, in single_request
    response = connection.getresponse()
  File "/usr/lib64/python2.7/httplib.py", line 1113, in getresponse
    response.begin()
  File "/usr/lib64/python2.7/httplib.py", line 444, in begin
    version, status, reason = self._read_status()
  File "/usr/lib64/python2.7/httplib.py", line 408, in _read_status
    raise BadStatusLine(line)
httplib.BadStatusLine: ''
Build step 'Execute shell' marked build as failure
```
Could you look into this?

hhaim:
@jsmoon will check.
some comments:

* This function ```is_device_flush_needed``` should not be virtual, just add flag and one function that check the flags 

```
bool is_device_flush_needed(CPhyEthIF * _if){
        reutrn ((m_cap & tdCAP_FLSUH_NEEDED) >0)
    }
```

```
CTRexExtendedDriverBonding::CTRexExtendedDriverBonding(){
    m_cap = tdCAP_ONE_QUE | tdCAP_MULTI_QUE | TREX_DRV_CAP_MAC_ADDR_CHG |  tdCAP_FLSUH_NEEDED;
}
```

speed matter 

* In some critical path you added this ```if``` inside the loop. instead you could read the flag and duplicate the loop

```if ( likely(lp_port->m_len > 0) || lp_port->m_port->is_dev_flush_needed() ) {```

by doing this you are slowing down all the driver due to one slow driver. 


* A general question, does the flush related to the driver, maybe other driver can benefit from this feature, maybe this should be a mode of operation?

* Add sample configuration (as in this commit) in our manual and github wiki

**Comment by hhaim** [comment]
recheck

**Comment by hhaim** [comment]
recheck

**Comment by jsmoon** [comment]
@hhaim thank you for your comments.

> This function is_device_flush_needed should not be virtual, just add flag and one function that check the flags

The device flush is not always needed for the bonding device. It is required only when mode 4 is used. Since the mode may be different among multiple bonding devices, I thought the device flush should be handled per bonding device.
If I changed my code as you suggested, all bonding devices will perform the device flush always.

>  In some critical path you added this if inside the loop. instead you could read the flag and duplicate the loop

This device flush is per device operation. The loop is to iterate operation for the devices.
Regarding read the flag, I made the `is_dev_flush_needed()` method just read `m_dev_flush_needed`.
```
    if (get_ex_drv()->is_device_flush_needed(this)) {
        m_dev_flush_needed = true;
    }
```
```
    bool is_dev_flush_needed() { return m_dev_flush_needed; }
```

> by doing this you are slowing down all the driver due to one slow driver.

I can hardly agree with this opinion. According to my analysis, it added just 4 instructions including 1 forward jump instruction per loop. Furthermore, this function is COLD_FUNC.

> A general question, does the flush related to the driver, maybe other driver can benefit from this feature, maybe this should be a mode of operation?

I think so.

> Add sample configuration (as in this commit) in our manual and github wiki

Ok. I'll prepare for them. Do you have your preferred locations?

**Comment by jsmoon** [comment]
In the latest push, I added a `bonding interfaces` section to the manual. And I added a TREX_PATCH to allow broadcast packets without promiscous mode.

@hhaim, I left my thought about your comment. Do you still think I should use tdCAP_FLUSH_NEEDED flag?

**Comment by hhaim** [comment]
@jsmoon 

looks good. In regards to this:

```
bool CTRexExtendedDriverBonding::is_device_flush_needed(CPhyEthIF * _if){
    /* In LACP mode, need to call TX/RX burst at least every 100ms. */
    if (rte_eth_bond_mode_get(_if->get_repid()) == BONDING_MODE_8023AD) {
        return true;
    }
    return false;
}
``` 

This function could run in configuration phase and ```or```  the flag to the ```m_cap``` flags. The capability is not dynamic as the driver either requires flush or not right? I prefer to have flags instead of having an overwhelming number of virtual functions. 

You can fix this in a new PR, I will rebase this

**Comment by hhaim** [comment]
rebased

**Comment by jsmoon** [comment]
@hhaim I'm sorry for this issue. I've fixed it like this.
```
@@ -6090,10 +6090,15 @@ COLD_FUNC void check_pdev_vdev_dummy() {
             exit(1);
         } else {
             g_opts->m_is_vdev = true;
-            g_opts->m_is_devs_in_vdev = found_devs_in_vdev;
-            g_opts->m_pdevs_in_vdev = pdevs_in_vdev;
-            if ( global_platform_cfg_info.m_if_list_vdevs.empty() ) {
-                global_platform_cfg_info.m_if_list_vdevs = list_vdevs;
+            if ( found_devs_in_vdev ) {
+                g_opts->m_is_devs_in_vdev = found_devs_in_vdev;
+                g_opts->m_pdevs_in_vdev = pdevs_in_vdev;
+                /* when a vdev has sub devices, DPDK will report more devices than specified.
+                 * To perform reorder_dpdk_ports successfully, set to use m_if_list_vdevs.
+                 */
+                if ( global_platform_cfg_info.m_if_list_vdevs.empty() ) {
+                    global_platform_cfg_info.m_if_list_vdevs = list_vdevs;
+                }
             }
         }
     } else if ( !found_pdev ) {
```

**Comment by jsmoon** [comment]
> This function could run in configuration phase and or the flag to the m_cap flags. The capability is not dynamic as the driver either requires flush or not right? I prefer to have flags instead of having an overwhelming number of virtual functions.

Basically, the flush is dynamically required due to the bonding mode selection. Only the bonding mode 4 (BONDING_MODE_8023AD) requires it.

When there are two bonding interfaces that have mode 4 and another mode, only mode 4 bonding device needs to call device flush.
```
    interfaces: [ '--vdev=net_bonding0,mode=4,slave=d8:00.0,slave=d8:00.1', # need to call flush 
                  '--vdev=net_bonding0,mode=0,slave=d8:00.0,slave=d8:00.1' ] # don't need to call flush
```
If I locate the flag in the `m_cap`, I should check the all bonding devices to set the flag and it will be applied to all bonding devices.

@hhaim, I agree that the function could run in configuration phase. So I changed the flag location from `CPhyEthIF` to `DpdkTRexPortAttr`. Now this function will run when `g_trex.m_drv->create_port_attr` is called at `CPhyEthIF::Create`.

hhaim:
rebased

PR #695, opened 2021-05-07
ignore EEXIST error when binding identical cards

teknoraver:
Linux 5.12 changed the behaviour of the sysfs interface:
adding the same device id twice now generates an error[1].

  echo "1af4 1041" > /sys/bus/pci/drivers/vfio-pci/new_id
  echo "1af4 1041" > /sys/bus/pci/drivers/vfio-pci/new_id
  -bash: echo: write error: File exists

Fix this by ignoring EEXIST errors when writing in that specific handle.

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=3853f9123c185eb4018f5ccd3cdda5968efb5e10

Signed-off-by: Matteo Croce <mcroce@microsoft.com>

### DIFF
--- a/scripts/dpdk_nic_bind.py
+++ b/scripts/dpdk_nic_bind.py
@@ -675,10 +675,14 @@ def bind_one(dev_id, driver, force):
         try:
             f.write("%04x %04x" % (dev["Vendor"], dev["Device"]))
             f.close()
-        except:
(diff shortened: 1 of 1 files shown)


hhaim:
rebased

Issue #657, opened 2021-03-10
gtpu mode: Inner packet header checksum could be invalid.


syaakov:
The inner packet header checksum could be invalid when the port has capable of header checksum offload.

Issue #654, opened 2021-03-10
gtpu mode: Memory leak in gtp_tunnel.cpp


syaakov:
In  GTPU class in all the constructors, the pre_cooked_header is being allocated but the memory never gets free.

GTPU::GTPU(uint32_t teid,uint32_t src_ip, uint32_t dst_ip)
{

  Encapsulation *pre_cooked_header = new Encapsulation[1];
  store_ipv4_gtpu_info(teid, src_ip, dst_ip, pre_cooked_header);
}

~GTPU(){
 }

syaakov:
The destructor should look like this:
GTPU::~GTPU()
{
     delete [] (Encapsulation *)m_outer_hdr;
}

and we need to initialize the field m_outer_hdr with NULL

private:
    uint8_t *m_outer_hdr=NULL;

Issue #652, opened 2021-03-10
gtpu mode: disable tunnel adjust in port 1


syaakov:
Trex adjusts the packet in RX 1 instead of adjusting it only in port 0.
To fix it add the following lines in tunnel.cpp to the start of the method Tunnel::RxCallback.

    if (!CGlobalInfo::m_options.m_ip_cfg[port].is_gtp_enabled()) {
        return nb_pkts;
    }

This code prevents adjusting the packet in port 1.

Now the Tunnel::RxCallback should look like this:

uint16_t Tunnel::RxCallback(uint16_t port, uint16_t queue,
                struct rte_mbuf *pkts[], uint16_t nb_pkts, uint16_t max_pkts,
                void *user_param) {
    if (!CGlobalInfo::m_options.m_ip_cfg[port].is_gtp_enabled()) {
        return nb_pkts;
    }
    GTPU t;
    u_int8_t i = 0;

    if (0 == nb_pkts)
        return nb_pkts;

    rte_mbuf *m = NULL;

    /*Process the packets*/
    for (i=0; i < nb_pkts ; i++) {
        m = pkts[i];
        t.Adjust(m,queue);
    }

    return nb_pkts;
}

PR #696, opened 2021-05-07
fix GTPU inner header checksum issue

jsmoon:
Hi, this change is to fix following GTPU mode issues.

- #657 Inner packet header checksum could be invalid.
- #654 Memory leak in gtp_tunnel.cpp
- #652 disable tunnel adjust in port 1


jsmoon:
@hhaim could you look into this result of failure?

hhaim:
@jsmoon there is an ongoing task to replace the code to a new design. I believe those issues were fixed in the new code, but until it will be replaced accepting this

hhaim:
rebased

jsmoon:
@hhaim could I ask you when the ongoing task will be finished?

hhaim:
@jsmoon we almost there, the code was ported to the simulation and can work on trex+sim. @syaakov is working on this task.

PR #697, opened 2021-05-08
don't fail on a non NUMA system

teknoraver:
On a non NUMA kernel, some handlers in sysfs are not present.
If --ignore-numa was specified, use the generic handles instead of the
NUMA ones.

Signed-off-by: Matteo Croce <mcroce@microsoft.com>


hhaim:
rebased

PR #699, opened 2021-05-10
Add stack option for config file creation

Truoris:
Add stack option for config file creation in automated environment.


hhaim:
rebased

PR #700, opened 2021-05-11
terminate l7_map flow immediately to prevent crash when all reference profiles are stopped.

jsmoon:
Hi, this change is to prevent crash from invalid memory access.

When the reference profile template was removed, `update_template_flows` tried to update it by new reference profile.
If there was no reference profile left, it triggerred graceful flow terminateion.
But, this graceful flow termination could access already freed memory (m_termplate_rw) and may cause crash.
So, I changed it to the immediate flow termination because the graceful way is not needed in this case.


hhaim:
rebased

PR #701, opened 2021-05-12
Fix bug with dpdk_setup_ports script

ohilyard:
If the --cores-include flag was used, and the system has more lcores than
the current MAX_LCORE_NUM (63 on x86_64), then the script attempted to
remove some number of lcores from a list twice. This causes an error to be
thrown. Replacing "if" with "elif" prevents this behavior by ensuring that
the removal code is run at most once per lcore.

Signed-off-by: Owen Hilyard <ohilyard@iol.unh.edu>


hhaim:
rebased

Issue #693, opened 2021-05-06
trex-console read-only mode need to receive async events from the T-Rex server.


jsmoon:
When I start a profile in ASTF mode in trex-console, I could read async events like this.
```
trex>events -c
2021-05-06 04:57:49 - [server][info]     - Moved to profile _ state: 1
2021-05-06 04:57:49 - [server][info]     - Moved to state: Loaded
2021-05-06 04:57:49 - [server][info]     - Moved to profile _ state: 2
2021-05-06 04:57:49 - [server][info]     - Moved to state: Parsing
2021-05-06 04:57:49 - [server][info]     - Moved to profile _ state: 3
2021-05-06 04:57:49 - [server][info]     - Moved to state: Building
2021-05-06 04:57:49 - [server][info]     - Moved to profile _ state: 4
2021-05-06 04:57:49 - [server][info]     - Moved to state: Transmitting
2021-05-06 04:57:51 - [server][info]     - Moved to profile _ state: 5
2021-05-06 04:57:51 - [server][info]     - Moved to state: Cleanup
2021-05-06 04:57:51 - [server][info]     - Moved to profile _ state: 1
2021-05-06 04:57:51 - [server][info]     - Moved to state: Loaded
```

But another trex-console with read-only mode could not receive any async events.
```
trex(read-only)>events -c
trex(read-only)>
```

Since I've been used the read-only mode to monitor the T-Rex server, I would like to get the async events again.

jsmoon:
@hhaim I've got an idea for the solution.

Current implementation is to keep the session only for the port owners. The session is created by "acquire" and removed when the ownership is lost.
I think the session could be created by "publish_now" request in the `barrier` method. It could be removed by aging out when "get_async_events" is not requested in the specific time(maybe it's a few seconds).

What do you think about my idea?

jsmoon:
@hhaim, I agree that `publish_now` is not appropriate for the backward compatibility.

I think about another way which is keeping the common session context for the read-only console clients.
In this idea, `get_async_events` has common read-only `session_id` (for example, 0) and optional `seq` wants to get from.
```
self.rpc.transmit("get_async_events", params = { 'session_id': 0, 'seq': self.seq })
```

What do you think about my new suggetion?

PR #702, opened 2021-05-13
Add async events reader mode support

jsmoon:
Hi, this change is for issue #693.

Since there is no acquire in reader mode trex-console, I just added a session implicitly when the first `get_async_events` is requested.

@hhaim please check my changes and give your feedback


hhaim:
@jsmoon looks good, actually this is a good idea, we are missing only a few events and it is not that bad. 
you could have added an ```acquire``` for read-only mode  but your suggestion is better.

hhaim:
rebased

PR #703, opened 2021-05-13
update inactive reference template info on the fly

jsmoon:
Hi, this change is to prevent SYN packet drop from unexpected inactive reference template.

The reference template will be deleted after the related profile is stopped.
But, when the profile is stopping, the template is still remained as a reference template, and the new open flow is rejected even though another profile is active.
This change will replace the inactive reference tempate to active one on the fly.


hhaim:
rebased

PR #704, opened 2021-05-18
remove endless loop possibility in detach pool

jsmoon:
Hi, this change is to prevent a watchdog from the endless loop in `detach_pool_and_client_node`.

When `m_ref_pool_ptr.size()` is more than `sizeof(uint8_t)`, the condition is always true and no exit condition met where the entry is located in out of the boundary.
This change will remove the possibility.


hhaim:
rebased

PR #705, opened 2021-05-18
load astf_schema.json once for multiple profiles

jsmoon:
Hi, this change is to load `astf_schema.json` file once for the performance of multiple profiles.

The additional profile creates a new CAstfDB instance. The instance allocates `m_validator` for it, but the `m_validator` is the same for all profiles. Furthermore, it initiates contents from `astf_schema.json` file each time.
It is performed in DP core and takes time. In my measurement, it showed up to several hundred milliseconds.
This change will remove this unnecessary file I/O and I expect better results during the additional profiles are added.


hhaim:
rebased

PR #709, opened 2021-05-25
Fix byte order in GTPU TEID

ivan4th:
Also, don't do extra multiple host-to-network conversions for GTPU
tunnel fields.

`src_ip` / `dst_ip` are obtained using `inet_pton()` here:
https://github.com/cisco-system-traffic-generator/trex-core/blob/master/src/stx/astf/trex_astf_rpc_cmds.cpp#L855-L856
So no need to convert them to the network byte order.
For some reason, that conversion was being done twice, so the result was correct in the end, but I believe that it was unnecessary.

`TEID` was also being converted twice but as initially it was in the host byte order:
https://github.com/cisco-system-traffic-generator/trex-core/blob/master/src/stx/astf/trex_astf_rpc_cmds.cpp#L859
the resulting byte order was wrong. I have removed one of the conversions to make it right.


syaakov:
You are right, we will fix it

hhaim:
fixed in the new code that was't commited yet

Issue #508, opened 2020-07-14
Dpdk 20.02 with new X?710 firmware fail to implement low-latency tx queues


hhaim:
**Background**: i40e dpdk driver was changed to add low-latency tx queue (using Qos firmware commands). The LL tx queue is always the **last** queue.  With the new firmware it seems that the new Qos command does not work anymore. 
**New firmware**: for XL?10 (**7.0.x**) 
**Implication**: 
* STL latency queue will crash with watchdog 
* ASTF - packet in first or last queue will not be sent, ARP for example won't work 

We still want to LL feature to work on old firmware 6.0.1< so we add a workaround to add a dummy tx queue for this case. 
The Qos of the TRex low-latency tx queue will not work in this case.

hhaim:
TRex v2.82 has a fix. Old DPDK version (DPDK 19.05 up to v2.79) does not present the issue. 
Intel is looking into this firmware/dpdk issue. 
With the workaround (v2.82+) LL queue will not work so the latency packets resolution will be impacted

PR #712, opened 2021-05-25
Latency Measurement by HW timestamping with IEEE 1588 packets in Latency Stream

abhiramnarayana:
[TREX Latency Measurement using PTP packets in Latency Stream.docx](https://github.com/cisco-system-traffic-generator/trex-core/files/6542031/TREX.Latency.Measurement.using.PTP.packets.in.Latency.Stream.docx)

Using the 2-step mechanism of PTP protocol we send the accurate timestamp (t1)
of SYNC packet which is read from the NIC in a Follow Up packet rather than in
the same packet. Received timestamp (t2) is still taken on the first packet
(SYNC) itself. In this way with a combination of software(PTP protocol, DPDK
APIs) and HW support (leveraging IEEE 1588 support) the accuracy of latency
and jitter measurements can be improved.

Currently all changes are under a macro and shouldn't impact any existing functionality.
We are also looking currently on how to make it work by run-time checks rather than compile time macro.
But would want to keep it for a later commit and keep macro for the Initial commit.

Please refer to the attached document for more details on implementation and testing 

Signed-off-by: Abhiram R N <arn@redhat.com>

### DIFF
--- a/src/main_dpdk.cpp
+++ b/src/main_dpdk.cpp
@@ -26,6 +26,7 @@
 #include <string.h>
 #include <unistd.h>
 #include <zmq.h>
+#include <sys/time.h>
 #include <rte_config.h>
 #include <rte_common.h>
 #include <rte_log.h>
@@ -1386,10 +1387,30 @@ COLD_FUNC void CPhyEthIF::start(){
 
     m_stats.Clear();
     int i;
+#ifdef LATENCY_IEEE_1588_TIMESTAMPING
+    struct timespec tp;
+#endif
     for (i=0;i<10; i++ ) {
         ret = rte_eth_dev_start(m_repid);
         if (ret==0) {
+#ifndef LATENCY_IEEE_1588_TIMESTAMPING
             return;
+#else
+            ret = rte_eth_timesync_enable(m_repid);
+            if (ret == 0 ) {
+                printf("TIMESYNC ENABLED %u \n",m_repid);
+                clock_gettime(CLOCK_REALTIME, &tp);
+                ret = rte_eth_timesync_write_time(m_repid, &tp);
+                if (ret == 0) {
+                    printf("TIMESYNC Write time success \n");
+                } else {
+                    printf("TIMESYNC Write FAILED \n",m_repid);
+                }
+            } else {
+                printf("TIMESYNC ENABLE FAILED \n",m_repid);
+            }
+            break;
+#endif
         }
         delay(1000);
     }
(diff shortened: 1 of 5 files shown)


hhaim:
@abhiramnarayana I think there are two use-cases here

1) with bare-metal server, with XL710 we have a low-latency tx queue for latency and using that there is max 10-20 usec accuracy 

```start -f stl/udp_1pkt_src_ip_split_latency.py --port 0  -m 1mpps --force  -t fsize=4000,lfsize=4000```

image

The interesting things happen when the packets are long (4K) and you get to the maximum line rate, the Tx Qos is important and you can see that it works fine. 

Could you check the same command in your case with XL710? there should be a one time spike but the sliding window should have very low numbers (<20 usec)

2) trex works on a hypervisor with  direct PCI access to XL710, in this case HW timestamp can help but I can argue that if you want accurate latency use (1)

Another limitation:

HW timestamp has limitation and require a specific UDP packet and in our use-case we need to measure latency in many types of packets.

abhiramnarayana:
Hi @hhaim ,

Thankyou so much for your quick reply on this. 
Below are the results what I got. And seems promising for HW timestamping :) 
I am testing on XL710 NIC with 10Gbps speed. I have done the below tests in Loopback mode.

The first one is without my changes. Just with the current TREX as it is.

start -f stl/udp_1pkt_src_ip_split_latency.py --port 0 -m 250kpps --force -t fsize=4000,lfsize=4000

reply1

The second one is with my changes. i.e the IEEE 1588 HW timestamping.
But Note here I need to keep the lfsize at 106 as that is the 1EEE 1588v2 packet. But the UDP traffic I have made 4000 (fsize).

start -f stl/udp_1pkt_src_ip_split_latency.py --port 0 -m 250kpps --force -t fsize=4000,lfsize=106

reply3

Clearly here we can see the difference that the extra latency which is due to CPU processing and also transmission delays are all removed with the HW timestamping.

Regarding your comment on limitation I do agree latency packet has to be specific UDP packet (PTP packet) and length currently needs be 106. But it can be multiplexed with any other stream with different types/sizes like we have done above with size of 4000. 

I did run some more tests without loopback and having testpmd fwd packets from the other node. Even there I observed considerable improvement in the latency numbers.
Hope this helps. Let me know your comments.

Thank & Regards,
Abhiram R N

hhaim:
Hi @abhiramnarayana thanks for the info,
You are probably hitting https://github.com/cisco-system-traffic-generator/trex-core/issues/508 due to new firmware that explains the very high difference. The main issue is in the tx side as rx side is free (due to the filter), when you use very large packets close to line-rate the qos configuration will be effective (in my setup it is about ~15usec maximum latency).

The limitation in your case is that usually we are using different type of packets for **latency** stream (not just for the non-latency) and with this model you are limited to a specific packet (size and type) and limited to a specific rate and specific NIC type ...

abhiramnarayana:
Hi @hhaim ,

Thanks for pointing out the firmware thing. To quickly check it I moved to v2.79 and I do see numbers similar to yours now . 

Latency numbers from the tests are as below... 
latency with v2.79  --> **14 - 18** usec
latency with v2.79 + my patch --> **1-2** usec
latency with v2.89 + my patch --> **1-3** usec

What I observe is there are cases where it goes below 1 usec when I test with my patch. Those we need to handle yet and that might make it even more accurate. 

While mostly I have to agree on the limitations you put forth as it is based on the protocol, few points which I wanted to mention.

**Type** - Yes it has to be a PTP packet (latency stream)
**Size** -  Yes , 106 currently.  [I did experiment by increasing to 110 and 122 by hardcoding the suffix bytes and it worked fine. But still there will be a limit on how much suffix it can carry]
**Rate** -  Not a limitation I would say. Though I kept 128 pps for initial testing. Now I have tried with 1000 pps and 5000 pps as well and it works fine.
**Specific NIC type** - When I tried to look for this found that around 8 types support IEEE 1588v2 . (TimeSync Feature) . Link for your reference https://doc.dpdk.org/guides/nics/overview.html

What I feel is at-least for those drivers which have support can make use of this. Also what I want to bring to your notice is with this we are not getting 'tied up with Hardware' and that the DPDK APIs which we are using is generic and also while we intend to move away from this macro based implementation this driver capability based implementation would help and based on that to do the latency test with hardware aided timestamping or not.


Just in case you are curious to see the full result screenshots.
v2.79 as is
reply4a

v2.79 + my patch changes
reply4b

Thanks & Regards,
Abhiram R N

hhaim:
@abhiramnarayana understood. It seems this features can be enabled on multiple hardware. 
I've accepted this, however, it is better to remove the ```#ifdef ``` and make it dynamic identified as a new feature in the stream.

Issue #708, opened 2021-05-24
Tracking ASTF profile state changes


jsmoon:
@hhaim, I'd like to get the ASTF profile state changes.
It is needed for a very short period of running due to an automatic profile stop from #679.
I think I can stack the state changes from the `get_async_events` messages.
But, while implementing this feature, I have found the following issues.

1. `get_async_events` and `sync`
After `get_async_events` was introduced, I think the `sync` is not required anymore after once performed at the `connect`.

2. `sync` and `epoch`
According to my investigation, by setting `self.epoch` as None, the async event messages won't update the profile state during `sync`. But, I need to stack all the event messages even though the `sync` is in progress.

3. `inc_epoch` in multi-threading
When I start a profile with blocking mode, `_transmit_async` triggers `inc_epoch` for the first profile. But, there is another thread that is starting another profile at that time, `inc_epoch` can be triggered also. I think this behavior is not thread-safe and unexpected.
In my thought, this action is not required. The T-Rex server is increasing the `epoch` internally. And, if I don't use blocking mode, there is no `inc_epoch` action.

@hhaim, what do you think about these issues?

hhaim:
@jsmoon 

In case you stack it, you can get a burst of changes that only the last is relevant, is it what you want?
1. Yes
2. didn't understand this
3. Our API is not multi-threaded safe.  (we internally protect it)

jsmoon:
@hhaim

Yes, I don't want to miss TX state while I'm waiting for it.

1. OK. But, I found an additional issue. When a profile is failed to load, the profile is STATE_IDLE in the T-Rex server. But the ASTFClient doesn't know it. Currently, only `sync` can get it.

2. Simply say, `get_async_events` messages are not updated during the `sync`  is in progress. If I remove the `sync` call, there is no issue anymore.

3. Understood. Anyway, I think the `inc_epoch` behavior is useless. What do you think about this?

hhaim:
@jsmoon so what is suggestion to solve it? in once section you say that you don't need the ```sync``` and in another you need it

jsmoon:
@hhaim, my suggestion for solving the issues is like the following:
1. I'll add publish the IDLE state and remove unnecessary `sync` calls.
2. If I removed the unnecessary `sync`, there is no other change for this issue.
3. I'd like to remove the unnecessary `inc_epoch` action.

hhaim:
[...]

PR #714, opened 2021-05-27
Add profile state change tracking in ASTFClient

jsmoon:
Hi, this PR includes all the changes discussed in #708.

From now on, when the client needs to wait for a specific state, it won't miss the state by using the `self.__get_profile_state_with_change` method.

### DIFF
--- a/scripts/automation/trex_control_plane/interactive/trex/astf/stats/traffic.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/astf/stats/traffic.py
@@ -64,6 +64,11 @@ class CAstfTrafficStats(object):
         if pid_input in self.tg_names_dict.keys():
             self.tg_names_dict.pop(pid_input)
 
+    def _remove_stats(self, pid_input = DEFAULT_PROFILE_ID):
+        if pid_input in self._ref.keys():
+            self._ref.pop(pid_input)
+        self._clear_tg_name(pid_input)
+
 
     def _epoch_changed(self, new_epoch, pid_input = DEFAULT_PROFILE_ID, is_sum = False):
         if is_sum:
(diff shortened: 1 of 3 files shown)


hhaim:
rebased

hhaim:
@jsmoon there is an issue with this patch. The test that load many profiles and change the state fails (randomly on each setup).
in the meantime, try to reconstruct it

hhaim:
@jsmoon 
image
The issue is not seen after reverting this PR

jsmoon:
@hhaim, I added a change to STATE_IDLE when the profile cleared.
But it was due to my misunderstanding. I'll remove it from my change.

Anyway, I think the following code needs to be changed.
```
                self.clear_profile(False, pid_input=ALL_PROFILE_ID)
                self.check_states(ok_states=[self.STATE_IDLE])
                self.clear_stats(ports, pid_input = ALL_PROFILE_ID) # <-- ALL_PROFILE_ID is meaningless.
```
There is no need to clear ALL_PROFILE_ID statistics after `clear_profile` is called.

jsmoon:
@hhaim the conflict was left from the previous push.
The conflict may be left as the last message if it is not resolved.

In push 3f5a218, I made my change from the latest version already and it has passed the test.
Anyway, I resolved the conflict and the message has gone.

jsmoon:
@hhaim, yesterday, I've checked the master branch but there is no revert commit in there.
But today, I can see the revert already had been done. It's weird in my point.
I'll prepare the full change again from the latest commit.

hhaim:
> @hhaim, yesterday, I've checked the master branch but there is no revert commit in there.
> But today, I can see the revert already had been done. It's weird in my point.
> I'll prepare the full change again from the latest commit.

We have an internal repo that is used by regression and it is pushed from time to time to github (a few days). This is the reason you haven't seen yersterday the revert that was done a few days back. 
The conflicts are due to the revert  commit.

hhaim:
rebased

Issue #716, opened 2021-05-27
ASTF set_send_blocking(False) does not work in the client side


hhaim:
```
        # client commands
        prog_c = ASTFProgram()        
        prog_c.connect()


        prog_c.send_chunk(message,205,0)
        prog_c.set_send_blocking(False)
        prog_c.send_chunk(message,205,0)
        prog_c.set_send_blocking(True)



        prog_s = ASTFProgram()
        prog_s.recv(len(my_message)*1000) # I do not need this since our test is a single-interface where TRex connects (prog_c) to another app which is not TRex
```

This program does not work as expected , the ```set_send_blocking(False)``` is not applied

hhaim:
@jsmoon this explains it

PR #717, opened 2021-06-01
Fix ASTF non-blocking send issue

jsmoon:
Hi, this change is to fix issue #716.

`send` will continue the next command immediately in non-blocking mode.

@hhaim please check this change.


hhaim:
rebased

Issue #720, opened 2021-06-07
ASTF multi-profile,  crash in case in counter update 


hhaim:
test:  ```test_astf_template_group_load_dynamic_profile```
crash: once in a few weeks

```
*** traceback follows ***

1             0x5701b2 ./_t-rex-64() [0x5701b2]
2       0x7f0645cc9330 /lib/x86_64-linux-gnu/libpthread.so.0(+0x10330) [0x7f0645cc9330]
3             0x62417b CGCountersUtl64::operator+=(CGCountersUtl64 const&) + 91
4             0x53ce57 CSTTCp::Update() + 311
5             0x4f61b0 CGlobalTRex::update_stats() + 602
6             0x4f90dd CGlobalTRex::sync_threads_stats() + 9
7             0x4f9e2b CGlobalTRex::global_stats_to_json(Json::Value&) + 17
8             0x6d5e60 TrexRpcCmdGetGlobalStats::_run(Json::Value const&, Json::Value&) + 64
9             0x6c3d7d TrexRpcCommand::run(Json::Value const&, Json::Value&) + 77
10            0x6be67c JsonRpcMethod::_execute(Json::Value&) + 44
11            0x6bbcdd TrexJsonRpcV2ParsedObject::execute(Json::Value&) + 125
12            0x6b9b16 TrexRpcServerReqRes::process_request_raw(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) + 486
13            0x6ba3d8 TrexRpcServerReqRes::process_zipped_request(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) + 456
14            0x6ba6fd TrexRpcServerReqRes::handle_request(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 509
15            0x6bade7 TrexRpcServerReqRes::_rpc_thread_cb_int() + 1575
16            0x6bb516 TrexRpcServerReqRes::_rpc_thread_cb() + 6
17      0x7f06455d427f so/x86_64/libstdc++.so.6(+0xba27f) [0x7f06455d427f]
18      0x7f0645cc1184 /lib/x86_64-linux-gnu/libpthread.so.0(+0x8184) [0x7f0645cc1184]
[...]

PR #721, opened 2021-06-08
Prevent counter access from deleted ASTF profile

jsmoon:
Hi, this change is to fix issue #720.

### DIFF
--- a/src/stx/astf/trex_astf.cpp
+++ b/src/stx/astf/trex_astf.cpp
@@ -876,6 +876,8 @@ void TrexAstfPerProfile::profile_change_state(state_e new_state) {
             break;
         case STATE_DELETE:
             m_stt_cp->m_update = false;
+            // to prevent unexpected access from reused default profile
+            m_stt_cp->clear_profile_ctx();
             m_active_cores = 1;
             break;
         case AMOUNT_OF_STATES:


hhaim:
@jsmoon  thanks for looking into this. 

The first thing  DELETE state does is to set ```update=false``` so there is no need to clear the context pointer. 
```
the function Update 

void CSTTCp::Update(){
    if (!m_update) return;  <<<<<
```

I think there is a place that change the ```m_update``` to true at some point and then access the countes that were free ..

in short, I  still don't understand how the issue happens 

Did you reconstruct it using a fast add/remove profiles?

jsmoon:
@hhaim, in my analysis, there is a case that can access the removed profile counter.
Since the default profile reuses the `m_stt_cp`, if the DP profile ctx left on `CSTTCp::m_sts[].m_profile_ctx` after the DP profile ctx was deleted.
In this status, a new profile is loaded to the default profile, the `m_update` is changed to true and it triggers to access the left removed DP profile ctx during the LOADED state.
My change will prevent the removed DP profile ctx access while the function `CSTTCp::Update()` is updating the statistics.

```
void TrexAstfPerProfile::profile_change_state(state_e new_state) {
    auto old_state = m_profile_state;
    
    switch ( new_state ) {
        case STATE_IDLE:
            m_active_cores = 0;
            break;
        case STATE_LOADED:
            m_stt_cp->m_update = true;  // <-- left profile ctx can be accessed by this on default profile case.
            m_profile_stopping = false;
            m_active_cores = 0;
            break;
```

jsmoon:
@hhaim, no, it clears `m_profile_ctx` in a CP profile (`TrexAstfPerProfile` instance), not all the profiles.
```
TrexAstfPerProfile::m_stt_cp->m_sts[].m_profile_ctx
```
It contains only `TrexAstfPerProfile::m_dp_profile_id` related DP profile context pointers in the DP cores.

hhaim:
@jsmoon who will call to ```update_profile_ctx()``` after that?
if I understand the code right, the TrexAstfPerProfile::m_stt_cp->m_sts[].m_profile_ctx will be empty after this call (```clear_profile_ctx()```) so, you will loose the counters for all the active ASTF profiles. 
However I see this  (```clear_profile_ctx()```) in case of profile ```build``` so it might works but I don't understand how

jsmoon:
@hhaim, as the comment left, it is not called in ASTF interactive mode.
```
                if (!stx) { // non-ASTF case. ASTF updated by DP to CP message
                    lpstt->update_profile_ctx();
                }
```
The `TrexAstfPerProfile::m_stt_cp->m_sts[].m_profile_ctx` will be filled by DP to CP message `TrexDpCoreProfileCtx` which is response to the CP's build request.
```
bool
TrexDpCoreProfileCtx::handle(void) {
    get_stx()->add_dp_profile_ctx(m_profile_id, m_client_pctx, m_server_pctx); // <-- each DP cores report they DP profile context pointer respectively.
    get_stx()->dp_core_finished(m_thread_id, m_profile_id);
    return true;
}   
```

hhaim:
@jsmoon 
1) the first problem I see in this code that DP and CP ***update*** the same object (e.g. add to vector) without lock and without any protection. The only thing that we usually  is to allow to read of a share object.

2) Your answer does not help me to understand 
let's say there are A ,B,C active profile and now there is a command to remove B, with your new code it will  remove A and C too . The function that you sent is for a **new** profile D that will add it to the list 

so I still does not understand how this code works from the first place

jsmoon:
@hhaim, I'm sorry my lack of explanation.
I think my comment TrexAstfPerProfile::m_stt_cp->m_sts[].m_profile_ctx makes you confused. Actually, <a TrexAstfPerProfile instance>.m_stt_cp->m_sts[].m_profile_ctx would be better. And the m_stt_cp is dedicated to the profile instance.

1. DP does not touch this vector object. Only CP updates it. That's the why DP to CP messsage delivers the DP profile context pointer.
2. The m_stt_cp vector object is dedicated to the B profile. The profile A and C remain active. They have also their own m_profile_ctx vector objects which contain their own DP profile contexts. If my code causes the problem you said, the same code in the build() should have caused the problem also.

I'm sorry if my prevous comments made you confused.

hhaim:
@jsmoon 

1) Understood. 
2) Understood. 

so back to this 
>> @hhaim, in my analysis, there is a case that can access the removed profile counter.
Since the default profile reuses the m_stt_cp, the DP profile ctx left on CSTTCp::m_sts[].m_profile_ctx after the DP profile ctx

Can you show me the code that set the default ```m_stt_cp``` as an old profile with a profile_id?

jsmoon:
@hhaim in the following code, the default profile is not deleted. Instead, `profile_init()` is called.
```
bool TrexAstfProfile::delete_profile(cp_profile_id_t profile_id) {
    if (!is_valid_profile(profile_id)) {
        return false;
    }

    auto profile = m_profile_list[profile_id];
    profile->clear_counters(false);
    profile->publish_astf_profile_clear();

    if (profile_id == DEFAULT_ASTF_PROFILE_ID) {
        profile->profile_change_state(STATE_IDLE);
        profile->profile_init(); // <--- here
    } else {
        m_profile_id_map.erase(profile->get_dp_profile_id());

        auto state = profile->get_profile_state();
        assert(m_states_cnt[state]>0);
        m_states_cnt[state]--;

        delete m_profile_list.find(profile_id)->second;
        m_profile_list.erase(profile_id);
    }

    return true;
}
```
But, `profile_init()` doesn't touch the `m_stt_cp` allocated in the TrexAstfPerProfile constructor.
```
void TrexAstfPerProfile::profile_init() {
    profile_check_whitelist_states({STATE_IDLE, STATE_LOADED});
    if ( m_profile_state == STATE_LOADED ) {
        profile_change_state(STATE_IDLE);
    }
    m_profile_buffer.clear();
    m_profile_hash.clear();
    m_profile_parsed = false;
}
```

hhaim:
@jsmoon understood. Nice analysis. What you are saying that once the default profile move to STATE_LOADED (before build that clean it in the CP) there is time that old counters uses dp.ctx pointers. The reason is that DP.ctx (profile_id==0) is freed and reallocated so the old pointer is used in this delta-time. 

I will try it

hhaim:
rebased

PR #723, opened 2021-06-10
dpdk: fix mbuf chain handling in the memif driver

### LINKED ISSUE #686: ip4 length > l2 length with VPP + memif + TRex
When trying any of the ASTF TRex profiles with VPP, I get multiple errors in VPP: `ip4 length > l2 length`.

These don't happen with each packet but rather with some of them. It looks like the TCP payload is truncated down to 600 bytes for some reason, while looking at the TRex simulation, it should be 1448 bytes.  I have collected a snippet from dispatch trace dissected by Wireshark and other info here:
https://gist.github.com/ivan4th/e0130dac619f91a6bb8e7987550c2466

VPP version: 21.01.
I'm not really sure if this is a VPP or TRex problem (or both?), so I'm posting this to the vpp-dev mailing list too.
There was an old ticket related to a similar problem in VPP CSIT: https://jira.fd.io/browse/VPP-1676

Passing `--tso-disable --lro-disable --checksum-offload-disable` flags to TRex don't change anything.

vpp-dev mailing list post: https://lists.fd.io/g/vpp-dev/message/19267

The problem doesn't happen in STF or STL modes.

ivan4th:
Fixes #686

### DIFF
--- a/src/dpdk/drivers/net/memif/rte_eth_memif.c
+++ b/src/dpdk/drivers/net/memif/rte_eth_memif.c
@@ -657,7 +657,7 @@ next_in_chain:
 			d0->length = dst_off;
 		}
 
-		if (rte_pktmbuf_is_contiguous(mbuf) == 0) {
+		if (mbuf->next != NULL) {
 			mbuf = mbuf->next;
 			goto next_in_chain;
 		}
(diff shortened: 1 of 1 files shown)


hhaim:
rebased

PR #726, opened 2021-06-11
Update get_files_list method to prevent server crash under python3

smn-3-14:
The old usage of next causes the server to crash when used with Python3. This should work for Python2.6+.

See: https://portingguide.readthedocs.io/en/latest/iterators.html#new-iteration-protocol-next


hhaim:
rebased

PR #728, opened 2021-06-14
Fix crash in memif_stats_get() in DPDK memif driver

ivan4th:
There was confusion between C2S (client->server) and
S2C (server->client) rings, causing a crash if there's different
number of RX and TX queues, which is the case with TRex with more
than 1 core configured.

### DIFF
--- a/src/dpdk/drivers/net/memif/rte_eth_memif.c
+++ b/src/dpdk/drivers/net/memif/rte_eth_memif.c
@@ -1388,8 +1388,8 @@ memif_stats_get(struct rte_eth_dev *dev, struct rte_eth_stats *stats)
 	stats->opackets = 0;
 	stats->obytes = 0;
 
-	tmp = (pmd->role == MEMIF_ROLE_CLIENT) ? pmd->run.num_c2s_rings :
-	    pmd->run.num_s2c_rings;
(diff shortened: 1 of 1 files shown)


ivan4th:
It is worth noting that TRex still has suboptimal performance with memifs if more than 2 or 3 cores are used, regardless of `--software` RSS setting. At this point, I was unable to find out why.

hhaim:
@ivan4th 
>> It is worth noting that TRex still has suboptimal performance with memifs if more than 2 or 3 cores are used, regardless of --software RSS setting. At this point, I was unable to find out why.

in regards to this,can you provide numbers?  it is better to ask this in DPDK driver forum, trex is scalable in PF driver case (e.g. ix40e) so I assume this is a driver issue and not a trex-core issue

hhaim:
rebased

ivan4th:
> in regards to this,can you provide numbers?
That's smth along the lines of not being able to squeeze more than 6.5..7 Gbps from TRex<->VPP memif link, with 2 cores for TRex and 5 worker cores for VPP. When N of cores is increased from 2, the speed goes down to smth like 3 Gbps IIRC.
But yes, that's likely another memif driver issue, as indeed there's no such problem when TRex uses an XL710 card on the same machine. Will post more info there when I will have a possibility to get back to this problem someday later.

PR #730, opened 2021-06-15
Changes to remove macro LATENCY_IEEE_1588_TIMESTAMPING in code

abhiramnarayana:
Adding feature support ‘--latency-ieee-1588’ . With this Latency
measurements use HW timestamping and without it default behavior
same as existing mechanism.
Note: Since there is some work involved in DPDK to keep both
Vector code and IEEE 1588 code working together currently we
cannot enable the DPDK config( RTE_LIBRTE_IEEE1588) by default.
So, in order to use this feature (--latency-ieee-1588) we need to
enable RTE_LIBRTE_IEEE1588 config COMPILE time and then we can
dynamically enable or disable HW timestamping based on requirement.

Signed-off-by: Abhiram R N <arn@redhat.com>

### DIFF
--- a/doc/trex_stateless.asciidoc
+++ b/doc/trex_stateless.asciidoc
@@ -3687,6 +3687,8 @@ Example using Python API:
 We expect next packet to arrive with sequence number 16.
 * Continuation of Scenario 2: Receive a packet with seq num 11. We increment 'seq_too_low' by 1. We increment 'out_of_order' by 1. We *decrement* 'dropped' by 1. (The assumption is that one of the packets that was considered as dropped before has actually arrived out of order.)
 
+
+
 ==== Tutorial: HLT profiles
 
 HLTAPI traffic_config() function has a set of arguments for creating streams. +
@@ -3995,6 +3997,48 @@ You can simulate the profile using the STL simulator:
 ----
 [bash]>./stl-sim -f stl/dependent_field_engine_vars.py -o b.pcap
 ----
+
+
+==== Tutorial: Latency using hardware assist
+
+This feature is a combination of software and hardware NIC’s IEEE 1588 support. It is disable by default and should be enabled by this compilation flag RTE_LIBRTE_IEEE1588. When enabled there is a 30%-10% performance impcat on i40e and mlx5 drivers.
+In HW terminology it is also referred as TimeSync feature. NIC can take timestamps of sent and received latency packets.
+Using DPDK APIs we can read these timestamps from the NIC HW register. Using the 2-step mechanism of IEEE 1588 (PTP) protocol we send the accurate timestamp (t1) which is read from the NIC in a Follow Up packet rather than in the same packet. Received timestamp (t2) is still taken on the first packet (SYNC) itself. In this way with a combination of software(PTP protocol, DPDK APIs) and HW support (leveraging IEEE 1588 support) more accurate latency and jitter measurements can be obtatined.
+
+*Limitations*::
+
+There are few limitations in this method of Latency Measurement which are listed below. (Some are Temporary)
+
+* The type and size of the latency packet is fixed. IP UDP packet with UDP payload carrying the PTP packet. Size is *106* bytes.
+* Since IEEE 1588 (PTP) protocol uses a well known port number. The latency stream which we create need to use destination port as *319*.
+* Not All types of NIC support this feature. Currently tested and verified on I40E. Around 8 types support IEEE 1588v2 as per DPDK documentation.
+* Since there is some work involved in DPDK to keep both Vector code and IEEE 1588 code working together currently we cannot enable the DPDK config (*RTE_LIBRTE_IEEE1588*) by default. So, in order to use this feature we need to enable RTE_LIBRTE_IEEE1588 config COMPILE time and then we can dynamically enable or disable this feature based on requirement.
+
+
+*How to Use*::
+
+There is a new property added in the Latency stream through which we can enable or disable IEEE_1588 in python script ( for enabling, ieee_1588 = true). Other than that the Type, packet size and port number needs to be set as mentioned above in limitations section. Also please refer to the example provided for the same. have a look
+
+`scripts/stl/udp_1pkt_src_ip_split_latency_ieee_1588.py`
+
+[source,python]
+----
+        # packet for IEEE 1588
+        base_lat_pkt = Ether()/IP(src=src_ip,dst=dst_ip)/UDP(dport=319,sport=1025)
+
+        pkt = STLPktBuilder(pkt = base_pkt/pad,
+                            vm = vm)
+        stream = [STLStream(packet = pkt,
+                            mode = STLTXCont(pps=1)),
+
+                  # latency stream
+                  STLStream(packet = STLPktBuilder(pkt = base_lat_pkt/pad_latency),
+                            mode = STLTXCont(pps=1000),
+                            flow_stats = STLFlowLatencyStats(pg_id = 12+port_id, ieee_1588 = True)) # enable 1588
+
+----
+
+You can query if the driver support this feature using get `get_system_info` API
 
 === Dynamic multiple profiles
 
--- a/doc/trex_stateless_latency_ieee_1588.asciidoc
+++ b/doc/trex_stateless_latency_ieee_1588.asciidoc
@@ -1,0 +1,26 @@
+Latency Measurments using IEEE 1588 packets in Latency Stream
+=============================================================
+
+== Feature Overview
+
+This feature is a combination of software and hardware NIC’s IEEE 1588 support. In HW terminology it is also referred as TimeSync feature. NIC can take timestamps of sent and received latency packets. Using DPDK APIs we can read these timestamps from the NIC HW register. Using the 2-step mechanism of IEEE 1588 (PTP) protocol we send the accurate timestamp (t1) which is read from the NIC in a Follow Up packet rather than in the same packet. Received timestamp (t2) is still taken on the first packet (SYNC) itself. In this way with a combination of software(PTP protocol, DPDK APIs) and HW support (leveraging IEEE 1588 support) more accurate latency and jitter measurements can be obtatined.
+
+== Limitations
+
+There are few limitations in this method of Latency Measurement which are listed below. (Some are Temporary)
+
+* The type and size of the latency packet is fixed. IP UDP packet with UDP payload carrying the PTP packet. Size is *106* bytes.
+* Since IEEE 1588 (PTP) protocol uses a well known port number. The latency stream which we create need to use destination port as *319*.
+* Not All types of NIC support this feature. Currently tested and verified on I40E. Around 8 types support IEEE 1588v2 as per DPDK documentation.
+* Since there is some work involved in DPDK to keep both Vector code and IEEE 1588 code working together currently we cannot enable the DPDK config (*RTE_LIBRTE_IEEE1588*) by default. So, in order to use this feature we need to enable RTE_LIBRTE_IEEE1588 config COMPILE time and then we can dynamically enable or disable this feature based on requirement.
+
+== How to Enable this Feature?
+
+Enable *RTE_LIBRTE_IEEE1588* in the corresponding DPDK config file (for example: src/pal/linux_dpdk/dpdk_2102_x86_64/rte_config.h) and build trex-core and copy the executables to the TRex Client.
+
+
+== How to Use?
+
+There is a new property added in the Latency stream through which we can enable or disable IEEE_1588 in python script ( for enabling, ieee_1588 = true). Other than that the Type, packet size and port number needs to be set as mentioned above in limitations section. Also please refer to the example provided for the same.
+
+scripts/stl/udp_1pkt_src_ip_split_latency_ieee_1588.py
--- a/scripts/automation/regression/functional_tests/stl_basic_tests.py
+++ b/scripts/automation/regression/functional_tests/stl_basic_tests.py
@@ -289,6 +289,7 @@ class CStlBasic_Test(functional_general_test.CGeneralFunctional_Test):
             'imix_wlc.py',          # expects tunables
             'udp_1pkt_vxlan.py',    # uses custom Scapy layer
             'icmpv6_fix_cs.py',     # cannot parse layer name from offset correctly (ICMPv6ND_NS classname and layer name do not match)
+            'udp_1pkt_src_ip_split_latency_ieee_1588.py', # cannot be tested on sim as IEEE 1588 is only supported on specific NICs
             ]
         exclude_dict = {}.fromkeys(exclude_list)
         output_file = os.path.join(self.generated_path, 'exported_to_code.py')
--- a/scripts/automation/trex_control_plane/interactive/trex/common/trex_client.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/common/trex_client.py
@@ -3424,6 +3424,7 @@ class TRexClient(object):
             print('  LED status:    %s' % info['led_change_supported'])
             print('  Flow control:  %s' % info['fc_supported'])
             print('  VXLAN FS:      %s' % info['is_vxlan_supported'])
+            print('  IEEE 1588:     %s' % info['is_ieee1588_supported'])
             print('')
         else:
             if not opts.ports:
--- a/scripts/automation/trex_control_plane/interactive/trex/common/trex_port.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/common/trex_port.py
@@ -668,6 +668,11 @@ class Port(object):
         else:
             info['is_vxlan_supported'] = 'N/A'
 
+        if 'is_ieee1588_supported' in info:
+            info['is_ieee1588_supported'] = 'yes' if info['is_ieee1588_supported'] else 'no'
+        else:
+            info['is_ieee1588_supported'] = 'N/A'
+
         if 'is_virtual' in info:
             info['is_virtual'] = 'yes' if info['is_virtual'] else 'no'
         else:
--- a/scripts/automation/trex_control_plane/interactive/trex/stl/trex_stl_streams.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/stl/trex_stl_streams.py
@@ -253,11 +253,12 @@ STLStreamDstMAC_PKT     =1
 STLStreamDstMAC_ARP     =2
 
 class STLFlowStatsInterface(object):
-    def __init__ (self, pg_id, vxlan):
+    def __init__ (self, pg_id, vxlan, ieee_1588):
         self.fields = {
             'enabled':      True,
             'stream_id':    pg_id,
             'vxlan':        vxlan,
+            'ieee_1588':    ieee_1588,
             }
 
     def to_json (self):
@@ -309,8 +310,8 @@ class STLFlowStats(STLFlowStatsInterface):
 
     """
 
-    def __init__(self, pg_id, vxlan = False):
-        super(STLFlowStats, self).__init__(pg_id, vxlan)
+    def __init__(self, pg_id, vxlan = False, ieee_1588 = False):
+        super(STLFlowStats, self).__init__(pg_id, vxlan, ieee_1588)
         self.fields['rule_type'] = 'stats'
 
 
@@ -325,8 +326,8 @@ class STLFlowLatencyStats(STLFlowStatsInterface):
 
     """
 
-    def __init__(self, pg_id, vxlan = False):
-        super(STLFlowLatencyStats, self).__init__(pg_id, vxlan)
+    def __init__(self, pg_id, vxlan = False, ieee_1588 = False):
+        super(STLFlowLatencyStats, self).__init__(pg_id, vxlan, ieee_1588)
         self.fields['rule_type'] = 'latency'
 
 
--- a/scripts/stl/udp_1pkt_src_ip_split_latency_ieee_1588.py
+++ b/scripts/stl/udp_1pkt_src_ip_split_latency_ieee_1588.py
@@ -1,0 +1,77 @@
+from trex_stl_lib.api import *
+import argparse
+
+
+
+# split the range of IP to cores 
+# add tunable by fsize to change the size of the frame 
+# latency frame is always 106 (uses IEEE 1588 HW support and DPDK APIs)
+# trex>start -f stl/udp_1pkt_src_ip_split_latency_ieee_1588.py -t fsize=64 -m 30% --port 0 --force
+#
+# 
+
+class STLS1(object):
+
+    def __init__ (self):
+        self.fsize  =64;
+        self.lfsize  =106; #The packets size in the latency stream with IEEE-1588 is fixed
+
+
+    def create_stream (self, dir,port_id):
+        # Create base packet and pad it to size
+        size = self.fsize - 4; # HW will add 4 bytes ethernet FCS
+
+        if dir==0:
+            src_ip="16.0.0.1"
+            dst_ip="48.0.0.1"
+        else:
+            src_ip="48.0.0.1"
+            dst_ip="16.0.0.1"
+
+        base_pkt = Ether()/IP(src=src_ip,dst=dst_ip)/UDP(dport=12,sport=1025)
+
+        pad = max(0, size - len(base_pkt)) * 'x'
+        pad_latency = max(0, (self.lfsize-4) - len(base_pkt)) * 'x'
+                             
+        vm = STLScVmRaw( [   STLVmFlowVar ( "ip_src",  min_value="10.0.0.1",
+                                            max_value="10.0.0.255", size=4, step=1,op="inc"),
+                             STLVmWrFlowVar (fv_name="ip_src", pkt_offset= "IP.src" ), # write ip to packet IP.src
+                             STLVmFixIpv4(offset = "IP")                                # fix checksum
+                                  ]
+                              ,cache_size =255 # the cache size
+                              );
+        #dport should be 319. Well known port for IEEE 1588 (PTP protocol)
+        base_lat_pkt = Ether()/IP(src=src_ip,dst=dst_ip)/UDP(dport=319,sport=1025)
+
+        pkt = STLPktBuilder(pkt = base_pkt/pad,
+                            vm = vm)
+        stream = [STLStream(packet = pkt,
+                            mode = STLTXCont(pps=1)),
+
+                  # latency stream   
+                  STLStream(packet = STLPktBuilder(pkt = base_lat_pkt/pad_latency),
+                            mode = STLTXCont(pps=1000),
+                            flow_stats = STLFlowLatencyStats(pg_id = 12+port_id, ieee_1588 = True))
+  
+        ]
+        return stream
+
+
+    def get_streams (self, direction, tunables, **kwargs):
+        parser = argparse.ArgumentParser(description='Argparser for {}'.format(os.path.basename(__file__)), 
+                                         formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+        parser.add_argument('--fsize',
+                            type=int,
+                            default=64,
+                            help="The packets size in the regular stream")
+        args = parser.parse_args(tunables)
+        self.fsize = args.fsize
+        return self.create_stream(direction,kwargs['port_id'])
+
+
+# dynamic load - used for trex console or simulator
+def register():
+    return STLS1()
+
+
+
--- a/src/bp_sim.cpp
+++ b/src/bp_sim.cpp
@@ -4945,6 +4945,14 @@ int CErfIFStl::update_mac_addr_from_global_cfg(pkt_dir_t  dir, uint8_t * p){
     return (0);
 }
 
+void CErfIFStl::fill_fsp_head(struct flow_stat_payload_header *fsp_head, uint16_t hw_id) {
+    fsp_head->seq = 0x12345678;
+    fsp_head->hw_id = hw_id - MAX_FLOW_STATS;
+    fsp_head->magic = FLOW_STAT_PAYLOAD_MAGIC;
+    fsp_head->flow_seq = FLOW_STAT_PAYLOAD_INITIAL_FLOW_SEQ;
+    fsp_head->time_stamp = 0x8899aabbccddeeff;
+}
+
 int CErfIFStl::send_sl_node(CGenNodeStateless *node_sl) {
     pkt_dir_t dir=(pkt_dir_t)node_sl->get_mbuf_cache_dir();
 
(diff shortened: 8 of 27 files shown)


hhaim:
@abhiramnarayana suggest to convert ```--latency-ieee-1588``` global configuration to be per-stream json configuration (a boolean that enable hardware assist if exists), this way this feature will be available to any user that has the right hardware

hhaim:
@abhiramnarayana one more thing, you need to merge to latest code as it was change a bit

abhiramnarayana:
Hi @hhaim ,

Thanks for looking into this. 

1) Regarding the merge, yeah I will do that . 

2) Regarding the 'per-stream json configuration' you mentioned.. I have one thing to discuss and also one query.
Query : Can you please help me to which file you are referring to. Or a sample json configuration which I can keep as reference to implement that way.

Note that, we are doing the enabling of timesync for using the HW capabilities 'per port' (i.e per interface).  We are doing that immediately after the 'rte_eth_dev_start' returns success.
i.e We are calling 'rte_eth_timesync_enable' per port and this we need to do in the beginning and not while we start playing the streams.
With the above in mind I am not sure if the 'per-stream json config' method will fit in here.
If you still feel this needs changing let me know.

Thanks & Regards,
Abhiram R N

hhaim:
@abhiramnarayana 

1. I don't see a problem to enable timesync  for each NIC that support it (add a global ```--disable-ieee-1588```)
2. To use it (tx+rx) (the mbuf->ieee) only if the stream.ieee_1588=true

abhiramnarayana:
@hhaim ,

Sorry to bother you with questions!. Please pardon me. Just want to be sure I understand correctly what you explained.

Just few last clarifications... PSB.

1. I don't see a problem to enable timesync for each NIC that support it (add a global --disable-ieee-1588)
[Abhiram] - Agreed. Yes will change accordingly.

2. To use it (tx+rx) (the mbuf->ieee) only if the stream.ieee_1588=true
[Abhiram] - Here, If I understood this 'stream.ieee_1588'   I should be adding it in 'trex_stl_stream.h' same way following examples of m_start_paused, m_self_start. Hope that is what you intended?

For,  mbuf->ieee ---> this one will be in DPDK code right? 
Do you want me to do this change in the 'rte_mbuf_core.h' (under the #ifdef TREX_PATCH)?(for example like - m_core_locality)

Thanks & Regards,
Abhiram R N

hhaim:
@abhiramnarayana 
in regards to (2) you should add it as a property of the stream like 

```
    /* RX check */
    struct {
        bool      m_enabled;
        bool      m_seq_enabled;
        bool      m_latency;
        bool      m_vxlan_skip;
        uint16_t  m_rule_type;
        uint32_t  m_pg_id;
        uint16_t  m_hw_id;
        bool    ieee_1588; <<< new
    } m_rx_check;

```

and the json format in this command should support this option

```
trex_rpc_cmd_rc_e
TrexRpcCmdAddStream::_run(const Json::Value &params, Json::Value &result) {

    try {
        uint8_t port_id = parse_port(params, result);

        string profile_id = parse_profile(params, result);


...


        /* if it is enabled - we need more fields */
        if (stream->m_rx_check.m_enabled) {

            if (port->get_rx_caps() == 0) {
                generate_parse_err(result, "RX stats is not supported on this interface");
            }


```


and of course the python side that defines a stream should add this. 


2. You should have a way to query the capability of the trex-server if it support this mode or not 
better to add it to ````TREX_RPC_CMD_EXT(TrexRpcCmdGetSysInfo,        "get_system_info",``` command 



```
trex_rpc_cmd_rc_e
TrexRpcCmdGetSysInfo::_run(const Json::Value &params, Json::Value &result) {
    string hostname;

    TrexPlatformApi &api = get_platform_api();
    
    Json::Value &section = result["result"];

    get_hostname(hostname);
```

3. You can't do that ! you are changing the way flow-stats works even for cases that hardware does not support IEEE_1588
you should keep the flow_stats header and create a new header for streams that has a ieee headers. 

```
struct flow_stat_payload_header {
    /*
     * Intentionally kept at the begining.
     * The PTP packet has to begin in the
     * UDP payload immediately after UDP
     * header ends.
     */
    struct sync_msg ptp_message;  <<<<<<<<<<<<<
    uint8_t magic;
    uint8_t flow_seq;
    uint16_t hw_id;
    uint32_t seq;
    uint64_t time_stamp;
} __attribute__((__packed__));

```

hhaim:
@abhiramnarayana one more thing
1. You should add a feature documentation in the STL tutorial  
2. You should add a test in our regression tests

abhiramnarayana:
Hi @hhaim ,

Thanks for your inputs.
I have incorporated your comments. 

Only the tutorial is pending.(Last point)

Summarizing the changes

1. enable timesync for each NIC that support it (add a global --disable-ieee-1588)  - DONE
2.  To use it (tx+rx) (the mbuf->ieee) only if the stream.ieee_1588=true    - DONE 
3.  json format in this command should support this option - TrexRpcCmdAddStream:  - DONE
4.  You should have a way to query the capability of the trex-server if it support this mode or not - Changes in  TrexRpcCmdGetSysInfo  - DONE 
5. create a new header for streams that has a ieee headers. - DONE
6.  You should add a test in our regression tests  - DONE (Added one regression test in trex_stateless_gtest.cpp)

7.  You should add a feature documentation in the STL tutorial  -- TO BE DONE - Sure, I will do it. Please provide some details where exactly I should add the details? Is there some steps to be followed?

Thanks & Regards,
Abhiram

hhaim:
@abhiramnarayana why do you need this field in mbuf? 
```
#ifdef TREX_PATCH
      bool      m_ieee1588;
 ```
2. try to reuse the same struct with flow_stat_payload_header_ieee_1588  and don't cut and past the header 
```
struct flow_stat_payload_header_ieee_1588 {
    /*
     * Intentionally kept at the begining.
     * The PTP packet has to begin in the
     * UDP payload immediately after UDP
     * header ends.
     */
    struct sync_msg ptp_message;
    uint8_t magic;
    uint8_t flow_seq;
    uint16_t hw_id;
    uint32_t seq;
    uint64_t time_stamp;
<<<<<<< HEAD
#ifdef LATENCY_IEEE_1588_TIMESTAMPING
} __rte_packed;
#else
    bool is_valid_ts(hr_time_t now);
};
=======
} __attribute__((__packed__));
>>>>>>> 034ae106c1499ee21a1b184de0fa9c5eccb1a538

struct flow_stat_payload_header {
    uint8_t magic;
    uint8_t flow_seq;
    uint16_t hw_id;
    uint32_t seq;
    uint64_t time_stamp;
};

```

**Comment by hhaim** [comment]
@abhiramnarayana one more thing, please try to fix the conflicts

**Comment by abhiramnarayana** [comment]
Hi @hhaim ,

1) Regarding 'mbuf' maybe I misunderstood your comment "To use it (tx+rx) (the **mbuf->ieee)** only if the stream.ieee_1588=true"  . I will try to see if that can be avoided. Probably that I can remove that. 

2) Sure, i will reuse the structure instead of copying its members 

3) Ohh I need to to do one more rebase. Will do it and upload.

Thanks & Regards,
Abhiram R N

**Comment by hhaim** [comment]
@abhiramnarayana 


just started to look into the code and it has a lot of duplication and it is not clean please review it again, 

for example

1.```handle_flow_latency_stats```


1.1. a duplication of code 
1.2. performance impact  in case this feature is not used at all -- create a ref-counter in CP to know if this feature is used or not to skip the check 

```
bool RXLatency::handle_flow_latency_stats(const rte_mbuf_t *m, int port, uint32_t& ip_id,bool check_non_ip) {
    bool readtime=false;
    hr_time_t hr_time_now =0;

    uint8_t tmp1_buf[sizeof(struct flow_stat_payload_header_ieee_1588)];
    uint8_t m_ieee1588 = false;

    struct flow_stat_payload_header_ieee_1588 *fsp_head_ieee_1588 =
        (flow_stat_payload_header_ieee_1588 *)utl_rte_pktmbuf_get_last_bytes(
            m, sizeof(struct flow_stat_payload_header_ieee_1588), tmp1_buf);

    if (((fsp_head_ieee_1588)->ptp_message.hdr.seq_id == htons(fsp_head_ieee_1588->fsp_hdr.seq))
        && (fsp_head_ieee_1588->ptp_message.hdr.message_length == htons(60))
        && (fsp_head_ieee_1588->ptp_message.hdr.flag_field[0] == 0x2)) {
        m_ieee1588 = true;
    }

    if ( CGlobalInfo::m_options.preview.getLatencyIEEE1588Disable() || m_ieee1588 == false) {
        uint8_t tmp_buf[sizeof(struct flow_stat_payload_header)];
        struct flow_stat_payload_header *fsp_head = 0;

        if (check_non_ip) {
            hr_time_now = os_get_hr_tick_64();
            readtime = true;
            fsp_head = (flow_stat_payload_header *)utl_rte_pktmbuf_get_last_bytes(
                    m, sizeof(struct flow_stat_payload_header), tmp_buf);

            /* check if flow-latency-stat packet by payload magic.
             * for the non-IP protocol (invalid ip_id) and IP proxy (changed ip_id) cases.
             */
            if ( (fsp_head->magic == FLOW_STAT_PAYLOAD_MAGIC) &&
                 (fsp_head->hw_id < MAX_FLOW_STATS_PAYLOAD)&& 
                 fsp_head->is_valid_ts(hr_time_now) ) {
                ip_id = FLOW_STAT_PAYLOAD_IP_ID;
            }
        }

        if (is_flow_stat_payload_id(ip_id)) {
            if (!fsp_head){
                fsp_head = (flow_stat_payload_header *)utl_rte_pktmbuf_get_last_bytes(
                    m, sizeof(struct flow_stat_payload_header), tmp_buf);
            }
            if (!readtime){
                hr_time_now = os_get_hr_tick_64();
                readtime = true;
            }
            update_stats_for_pkt(fsp_head, m->pkt_len, hr_time_now);
            return true;
        }

    } else {
        uint8_t tmp_buf_ieee_1588[sizeof(struct flow_stat_payload_header_ieee_1588)];
        struct flow_stat_payload_header_ieee_1588 *fsp_head_ieee_1588 = 0;

        if (check_non_ip) {
            hr_time_now = os_get_hr_tick_64();
            readtime = true;
            fsp_head_ieee_1588 = (flow_stat_payload_header_ieee_1588 *)utl_rte_pktmbuf_get_last_bytes(
                                  m, sizeof(struct flow_stat_payload_header_ieee_1588), tmp_buf_ieee_1588);

            /* check if flow-latency-stat packet by payload magic.
             * for the non-IP protocol (invalid ip_id) and IP proxy (changed ip_id) cases.
             */
            if ( (fsp_head_ieee_1588->fsp_hdr.magic == FLOW_STAT_PAYLOAD_MAGIC) &&
                 (fsp_head_ieee_1588->fsp_hdr.hw_id < MAX_FLOW_STATS_PAYLOAD)&& 
                 fsp_head_ieee_1588->fsp_hdr.is_valid_ts(hr_time_now) ) {
                ip_id = FLOW_STAT_PAYLOAD_IP_ID;
            }
        }

        if (is_flow_stat_payload_id(ip_id)) {
            if (!fsp_head_ieee_1588){
                fsp_head_ieee_1588 = (flow_stat_payload_header_ieee_1588 *)utl_rte_pktmbuf_get_last_bytes(
                    m, sizeof(struct flow_stat_payload_header_ieee_1588), tmp_buf_ieee_1588);
            }
            if (!readtime){
                hr_time_now = os_get_hr_tick_64();
                readtime = true;
            }
            if (m->ol_flags & PKT_RX_IEEE1588_TMST) {
                save_timestamps_for_sync_pkt(m, fsp_head_ieee_1588, port);
                m_rx_pg_stat_payload[fsp_head_ieee_1588->fsp_hdr.hw_id].add_pkts(1);
                m_rx_pg_stat_payload[fsp_head_ieee_1588->fsp_hdr.hw_id].add_bytes(m->pkt_len + 4); // +4 for ethernet CRC
            } else {
                update_timestamps_for_fup_pkt(fsp_head_ieee_1588);
                update_stats_for_pkt_ieee_1588(fsp_head_ieee_1588, m->pkt_len, m_sync_arrival_time_nsec);
            }
            return true;
        }
    }
    return false;
}
```

2.

```
        if (node_sl->is_stat_needed() && (node_sl->get_stat_hw_id() >= MAX_FLOW_STATS) ) {
            /* latency packet. flow stat without latency handled like normal packet in simulation */
            uint16_t hw_id = node_sl->get_stat_hw_id();
            rte_mbuf_t *mi;
            if ( CGlobalInfo::m_options.preview.getLatencyIEEE1588Disable() || m_ieee1588 == false) {
```

why do you need to check ```CGlobalInfo::m_options.preview.getLatencyIEEE1588Disable()``` this in this stage, just make sure the stream in the CP is disabled and skip this condition checking in all the code 

 3. isn't the ieee1588 packet has the limited structure and limited size? why do you need to make this function so complex to support all the combination, just allocated the right size without 2/3 mbufs ..
```
rte_mbuf_t * CGenNodeStateless::alloc_flow_stat_mbuf_ieee_1588(rte_mbuf_t *m,
                                     struct flow_stat_payload_header_ieee_1588 *&fsp_head) {
```

**Comment by abhiramnarayana** [comment]
Hi @hhaim ,

Mainly regarding 1.2 I want your help. Remaining other issues I have fixed them and given details as below. Please check and let me know your comments.

1.1 - I have removed the duplication of code in handle_flow_latency_stats - We would not come there for check_non_ip = true and also the is_flow_stat_payload_id(ip_id) is already checked. So IEEE 1588 case will not have any duplicate code and will look as below. 

    } else {
        uint8_t tmp_buf_ieee_1588[sizeof(struct flow_stat_payload_header_ieee_1588)];
        struct flow_stat_payload_header_ieee_1588 *fsp_head_ieee_1588 = 0;

        fsp_head_ieee_1588 = (flow_stat_payload_header_ieee_1588 *)utl_rte_pktmbuf_get_last_bytes(
            m, sizeof(struct flow_stat_payload_header_ieee_1588), tmp_buf_ieee_1588);

        if (m->ol_flags & PKT_RX_IEEE1588_TMST) {
            save_timestamps_for_sync_pkt(m, fsp_head_ieee_1588, port);
            m_rx_pg_stat_payload[fsp_head_ieee_1588->fsp_hdr.hw_id].add_pkts(1);
            m_rx_pg_stat_payload[fsp_head_ieee_1588->fsp_hdr.hw_id].add_bytes(m->pkt_len + 4); // +4 for ethernet CRC
        } else {
            update_timestamps_for_fup_pkt(fsp_head_ieee_1588);
            update_stats_for_pkt_ieee_1588(fsp_head_ieee_1588, m->pkt_len, m_sync_arrival_time_nsec);
        }
        return true;
    }

1.2 - Agree to what you said. When you say ' create a ref-counter in CP to know if this feature is used or not' where should this ref-counter be added ?. Need your inputs for handling this .
Can you kindly point me on how to go about this.

2 - I have removed this below checks everywhere .
        `if ( CGlobalInfo::m_options.preview.getLatencyIEEE1588Disable() || m_ieee1588 == false) {`
     and replaced with below check
        `if ( !(node_sl->is_latency_ieee_1588_enabled()) )`

Basically we are using the Stream ieee_1588 check 

    inline bool is_latency_ieee_1588_enabled() {
        return m_ref_stream_info->m_rx_check.m_ieee_1588;
    }

3 - Agreed.. I have fixed it and no extra logic needed here.  - No doubts.

Thanks & Regards,
Abhiram R N

**Comment by hhaim** [comment]
@abhiramnarayana 
1.2 In the RPC command  (add stream/remove stream)  you can have a global atomic counter that counts the number of active streams with latency_ieee. in the  ```RXLatency::handle_flow_latency_stats``` that handle the packets back you could skip the test of mbuf if the ref-count==0

**Question**: Is there an indication in mbuf for ieee packets?if this is the case it is possible to eliminate the ref-counter and intead using this indication. 

(BTW this idea will not support sending ieee packet from two different trexs )

**Comment by abhiramnarayana** [comment]
Hi @hhaim ,

Thanks for your quick reply.

Yeah, I do see below 2 things in mbuf for ieee 1588 which are there. .

A). m->timesync   
    which I am using in below function call.
    if (rte_eth_timesync_read_rx_timestamp(port, &timestamp, **m->timesync**) < 0) { 

B).  In m->ol_flags  there is a specific bit.      --> Set only for SYNC packet
     if (m->ol_flags & PKT_RX_IEEE1588_TMST) {

Let me try and see if this A) works for our case.
In which case we can use this indication.

Thanks & Regards,
Abhiram R N

**Comment by abhiramnarayana** [comment]
Hi @hhaim ,

Option A)  I tried but we cant use as 'm->timesync' is index and zero is also valid value. 

But, I could successfully use option B)  ' if (m->ol_flags & PKT_RX_IEEE1588_TMST) { ' and make it work.  
This flag is set for SYNC packet and for every SYNC there is a FollowUp packet for which I have added a ref cnt . 
For the FollowUP packet m->ol_flags is not set so we need ref cnt for that.

Thanks & Regards,
Abhiram R N

**Comment by hhaim** [comment]
@abhiramnarayana looks better now

1)
keep this code as a separate function for handling IEEE and keep the normal case as it was (IEEE should be unlikey and as an exception) 

```
bool RXLatency::handle_flow_latency_stats(const rte_mbuf_t *m, int port, uint32_t& ip_id,bool check_non_ip) {
    uint8_t tmp_buf[sizeof(struct flow_stat_payload_header)];
    struct flow_stat_payload_header *fsp_head = 0;
    uint8_t tmp_buf_ieee_1588[sizeof(struct flow_stat_payload_header_ieee_1588)];
    struct flow_stat_payload_header_ieee_1588 *fsp_head_ieee_1588 = 0;
    bool readtime=false;
    hr_time_t hr_time_now =0;

    if (m->ol_flags & PKT_RX_IEEE1588_TMST) {  /* IEEE-1588 - SYNC Pkt */
        m_is_ieee_ref_cnt_set = true; /* Expect a FUP pkt. Set ref cnt */
        fsp_head_ieee_1588 = (flow_stat_payload_header_ieee_1588 *)utl_rte_pktmbuf_get_last_bytes(
            m, sizeof(struct flow_stat_payload_header_ieee_1588), tmp_buf_ieee_1588);

        save_timestamps_for_sync_pkt(m, fsp_head_ieee_1588, port);
        m_rx_pg_stat_payload[fsp_head_ieee_1588->fsp_hdr.hw_id].add_pkts(1);
        m_rx_pg_stat_payload[fsp_head_ieee_1588->fsp_hdr.hw_id].add_bytes(m->pkt_len + 4); // +4 for ethernet CRC
        return true;

    } else if (m_is_ieee_ref_cnt_set == true) { /* IEEE-1588 FUP Pkt */
        m_is_ieee_ref_cnt_set = false; /* Got FUP pkt. Clear the ref cnt */
        fsp_head_ieee_1588 = (flow_stat_payload_header_ieee_1588 *)utl_rte_pktmbuf_get_last_bytes(
            m, sizeof(struct flow_stat_payload_header_ieee_1588), tmp_buf_ieee_1588);

        update_timestamps_for_fup_pkt(fsp_head_ieee_1588);
        update_stats_for_pkt_ieee_1588(fsp_head_ieee_1588, m->pkt_len, m_sync_arrival_time_nsec);
        return true;

    } else {  /* Normal Latency Pkt */

```

2) 
```
        save_timestamps_for_sync_pkt(m, fsp_head_ieee_1588, port);
        m_rx_pg_stat_payload[fsp_head_ieee_1588->fsp_hdr.hw_id].add_pkts(1);
        m_rx_pg_stat_payload[fsp_head_ieee_1588->fsp_hdr.hw_id].add_bytes(m->pkt_len + 4); // +4 for ethernet CRC
```


this code should be converted to a function that check the validity of the hw_id ..
see ```handle_correct_flow code handle_correct_flow_ieee_1588``` 

3) why __rte_packed was replaced with __attribute__((__packed__)); ?

4) need to refactor this function too much duplication ```HOT_FUNC int CCoreEthIF::send_pkt_lat```

4.1) this part should be in a function 

```
    if (likely( CGlobalInfo::m_options.m_is_queuefull_retry )) {
        while ( unlikely( ret != 1 ) ){
            rte_delay_us(1);
            lp_stats->m_tx_queue_full += 1;
            ret = lp_port->m_port->tx_burst(lp_port->m_tx_queue_id_lat, &m, 1);
        }
    } else {
        if ( unlikely( ret != 1 ) ) {
            lp_stats->m_tx_drop ++;
            rte_pktmbuf_free(m);
            return 0;
        }
    }
```

4.2) there should be a separate function for case of     if ( node_sl->is_latency_ieee_1588_enabled())  and normal case, the code looks like spaghetti 


```
HOT_FUNC int CCoreEthIF::send_pkt_lat(CCorePerPort *lp_port, CGenNodeStateless * node_sl, rte_mbuf_t *m, CVirtualIFPerSideStats *lp_stats) {
    // We allow sending only from first core of each port. This is serious internal bug otherwise.
    assert(lp_port->m_tx_queue_id_lat != INVALID_Q_ID);


    struct timespec tstamp_tx;
    //struct flow_stat_payload_header *fsp_head = NULL;
    struct flow_stat_payload_header_ieee_1588 *fsp_head_ieee_1588 = NULL;

    if ( node_sl->is_latency_ieee_1588_enabled()) {
        rte_eth_timesync_read_tx_timestamp(lp_port->m_port->get_tvpid(), &tstamp_tx);

        /*
         * The SYNC packet and Follow up packet are sent using the same mbuf node.
         * We are not generating a new packet. We use the same SYNC packet and just
         * update the timestamp read from the NIC register, ptp msg type and resend it.
         * Hence update the reference count here.
         */
        rte_pktmbuf_refcnt_update(m,1);
    }
    int ret = lp_port->m_port->tx_burst(lp_port->m_tx_queue_id_lat, &m, 1);

    if (likely( CGlobalInfo::m_options.m_is_queuefull_retry )) {
        while ( unlikely( ret != 1 ) ){
            rte_delay_us(1);
            lp_stats->m_tx_queue_full += 1;
            ret = lp_port->m_port->tx_burst(lp_port->m_tx_queue_id_lat, &m, 1);
        }
    } else {
        if ( unlikely( ret != 1 ) ) {
            lp_stats->m_tx_drop ++;
            rte_pktmbuf_free(m);
            return 0;
        }
    }

    if ( node_sl->is_latency_ieee_1588_enabled()) {
        port_ieee1588_tx_timestamp_check(lp_port->m_port->get_tvpid(), &tstamp_tx);
        const rte_mbuf_t *last = m;

        while (last->next != NULL) {
            last = last->next;
        }

        fsp_head_ieee_1588 = (flow_stat_payload_header_ieee_1588 *)(rte_pktmbuf_mtod(m, uint8_t*) + (last->data_len - (sizeof(struct flow_stat_payload_header_ieee_1588))));

        fsp_head_ieee_1588->ptp_message.hdr.msg_type = FOLLOW_UP;
        fsp_head_ieee_1588->ptp_message.hdr.control = 2;
        fsp_head_ieee_1588->ptp_message.origin_tstamp.ns = htonl(tstamp_tx.tv_nsec);
        fsp_head_ieee_1588->ptp_message.origin_tstamp.sec_msb = htons((uint16_t)(tstamp_tx.tv_sec >> 32));
        fsp_head_ieee_1588->ptp_message.origin_tstamp.sec_lsb = htonl((uint32_t)(tstamp_tx.tv_sec & UINT32_MAX));


        m->ol_flags &= ~PKT_TX_IEEE1588_TMST; /* FUP packet doesnt need to be timestampped */

        ret = lp_port->m_port->tx_burst(lp_port->m_tx_queue_id_lat, &m, 1);

        if (likely( CGlobalInfo::m_options.m_is_queuefull_retry )) {
            while ( unlikely( ret != 1 ) ){
                rte_delay_us(1);
                lp_stats->m_tx_queue_full += 1;
                ret = lp_port->m_port->tx_burst(lp_port->m_tx_queue_id_lat, &m, 1);
            }
        } else {
            if ( unlikely( ret != 1 ) ) {
                lp_stats->m_tx_drop ++;
                rte_pktmbuf_free(m);
                return 0;
            }
        }
    }

    return ret;
}
```

**Comment by hhaim** [comment]
@abhiramnarayana one more thing, you are not handling the case that       FOLLOW_UP packet is drop and state machine will stuck in the rx side

**Comment by abhiramnarayana** [comment]
Hi @hhaim ,

Thanks for patiently reviewing and providing your inputs.

I checked the comments and I have fixed the 1, 2 and 4 (4.1 and 4.2) as suggested. ( I have uploaded just now. Regression is running).

Below answers for your questions.
why __rte_packed was replaced with attribute((packed)); ?      
 Using the '__rte_packed' like before gives compilation error for linux build with gcc6. 
Then I checked that in mbuf.cpp (src/pal/linux) __attribute__((packed)); was used which works for both linux and linux_dpdk
Also I checked that in rte_common.h below is defined.
#define __rte_packed __attribute__((__packed__)) .
So effectively it's the same. Hence changed it.

Question regarding  Followup drop. In case the "FOLLOW_UP packet is drop" ...
Though the previous Follow Up(FUP) packet is dropped then once we get the next SYNC packet things should move ahead and not get stuck.Currently to process the SYNC packet we are not dependent on the ref cnt but only on the mbuf flag set by the driver. And  since once we get any next SYNC packet we update the 'm_fup_seq_exp' (expected next sequence number of FUP)  and also reset the ref cnt. 

Thanks & Regards,
Abhiram R N

**Comment by hhaim** [comment]
@abhiramnarayana much better a few more comments. 

1) ``` __attribute__((__packed__));  >>> __rte_packed;```  I would keep the ```__rte_packed``` and just add the implementation to the simulator (linux make file) if you have issue with it. 

2. Better not to enable the ```m_ieee_1588``` flag if the driver does not support it, I would return a RPC error in this case. this is the entry point to enable this feature and the server should verify that it is possible to do so

```
if (stream->m_rx_check.m_enabled) {

            if (port->get_rx_caps() == 0) {
                generate_parse_err(result, "RX stats is not supported on this interface");
            }

            stream->m_rx_check.m_pg_id      = parse_uint32(rx, "stream_id", result);
            stream->m_rx_check.m_vxlan_skip = parse_bool(rx, "vxlan", result, false);
            stream->m_rx_check.m_ieee_1588  = parse_bool(rx, "ieee_1588", result, false); <<<<<<
            std::string type = parse_string(rx, "rule_type", result);
            if (type == "latency") {
                stream->m_rx_check.m_rule_type = TrexPlatformApi::IF_STAT_PAYLOAD;
                if (stream->m_core_id_specified) {
                    std::stringstream ss;
                    ss << "Core ID pinning is not supported for latency streams.";
                    generate_execute_err(result, ss.str());
                }
            } else {
                stream->m_rx_check.m_rule_type = TrexPlatformApi::IF_STAT_IPV4_ID;
            }
        }
```

3. Add a doc, limitation of this new feature 

4. Add a profile that enable this feature 

5. How do you know that the last chunk of mbuf has enough space for ```flow_stat_payload_header_ieee_1588```? In this implementation you alloc **one** mbuf, so it is better to check that you have one mbuf and remove the loop 
```
 while (last->next != NULL) {
        last = last->next;
    }
```

```
 while (last->next != NULL) {
        last = last->next;
    }

    fsp_head_ieee_1588 = (flow_stat_payload_header_ieee_1588 *)(rte_pktmbuf_mtod(m, uint8_t*) + (last->data_len - (sizeof(struct flow_stat_payload_header_ieee_1588))));

    fsp_head_ieee_1588->ptp_message.hdr.msg_type = FOLLOW_UP;
    fsp_head_ieee_1588->ptp_message.hdr.control = 2;
    fsp_head_ieee_1588->ptp_message.origin_tstamp.ns = htonl(tstamp_tx.tv_nsec);
    fsp_head_ieee_1588->ptp_message.origin_tstamp.sec_msb = htons((uint16_t)(tstamp_tx.tv_sec >> 32));
    fsp_head_ieee_1588->ptp_message.origin_tstamp.sec_lsb = htonl((uint32_t)(tstamp_tx.tv_sec & UINT32_MAX));
```


6. Remove duplication of code, see >>>1 and >>>2


```
        if ( !(node_sl->is_latency_ieee_1588_enabled()) ) {
            mi = node_sl->alloc_flow_stat_mbuf(m, fsp_head, is_const);  >>>1
            fsp_head->seq = lp_stats->m_lat_data[hw_id_payload].get_seq_num();
            fsp_head->hw_id = hw_id_payload;
            fsp_head->flow_seq = lp_stats->m_lat_data[hw_id_payload].get_flow_seq();
            fsp_head->magic = FLOW_STAT_PAYLOAD_MAGIC;
        } else {
            mi = node_sl->alloc_flow_stat_mbuf_ieee_1588(m, fsp_head_ieee_1588);  >>>2
            fsp_head_ieee_1588->fsp_hdr.seq = lp_stats->m_lat_data[hw_id_payload].get_seq_num();
            fsp_head_ieee_1588->fsp_hdr.hw_id = hw_id_payload;
            fsp_head_ieee_1588->fsp_hdr.flow_seq = lp_stats->m_lat_data[hw_id_payload].get_flow_seq();
            fsp_head_ieee_1588->fsp_hdr.magic = FLOW_STAT_PAYLOAD_MAGIC;
        }
```

**Comment by hhaim** [comment]
@abhiramnarayana 
I've tried the profile on my setup, it does not seems to work 

I get this 
```
Port 0 TX timestamp registers not valid after 1000 micro-seconds
Port 0 TX timestamp registers not valid after 1000 micro-seconds
```

```
   PG ID     |       12       
-------------+---------------
TX pkts      |           8812 
RX pkts      |           8811 
Max latency  |           1746 
Min latency  |              4 
Avg latency  |            520 
-- Window -- |                
Last max     |           1039 
Last-1       |           1110 
Last-2       |           1042 
Last-3       |           1037 
Last-4       |           1069 
Last-5       |           1038 
Last-6       |           1040 
Last-7       |           1127 
Last-8       |           1044 
Last-9       |           1038 
Last-10      |                
Last-11      |                
Last-12      |                
```

hhaim:
I did some minor changes to doc and printf 
https://github.com/hhaim/trex-core/tree/abhiramnarayana-latency-ieee-1588

abhiramnarayana:
hi @hhaim ,

Only time I got this Error was initially when DPDK config I had missed configuring.
Port 0 TX timestamp registers not valid after 1000 micro-seconds
Port 0 TX timestamp registers not valid after 1000 micro-seconds

Most likely looks like that is the same cause here as well. Just want to know if you have enabled it?.
Some details below.

Like I have put in the Limitation doc. This Config needs to be enabled as it is default cant be kept enabled for now.(Due to conflict with Vector code)
So, in order to use this feature we need to enable RTE_LIBRTE_IEEE1588 config COMPILE time and then we can dynamically enable or disable this feature based on requirement.

File --> src/pal/linux_dpdk/dpdk_2102_x86_64/rte_config.h

At the end below define you can find. Please uncomment it and COMPILE and use if you havent done it.
/*
 * Enable this to test Latency
 * using HW timestamping by using
 * IEEE 1588 feature
 */
//#define RTE_LIBRTE_IEEE1588 1

Regarding the doc and printfs --> Yeah  ok. Should be fine. I will read through once.

Thanks & Regards,
Abhiram R N

hhaim:
@abhiramnarayana the all point is to have it working without changing the defines in the code. 
would the performance will drop if I enable **RTE_LIBRTE_IEEE1588** ?

hhaim:
@abhiramnarayana ok, I manage to make it work using the DEFINE. do you have the patch to make i40e driver to keep working in vector mode even with this define enabled ?

abhiramnarayana:
Hi @hhaim ,

Yeah, currently if we enable 'RTE_LIBRTE_IEEE1588' it will drop performance. I had seen it during my initial commit.

I have tried to make i40e driver to keep working in vector mode even with this define enabled. 
Just we need to REMOVE the check in below line in DPDK code. I had tried with this and had not seen issue in TRex Stateless Mode which I am testing with.  Just keep the code open without that DEFINE check. 

https://github.com/hhaim/trex-core/blob/abhiramnarayana-latency-ieee-1588/src/dpdk/drivers/net/i40e/i40e_rxtx_vec_common.h#L195

But when I contacted I40E driver guys I got below reply 
"that IEEE1588 requires special handles that are in the data path and those equivalent handles are not supported in the vector path.  If you do enable both, the PMD will very likely crash out when it tries to access the handle."

Thanks & Regards,
Abhiram R N

hhaim:
@abhiramnarayana I've just looked into the i40e driver, it does not seem that the functionality is there for vector mode (in case it is enabled). Did you verify that it IEEE 1588 function is there in case of the patch? It a bit lame to get more accurate latency and have a significant  performance drop.

abhiramnarayana:
Hi @hhaim ,

Yes. 
With the patch I have retested IEEE 1588 just now again. 
Just to be sure. Below is patch for your reference. IEEE 1588 works fine by keeping this below change as well.

patchpic

Thanks & Regards,
Abhiram R N

hhaim:
image
 it seems that Mellanox CX-5EX has a performance impact too ..
I will keep the code masked (so it has a lower probability to be used).

hhaim:
@abhiramnarayana there are one more issue
please use this [latest](https://github.com/hhaim/trex-core/tree/abhiramnarayana-latency-ieee-1588) with my minor changes of doc. 

the issue is this, if the feature is disable by define/CLI you should disable the feature in ```add_stream``` but you didn't and I got this error ```Port 0 TX timestamp registers not valid after 1000 micro-seconds``` 
please fix this on top of the branch I've provided

abhiramnarayana:
Hi @hhaim ,

Yeah, need to handle that. 
But I felt we shouldn't be starting the Latency stream itself  when user tries to run this profile(with ieee_1588 = true) and without the DEFINE. So I am throwing an exception.
If user has disabled from CLI (--disable-ieee-1588) and trying to start stream with IEEE 1588 I am disabling the same.
Let me know your comments on the below change. Or should we be throwing exception here also ? (Because we have custom profile for IEEE 1588 profile with custom size and well known port number for IEEE 1588)

ieee1588patch

Thanks & Regards,
Abhiram R N

hhaim:
rebased

PR #659, opened 2021-03-11
prevent surplus packets from ARP on STLFlowLatencyStats

jsmoon:
Hi, this PR is for issue #650 solution.

Since I didn't find the proper stub location for the `rte_ether.h`,
I added `MacAddress::isUnicastAddress` and used it.

@hhaim, please review my change and give your feedback.

PR #731, opened 2021-06-16
prevent surplus packets from ARP on STLFlowLatencyStats

jsmoon:
Hi, this PR is the change for the reverted PR #659.

In this change, only the broadcast mac address will be excluded using by `isDefaultAddress()` method.
I rewrote the `RXLatency::handle_pkt()` also.

@hhaim please check my code and give your feedback.

### DIFF
--- a/src/stx/common/trex_latency_counters.cpp
+++ b/src/stx/common/trex_latency_counters.cpp
@@ -234,56 +234,85 @@ RXLatency::update_timestamps_for_fup_pkt(flow_stat_payload_header *fsp_head) {
 
 #endif /* LATENCY_IEEE_1588_TIMESTAMPING */
 
-void RXLatency::handle_pkt(const rte_mbuf_t *m, int port) {
-  uint8_t tmp_buf[sizeof(struct flow_stat_payload_header)];
-  CFlowStatParser parser(CFlowStatParser::FLOW_STAT_PARSER_MODE_SW);
-  parser.set_vxlan_skip(CGlobalInfo::m_options.m_ip_cfg[port].get_vxlan_fs());
-  int ret = parser.parse(rte_pktmbuf_mtod(m, uint8_t *), m->pkt_len);
+void RXLatency::handle_flow_latency_stats(const rte_mbuf_t *m, uint32_t& ip_id) {
+    uint8_t tmp_buf[sizeof(struct flow_stat_payload_header)];
+    struct flow_stat_payload_header *fsp_head =
+        (flow_stat_payload_header *)utl_rte_pktmbuf_get_last_bytes(
+            m, sizeof(struct flow_stat_payload_header), tmp_buf);
 
-  if (m_rcv_all || (ret == 0)) {
+    if (!is_flow_stat_payload_id(ip_id)) {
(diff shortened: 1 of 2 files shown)


hhaim:
see minor changes here
https://github.com/hhaim/trex-core/tree/bidtt4-update_RXLatency_handle_pkt

jsmoon:
@hhaim, better than my code.

hhaim:
rebased with minor changes

jsmoon:
@hhaim I'm sorry I missed one point. your change may not cover the following case.

<non-IP packet> --> DUT --> <IP packet>

In this case, the flow latency packet will not be detected properly.

hhaim:
@jsmoon who said that we need to support it? The original design supported **only** valid IP with ip.id. 
Then we added non-ip with meta-data only in **software mode**. The fact that the old code supported very strange options is not an indication that this is what we wanted to support.

jsmoon:
@hhaim, actually my user uses it. My company's custom packet is translated to the IP packet and vice versa.
So I added to check the latency magic first in software mode regardless of ip.id.

hhaim:
@jsmoon what if the ip.id in your case? it is a valid ip.id or not?

jsmoon:
@hhaim it's random. some packets have valid ip.id and some packets haven't.

hhaim:
@jsmoon it means that there should be a better way to identify that this packet is a latency packet. 
MAGIC of one byte (0xAB) is lame ..
better to change the header  magic and the validation in this case

jsmoon:
@hhaim I agree. what about adding a additional field to confirm the MAGIC byte is valid?
I think the last byte can has the checksum or another MAGIC field.
What do you think about it?

hhaim:
I would create another verification function that will have more strength (only on software mode) but keep the same size. 

1. check magic 
2. check that time_stamp is reasonable (diff is less than +-2 sec)
3. hw_id is in range
4. seq is in range (MAXIMUM DROP +=1000) 

```
    uint8_t magic;
    uint8_t flow_seq;
    uint16_t hw_id;
    uint32_t seq;
    uint64_t time_stamp;
```

jsmoon:
@hhaim 1. and 3. are already checked currently.
for 2. and 4. I'm not sure how to handle the first packet when there is no previous packet.
I would make the decision by the packet itself.
time_stamp could be compared with *the current time* for checking the packet is a valid latency packet.

egwakim:
Hi @hhaim
The balanced way(performance vs strong checking) would be checking existing fields as you suggested.
(Keep 16 bytes header for the cache and pipeline)
Checking existing 64 bits time_stamp(Valid if now()-5sec ~ now()+5sec) seems lightweight and strong enough to check the validity.

Limiting flow_latency feature to IP packet only is not an good idea, many users have been used flow_latency stats because of the custom packet feature. Since flow_latency add own custom header, it can distinguish flows whatever header(IP or non-IP) changed or not. It's very useful feature.

hhaim:
@egwakim agreed. Adding time_stamp ~4sec validation will add 32bit header comparison which is stronger than  ip.id (16 bit). 
I suggest to **add** this validation only in **software** mode.

Issue #734, opened 2021-06-21
ASTFProgram delay(10) has no delay effect.


jsmoon:
I know that the delay command should have greater or equal to 20.
When the value is less than 20, the delay command will be performed but have no time delay.
Rather, it would break the burst commands limit (1000) in `process_cmd()` and cause the watchdog crash.

```
WATCHDOG: task 'Trex DP core 1' has not responded for more than 1.00027 seconds - timeout is 1 seconds

*** traceback follows ***

1             0x5760bc ./_t-rex-64() [0x5760bc]
2       0x7f7f4fb6b8a0 /lib/x86_64-linux-gnu/libpthread.so.0(+0x128a0) [0x7f7f4fb6b8a0]
3             0x86b28a rte_delay_us_block + 74
4             0x512edb CCoreEthIF::send_burst(CCorePerPort*, unsigned short, CVirtualIFPerSideStats*) + 155
5             0x513dfb CCoreEthIFTcp::send_node(CGenNode*) + 347
6             0x536a75 CTcpIOCb::on_tx(CTcpPerThreadCtx*, tcpcb*, rte_mbuf*) + 37
7             0x52f978 CEmulApp::on_tick() + 168
8             0x51968c ./_t-rex-64() [0x51968c]
9             0x5c9fb5 CNATimerWheel::on_tick_level0(void*, void (*)(void*, CHTimerObj*)) + 117
10            0x5ca6bd CNATimerWheel::on_tick_level(void*, void (*)(void*, CHTimerObj*), unsigned short) + 29
11            0x51b95a CTcpPerThreadCtx::timer_w_on_tick() + 26
12            0x5395e8 CFlowGenListPerThread::handle_tw(CGenNode*, bool) + 136
13            0x55c1cf CNodeGenerator::handle_slow_messages(unsigned char, CGenNode*, CFlowGenListPerThread*, bool) + 399
14            0x511377 int CNodeGenerator::flush_file_realtime<24, false>(double, double, CFlowGenListPerThread*, double&) + 1127
15            0x767292 TrexAstfDpCore::start_scheduler() + 898
16            0x6b1bd9 TrexDpCore::start() + 89
17            0x5511e0 CFlowGenListPerThread::start(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, CPreviewMode&) + 112       
18            0x4f3ace CGlobalTRex::run_in_core(unsigned char) + 462
[...]

PR #735, opened 2021-06-22
prevent crash from ASTF zero delay request

jsmoon:
Hi, this change is for issue #734.

From this code change, delay(<20) behavior will be changed.
With this change, the `delay_rand(10, 50)` in http_manual_commands_pipline4.py makes a different result from golden pcap. 
So, I updated the golden pcap file also.


hhaim:
rebased

PR #736, opened 2021-06-22
try flow latency with magic in software mode

jsmoon:
Hi, this change is for the discussion result at #731. 

In this change, in software mode, the flow latency will be checked at first by magic and timestamp in the payload.
If the magic matched and the timestamp has an acceptable value, from 5 seconds before to now, it is treated as a flow latency packet regardless of IP.id value.
Since the timestamp could not exceed the current time, the `~now+5` case is ignored.

@hhaim please review my changes and give your feedback.

### DIFF
--- a/src/stx/common/trex_latency_counters.cpp
+++ b/src/stx/common/trex_latency_counters.cpp
@@ -234,20 +234,24 @@ RXLatency::update_timestamps_for_fup_pkt(flow_stat_payload_header *fsp_head) {
 
 #endif /* LATENCY_IEEE_1588_TIMESTAMPING */
 
+/* check if given time is reasonable, from 5 seconds before to now. */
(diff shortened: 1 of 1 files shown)


hhaim:
@jsmoon consider this 
https://github.com/hhaim/trex-core/tree/latency_check

jsmoon:
@hhaim it's a good idea adding `is_vaild_ts` method to `flow_stat_payload_header`.
I agree to merge your code instead of this change.

hhaim:
@jsmoon I didn't check the change, i don't know if it compile. Please verify that it works

jsmoon:
@hhaim I did compile your change and STL regression in my simple environment. It seems O.K.

hhaim:
@jsmoon we don't see this issue on our setups maybe it is related to PF driver (linux) or firmware

hhaim:
@jsmoon ok, I will merge it to master

PR #739, opened 2021-06-29
prevent watchdog from lots of non-blocking send

jsmoon:
Hi, recently I've got a watchdog case at `tcp_socket.cpp:1097`.
```
1093         /* with new vector, calculate the sum again, could be size of 1 */
1094         for (i=0;i<(int)z; i++ ) {
1095             CMbufBuffer * b=m_q[i];
1096             uint32_t c_len =b->len();
1097             sum+=c_len;                       // <-- watchdog here
1098         }
```
This issue happened after #717 merged when my user sent tera-bytes in non-blocking mode. In this case, the number of buffers(`z`) could be about millions in there.

So, I removed the loop getting the total number of bytes in buffers by adding `m_q_tot_bytes`.
Furthermore, I made non-blocking send working only when there is a space in socket buffer using by `m_api->get_tx_sbspace(m_flow)`.
I think this change prevents the large number of buffers in `m_q`.

@hhaim could you check my changes and give your feedback?

### DIFF
--- a/src/44bsd/tcp_socket.cpp
+++ b/src/44bsd/tcp_socket.cpp
@@ -384,7 +384,7 @@ CEmulAppCmd* CEmulApp::process_cmd_one(CEmulAppCmd * cmd){
                 m_api->tx_tcp_output(m_pctx,m_flow);
             }
 
-            if (get_tx_mode_none_blocking()) {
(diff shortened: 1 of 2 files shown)


hhaim:
@jsmoon looks good. Usually there is one mbuf in the tx queue.

hhaim:
rebased

PR #740, opened 2021-06-29
Additional GCC compilation error fixes

mykaul:
1. Similar to 0bd600860fd504290c7c87adbf3044b970ed4b6d - clearing an object of non-trivial type
I'm less happy about this fix, as I'm using both the clear() method AND specific setters.
I wish I could just use clear()

2. Fixing this error:
```
../../src/trex_watchdog.cpp: In function ‘std::string exec(const char*)’:
../../src/trex_watchdog.cpp:52:10: error: ‘shared_ptr’ is not a member of ‘std’
   52 |     std::shared_ptr<FILE> pipe(popen(cmd, "r"), pclose);
      |          ^~~~~~~~~~
../../src/trex_watchdog.cpp:38:1: note: ‘std::shared_ptr’ is defined in header ‘<memory>’; did you forget to ‘#include <memory>’?
   37 | #include <iostream>
  +++ |+#include <memory>
   38 | #include  <stdexcept>
../../src/trex_watchdog.cpp:52:25: error: expected primary-expression before ‘>’ token
   52 |     std::shared_ptr<FILE> pipe(popen(cmd, "r"), pclose);
      |                         ^
../../src/trex_watchdog.cpp:52:37: error: cannot convert ‘FILE*’ to ‘int*’
   52 |     std::shared_ptr<FILE> pipe(popen(cmd, "r"), pclose);
      |                                ~~~~~^~~~~~~~~~
      |                                     |
      |                                     FILE*
```
Note: there are other compilation errors with GCC (I'm using 11.1.1), squashing them slowly.

### DIFF
--- a/src/stx/stl/trex_stl_fs.cpp
+++ b/src/stx/stl/trex_stl_fs.cpp
@@ -112,11 +112,7 @@ void dbg_printf(const char *fmt, ...)
 /************** class CFlowStatUserIdInfo ***************/
 CFlowStatUserIdInfo::CFlowStatUserIdInfo(uint16_t l3_proto, uint8_t l4_proto, uint8_t ipv6_next_h) {
     m_rx_cntr->clear();
-    m_rx_cntr->set_byte_base(0);
-    m_rx_cntr->set_pkt_base(0);
     m_tx_cntr->clear();
-    m_tx_cntr->set_byte_base(0);
-    m_tx_cntr->set_pkt_base(0);
     m_hw_id = HW_ID_INIT;
     m_l3_proto = l3_proto;
     m_l4_proto = l4_proto;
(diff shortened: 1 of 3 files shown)


hhaim:
@mykaul why not just fix the function ```clear()```?
and keep only ```m_tx_cntr[i].clear();``` why do you need to add 

```
 m_tx_cntr[i].set_byte_base(0);        m_tx_cntr[i].set_pkt_base(0);```

mykaul:
> @mykaul why not just fix the function `clear()`?
> and keep only `m_tx_cntr[i].clear();` why do you need to add
> 
> ```
>  m_tx_cntr[i].set_byte_base(0);        m_tx_cntr[i].set_pkt_base(0);```
> ```

Excellent question:
1. I suspected there was a reason they were not added to clear()
2. I did not know if there are any users that rely on those counters NOT being cleared.

If you are OK with it, I'll certainly push another commit which improves clear() and then uses it.

(btw, for consistency, we perhaps need to call it reset() - unsure)

hhaim:
rebased

PR #743, opened 2021-07-08
reduce get_async_events requests

jsmoon:
Hi, this change will reduce the `get_async_events` requests.

When there are lots of profiles are running, lots of state change events would be published simultaneously.
Since the `self.socket.recv()` get the event messages one by one, `get_async_events` requests by the following event messages would get a vacant response.
My change will send it once for all the event messages received at the same time.

In addition, I add a code to ignore other messages (trex-global, ...) and fix a minor bug derived from this change.

### DIFF
--- a/scripts/automation/trex_control_plane/interactive/trex/common/trex_subscriber.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/common/trex_subscriber.py
@@ -294,13 +294,21 @@ class TRexSubscriber():
             self.socket.close(linger = 0)
         
     def _wait_for_trigger (self):
-            try:
-               self.socket.recv() # block on event
-               return True 
-            except zmq.Again:
-               return False
-            except zmq.ContextTerminated:
-               return False
+        is_event = False
+        try:
+            while not is_event:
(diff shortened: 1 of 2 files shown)


hhaim:
@jsmoon not sure this is the best way to achieve your need. 
Adding a sleep (of 10-20msec) after the execution of the task will merge events and will limit the number of events in 1 sec to a certain limit. 
in your case you don't have a limit, if there is a temporary empty queue you will have a new event. 

```
        while self.t_state != self.THREAD_STATE_DEAD:
            res = self._wait_for_trigger()
             ...
                self.last_data_recv_ts = time.time()

                data = rc.data()['data']
                for msg in data:
                        self.monitor.on_recv_msg(msg)
                        name     = msg['name']
                        data     = msg['data']
                        msg_type = msg['type']
                        self.__dispatch(name, msg_type, data, baseline)
                self.async_data   = True

            time.sleep(0.05)  <<<<<<<<<<<<<<<<<<<<< sleep to throttle the number of events 

```

hhaim:
rebased

PR #761, opened 2021-09-22
fix 2 potential buffer overflows in DpdkTRexPortAttr

ohilyard:
Both rte_eth_xstats_get and rte_eth_xstats_get_names do not include
xstats in their return values if the provided buffer is NULL. This
resulted in a case where if the number of xstats is greater than 0,
both DpdkTRexPortAttr::get_xstats_values and
DpdkTRexPortAttr::get_xstats_names would attempt to write past
the end of the vector provided to hold the return value. This was
observed to cause segfaults on both Intel and Broadcom NICs.

It may be possible to remove the do-while loop by passing in a valid
buffer in place of the NULL in both of these methods. However, that may
be less reliable since the do-while loop will continue attempting to get
the information until either it is successful or an error code is
returned.

Signed-off-by: Owen Hilyard <ohilyard@iol.unh.edu>

### DIFF
--- a/src/dpdk_trex_port_attr.cpp
+++ b/src/dpdk_trex_port_attr.cpp
@@ -112,9 +112,13 @@ int DpdkTRexPortAttr::get_xstats_values(xstats_values_t &xstats_values) {
     if (size < 0) {
         return size;
     }
-    xstats_values_tmp.resize(size);
-    xstats_values.resize(size);
-    size = rte_eth_xstats_get(m_repid, xstats_values_tmp.data(), size);
+
+    do {
+        xstats_values_tmp.resize(size);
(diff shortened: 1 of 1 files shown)


ohilyard:
> Checked SHA: [e9ccacb](https://github.com/cisco-system-traffic-generator/trex-core/commit/e9ccacbd8a9bb228f11f57b997a7d3a4d328ce31)
> Status: **FAILURE** 
> Link to job: http://81.218.86.50:8080/job/trex_build/603/

This looks to me like a connection failure in CI. A python script failed during what looks like the setup phase for an RPC server with a socket timeout.

ybrustin:
recheck

PR #769, opened 2021-10-20
configure --new-memory works w/o -o option

jsmoon:
Hi, this PR will fix the issue --new-memory does not work without -o option.

```
cd linux_dpdk/
./b configure --new-memory
./b build
```
This has been failed with the following message.
```
In file included from <command-line>:0:0:
../../src/../src/pal/linux_dpdk/dpdk_2102_x86_64/rte_config.h:164:1: fatal error: dpdk_config.h: No such file or directory
#endif /* _RTE_CONFIG_H_ */
^
compilation terminated.
```

The reason is that the implicit `dpdk_config.h` file is created in the current directory without `-o` option.
My change will fix this issue by creating the file in the `out` directory.


hhaim:
rebased

Issue #741, opened 2021-06-29
GCC compilation failure in member function ‘bool CFlowStatRuleMgr::dump_json taking address of rvalue


mykaul:
Using GCC 11.1:
```
[130/342] Compiling ../src/trex_platform.cpp
../../src/stx/stl/trex_stl_fs.cpp: In member function ‘bool CFlowStatRuleMgr::dump_json(Json::Value&, std::vector<unsigned int>)’:
../../src/stx/stl/trex_stl_fs.cpp:1315:93: error: taking address of rvalue [-fpermissive]
 1315 |             std::string str_port = static_cast<std::ostringstream*>( &(std::ostringstream() << int(port) ) )->str();
      |                                                                       ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
../../src/stx/stl/trex_stl_fs.cpp:1358:92: error: taking address of rvalue [-fpermissive]
 1358 |         std::string str_user_id = static_cast<std::ostringstream*>( &(std::ostringstream() << user_id) )->str();
      |                                                                      ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
../../src/stx/stl/trex_stl_fs.cpp:1362:93: error: taking address of rvalue [-fpermissive]
 1362 |             std::string str_port = static_cast<std::ostringstream*>( &(std::ostringstream() << int(port) ) )->str();
      |                                                                       ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~

```

mykaul:
Another very similar issue:
```
../../src/time_histogram.cpp: In member function ‘void CTimeHistogram::dump_json(Json::Value&, bool)’:
../../src/time_histogram.cpp:289:75: error: taking address of rvalue [-fpermissive]
  288 |                     std::string key = static_cast<std::ostringstream*>( &(std::ostringstream()
      |                                                                          ~~~~~~~~~~~~~~~~~~~~~
  289 |                                                                           << int(base * (i + 1)) ) )->str();
      |                                                                           ^~~~~~~~~~~~~~~~~~~~~~~~

```

bdollma:
Closing because the fix was rebased.

mykaul:
Thanks - confirming it compiles

> Closing because the fix was rebased.

Confirming it almost works with: gcc (GCC) 11.2.1 20211203 (Red Hat 11.2.1-7) - the above issue is solved, but it failed with:
```
[...]

PR #770, opened 2021-11-10
Fix string casting issue on gcc11

Maickii:
gcc11 does not allow 'taking address of rvalue' like:
&(std::ostringstream())

This leads to compilation errors on gcc11. Fix this issue by correctly
casting to string using std::to_string() on the offending lines

Signed-off-by: Michael Santana <msantana@redhat.com>


Maickii:
This fixes the issue reported [here](https://github.com/cisco-system-traffic-generator/trex-core/issues/741)

hhaim:
reabsed

PR #772, opened 2021-11-12
mlx5: fix timestamp

viacheslavo:
This patch fixes ConnectX-6DX probing with newest firmware and REAL_TIME_CLOCK_ENABLE=1 configured


hhaim:
rebased

PR #782, opened 2021-12-07
Bnxt xstats fix

hlrichardson:
This PR has two fixes for issues in the bnxt PMD xstats APIs that are currently causing memory corruption (overrun) issues in TRex.


hhaim:
rebased

PR #786, opened 2022-01-17
Improve fetch_capture_packets to reduce RPC messages

yonghwan007:
Hi, this PR is to improve fetch_capture_packets operation to reduce RPC messages.

**1. Background**
 Pakcet capture performance can be degraded by too many RPC messages and large packet binary size in RPC response.
 Improvement point is as below.
  - Reduce the number of RPC messages for fetching packets.
  - Reduce the amount of packet binary in RPC response

**2. Code changes**
(1) client side.
 Added new parameter in 'fetch_capture_packets'.
```
  - fetch_limit: limit the number of packets to be fetched with one RPC _transmit() call. Default is 50 for backward compatibility.
  - snaplen: snapshot length is the amount of data for each packet that is actually fetched. If snaplen is 0, fetch full packet without snapshot.
```
(2) server side.
 Packet 'binary' is cut by snaplen.
 Added new 'wirelen' field in RPC command response.
```
 'binary’  - binary bytes of the snapped packet
 ‘wirelen’ - the length of packet on the wire
```

**3. Backward compatibility**
It is fully backward compatible in new_client+old_server and old_client+new_server.

@hhaim, please review my change.

### DIFF
--- a/src/stx/common/trex_rpc_cmds_common.cpp
+++ b/src/stx/common/trex_rpc_cmds_common.cpp
@@ -1541,7 +1541,8 @@ TrexRpcCmdCapture::parse_cmd_fetch(const Json::Value &params, Json::Value &resul
     
     uint32_t capture_id = parse_uint32(params, "capture_id", result);
     uint32_t pkt_limit  = parse_uint32(params, "pkt_limit", result);
-    
+    uint16_t snaplen    = parse_uint16(params, "snaplen", result, 0);
+
     /* generate a fetch command */
     
     static MsgReply<TrexCaptureRCFetch> reply;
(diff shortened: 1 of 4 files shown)


src/stx/common/trex_rpc_cmds_common.cpp
```diff
@@ -1541,7 +1541,8 @@ TrexRpcCmdCapture::parse_cmd_fetch(const Json::Value &params, Json::Value &resul
     
     uint32_t capture_id = parse_uint32(params, "capture_id", result);
     uint32_t pkt_limit  = parse_uint32(params, "pkt_limit", result);
-    
+    uint16_t snaplen  = params.isMember("snaplen") ? parse_uint16(params, "snaplen", result) : 0;
```
bdollma, line 1544: @yonghwan007 
I suggest you change this with 
```
uint16_t snaplen = parse_uint16(params, "snaplen", result, 0);
```
The `parse_uint16` function supports default values.

egwakim:
It's new useful snaplen option like tcpdump(or wireshark) to eliminate unused data payload when capturing packets.

src/stx/common/trex_rpc_cmds_common.cpp
yonghwan007: I updated code as your comment.

hhaim:
@yonghwan007 looks good.

hhaim:
rebased

Issue #792, opened 2022-01-24
gcc compilation failure: ../../src/common/gtest-all.cc:6962:31: error: ‘dummy’ may be used uninitialized [-Werror=maybe-uninitialized]


mykaul:
[118/346] Compiling ../src/tunnels/gtp_man.cpp
../../src/common/gtest-all.cc: In function ‘bool testing::internal::StackGrowsDown()’:
../../src/common/gtest-all.cc:6962:31: error: ‘dummy’ may be used uninitialized [-Werror=maybe-uninitialized]
 6962 |   return StackLowerThanAddress(&dummy);
      |          ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
../../src/common/gtest-all.cc:6955:6: note: by argument 1 of type ‘const void*’ to ‘bool testing::internal::StackLowerThanAddress(const void*)’ declared here
 6955 | bool StackLowerThanAddress(const void* ptr) {
      |      ^~~~~~~~~~~~~~~~~~~~~
../../src/common/gtest-all.cc:6961:7: note: ‘dummy’ declared here
 6961 |   int dummy;
      |       ^~~~~
cc1plus: all warnings being treated as errors

Waf: Leaving directory `/home/tlv/ykaul/github/trex-core/linux/build'
Build failed
 -> task in 'bp-sim-64-debug' failed with exit status 1 (run with -v to display more information)
(which can be easily fixed by initializing dummy to 0, as can be seen in https://github.com/google/googletest/blob/1b18723e874b256c1e39378c6774a90701d70f7a/googletest/src/gtest-death-test.cc#L1286 )

It's unclear though if we should not just update to a newer gtest framework though?

PR #793, opened 2022-01-24
Fix gcc compilation failure: error: ‘dummy’ may be used uninitialized

mykaul:
Fixes the below:
[118/346] Compiling ../src/tunnels/gtp_man.cpp
../../src/common/gtest-all.cc: In function ‘bool testing::internal::StackGrowsDown()’:
../../src/common/gtest-all.cc:6962:31: error: ‘dummy’ may be used uninitialized [-Werror=maybe-uninitialized]
6962 | return StackLowerThanAddress(&dummy);
| ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
../../src/common/gtest-all.cc:6955:6: note: by argument 1 of type ‘const void*’ to ‘bool testing::internal::StackLowerThanAddress(const void*)’ declared here
6955 | bool StackLowerThanAddress(const void* ptr) {
| ^~~~~~~~~~~~~~~~~~~~~
../../src/common/gtest-all.cc:6961:7: note: ‘dummy’ declared here
6961 | int dummy;
| ^~~~~

Fixes: #792

### DIFF
--- a/src/common/gtest-all.cc
+++ b/src/common/gtest-all.cc
@@ -6953,12 +6953,12 @@ static int ExecDeathTestChildMain(void* child_arg) {
 // function, but we want to guard against the unlikely possibility of
 // a smart compiler optimizing the recursion away.
 bool StackLowerThanAddress(const void* ptr) {
-  int dummy;
+  int dummy = 0;
   return &dummy < ptr;
 }
 
 bool StackGrowsDown() {
-  int dummy;
+  int dummy = 0;
   return StackLowerThanAddress(&dummy);
 }
 


src/common/gtest-all.cc
```diff
@@ -6958,7 +6958,7 @@ bool StackLowerThanAddress(const void* ptr) {
 }
 
 bool StackGrowsDown() {
-  int dummy;
+  int dummy = 0;
   return StackLowerThanAddress(&dummy);
```
bdollma, line 6962: In `StackLowerThanAddress` the same problem occurs: https://github.com/cisco-system-traffic-generator/trex-core/blob/5147003a7fc6cc5859c06d9cf86e6896197d4fb7/src/common/gtest-all.cc#L6956
Maybe this compiler is smart enough to understand that in `StackLowerThanAddress` the value is not in use, but it doesn't understand that for `StackGrowsDown`. Anyway, I suggest you fix that too.

bdollma: Also, since I can't comment on it, why did you make changes on `scripts/so/x86_64/libstdc++.so.6`?

mykaul: I certainly did not mean to :(
Let me re-send!

Issue #759, opened 2021-09-09
[Improvement] Uplift TCP/IP stack in TRex


egwakim:
Hi,
Current TCP stack in TRex is quite old(4.4 BSD-Lite), it’s not fit to simulate modern TCP traffic behavior.
To simulate modern TCP traffic behavior, it would be good to uplift TCP stack. 
Following features would be good for new TCP stack.

- Uplift TCP stack to FreeBSD 13.0.0
- All features and APIs shall be backward compatible with legacy TCP stack
- Supports new TCP congestion control algorithms : NewReno(FreeBSD default), CUBIC (Most popular CC)
- In the future, supporting BBRv1 and BBRv2 would be better

Let's discuss the detailed designs in this issue thread.

egwakim:
Hi @hhaim, 
Regression test for new TCP stack almost passed now.
It's time to start discussion before push.
Here is branch for it.
https://github.com/bidtt4/trex-core/commits/freebsd_int

There are several topics to be discussed before push,
Let's discuss topics one by one.

First topic is 'timer' design,
I want to get your opinion about it.
We need to decide policy whether we should use defaults from new TCP stack or use old TCP stack's default values.
I believe that the purpose of using TRex is simulating real-world traffic to verify products.
I think it would be better to keep the new TCP stack default if no critical issues.
How about your opinion about following issues?

![image](https://user-images.githubusercontent.com/4927607/143422789-0c725199-832b-4ca9-ab1f-65e55d630240.png)

hhaim:
@egwakim @jsmoon 
I looked into the code (didn't compile and test it yet) but it looks promising. 
I saw that the ws_script still has the old TCP files, so I don't understand how the new one are chosen 
in regards to the timers: 
I don't care to set the default to 20 msec, we just need to have a way to set all the timer to a longer value (tuable) to support M of flows (we have use-case of 40M active flows)

Did you compare performance with bare -metal setup?

jsmoon:
@hhaim
All the new stack codes reside in `src/44bsd/netinet/` folder.
I added a `-DTREX_FBSD` option for building the new TCP stack. The new TCP stack will be built as a static library.
I declared the library interface in the header.
In `src/44bsd/netinet/integration-note.md`, the interface functions and structures are explained shortly.
The old TCP stack files should keep legacy code for the new stack integration.
- tcp_input.cpp: TCP reassembly
- tcp_output.cpp: build TCP packet
[...]

PR #796, opened 2022-01-27
Uplift TCP/IP stack (FreeBSD 13 based)

jsmoon:
This PR is related to #759.

This change replaces the old 4.4BSD TCP/IP stack with the new FreeBSD 13 TCP/IP stack.
The new stack supports SACK and the new CC algorithm(newreno and CUBIC).

@hhaim please give your feedback on this change.


hhaim:
@jsmoon 

It is still under review, I running full regression. In general it looks very good! 


**few comments**: 

1) you are writing data into the **mbuf packets** which is share data with PCI/other cores and could create false sharing
instead it is better to make a copy in the stack and convert it 

```
static inline void
tcp_fields_to_host(struct tcphdr *th)
{
	th->th_seq = ntohl(th->th_seq);
	th->th_ack = ntohl(th->th_ack);
	th->th_win = ntohs(th->th_win);
	th->th_urp = ntohs(th->th_urp);
}

instead:
	tcp_fields_to_host(th);
```

```
struct tcphdr *th_pkt

// in the stack -- fast in the cache 
struct tcphdr th

convert_to_host(th,th_pkt){
    th->flags =th_pkt->flags 
	th->th_seq = ntohl(th_pkt->th_seq);
	th->th_ack = ntohl(th_pkt->th_ack);
	th->th_win = ntohs(th_pkt->th_win);
	th->th_urp = ntohs(th_pkt->th_urp);
    .. all the fields
}
   ```



2) 
	you don't have this sanity check to drop the packet !, it could create a crash 

```
    if (l7off + ti->ti_len > m->pkt_len )){
        /* somthing wrong here drop the packet */
        goto drop;
    }


```


3) our regression fail with gcc 7 I need to do this

```
diff --git a/linux_dpdk/ws_main.py b/linux_dpdk/ws_main.py
index 13de4f296..864dd92d8 100755
--- a/linux_dpdk/ws_main.py
+++ b/linux_dpdk/ws_main.py
@@ -2510,11 +2510,13 @@ def build_prog (bld, build_obj):
               source   = bpf.file_list(top),
               target   = build_obj.get_bpf_target())
 
-    bld.stlib(features = 'c',
-              includes = tcp_includes_path,
-              cflags   = cflags,
-              source   = tcp_src.file_list(top),
-              target   = build_obj.get_tcp_target())
+    bld.objects(
+      features='c ',
+      includes = tcp_includes_path,
+      cflags   = cflags,
+      source   = tcp_src.file_list(top),
+      target=build_obj.get_tcp_target()
+      )
 
     inc_path = dpdk_includes_path + includes_path
     cxxflags_ext = ['',]
diff --git a/src/44bsd/netinet/tcp_timer.c b/src/44bsd/netinet/tcp_timer.c
index e3fa9ce79..86138ea1b 100644
--- a/src/44bsd/netinet/tcp_timer.c
+++ b/src/44bsd/netinet/tcp_timer.c
@@ -340,8 +340,8 @@ static void
 _handle_slow_timers(struct tcpcb *tp, uint32_t tick_passed)
 {
        uint32_t tt_flags = tp->m_timer.tt_flags & ((1 << TCPT_NTIMERS) - 1);
-
-       for (int i = TT_REXMT; tt_flags && i < TCPT_NTIMERS; i++ ) {
+       int i;
+       for (i = TT_REXMT; tt_flags && i < TCPT_NTIMERS; i++ ) {
                if ((tt_flags & (1 << i)) == 0)
                        continue;
                tt_flags &= ~(1 << i);

```

4) quick test with eflow wth one core should 15gbps vs 23gbps in the original version

jsmoon:
@hhaim, thank you for your review.
1. I copied it by TCP_MAXHLEN(=60) for tcp_dooptions also.
2. I added it. But I think it has been done already in the `CFlowTable::parse_packet` with counting `m_err_len_err`.
3. I changed it as you suggested.
4. Could you share your command? I cannot make it.

hhaim:
@jsmoon 

Almost there,

Looking more into the code and comparing it to the old stack, it shows much more logic so I understand why there is more cpu cycles now. 
The code integration is very clean, nice work!

* in reagards to   'CFlowTable:: parse_packet with counting m_err_len_err', it  checks something else. it checks aginst the ipv4.len()  and not aginst the real mbuf size and mbuf_len == ipv4.len()+tcp.headerlen()+payload. Maybe we could add this check into the parser but it is not there today 

there is a need to convert all the tabs to spaces in the new tcp stack files, let me know if you me to do this

jsmoon:
@hhaim, yes please do it.

PR #801, opened 2022-02-16
TCP receive offload packets by software

jsmoon:
This PR is to support TCP LRO per flow by software.

It is activated when the NIC does not support LRO and the `--lro-disable` is not used.
Only the packets in the same flow within 1 milli-second will be merged.

### DIFF
--- a/src/44bsd/flow_table.cpp
+++ b/src/44bsd/flow_table.cpp
@@ -86,11 +86,22 @@ bool CFlowTable::Create(uint32_t size,
 
     m_service_status        = SERVICE_OFF;
     m_service_filtered_mask = 0;
+
+    m_tcp_lro = nullptr;
+    if (!CGlobalInfo::m_options.preview.get_dev_lro_support() &&
+        !CGlobalInfo::m_options.preview.getLroOffloadDisable()) {
+        m_tcp_lro = new CTcpRxOffload;
+        m_tcp_lro->Create();
+    }
     return(true);
 }
 
 void CFlowTable::Delete(){
     m_ft.Delete();
+    if (m_tcp_lro) {
+        m_tcp_lro->Delete();
+        delete m_tcp_lro;
+    }
 }
 
 
@@ -394,6 +405,15 @@ bool CFlowTable::update_new_template(CTcpPerThreadCtx * ctx,
     return true;
 }
 
+static inline void tcp_flow_input(CTcpFlow *  flow,
+                                  struct rte_mbuf * mbuf,
+                                  TCPHeader    * lpTcp,
+                                  CFlowKeyFullTuple &ftuple){
+
+    tcp_int_input(&flow->m_tcp, (struct mbuf*)mbuf, (struct tcphdr*)lpTcp, ftuple.m_l7_offset, ftuple.m_l7_total_len, 0);
+
+    flow->check_defer_function();
+}
 
 void HOT_FUNC CFlowTable::process_tcp_packet(CTcpPerThreadCtx * ctx,
                                     CTcpFlow *  flow,
@@ -403,11 +423,14 @@ void HOT_FUNC CFlowTable::process_tcp_packet(CTcpPerThreadCtx * ctx,
     TCPHeader tcphdr;
     if (unlikely(!flow->is_activated())) {
         tcphdr = *lpTcp;
+    } else {
+#ifndef TREX_SIM
+        if (process_software_lro(ctx, flow, mbuf, lpTcp, ftuple))
+            return;
+#endif
     }
 
-    tcp_int_input(&flow->m_tcp, (struct mbuf*)mbuf, (struct tcphdr*)lpTcp, ftuple.m_l7_offset, ftuple.m_l7_total_len, 0);
-
-    flow->check_defer_function();
+    tcp_flow_input(flow, mbuf, lpTcp, ftuple);
 
     if (unlikely(flow->new_template_identified())) {
         /* identifying new template has been done */
@@ -932,6 +955,180 @@ bool CFlowTable::rx_handle_packet_tcp_no_flow(CTcpPerThreadCtx * ctx,
     return(true);
 }
 
+
+void CTcpRxOffloadBuf::set(CTcpFlow* flow, hr_time_t start_time) {
+    m_flow = flow;
+    flow->m_lro_buf = this;
+    m_start_time = start_time;
+}
+
+void CTcpRxOffloadBuf::do_clear() {
+    if (m_flow) {
+        if (m_mbuf) {
+            rte_pktmbuf_free(m_mbuf);
+            m_mbuf = nullptr;
+        }
+        m_flow = nullptr;
+    }
+}
+
+void CTcpRxOffloadBuf::update_mbuf(struct rte_mbuf* mbuf, TCPHeader* lpTcp, CFlowKeyFullTuple& ftuple) {
+    m_mbuf = mbuf;
+    m_lpTcp = lpTcp;
+    m_ftuple = ftuple;
+}
+
+bool CTcpRxOffloadBuf::reassemble_mbuf(struct rte_mbuf * mbuf, TCPHeader* lpTcp, CFlowKeyFullTuple& ftuple) {
+#ifndef TREX_SIM
+    /* merge if the two packets are neighbors */
+    if (m_ftuple.m_l7_total_len + ftuple.m_l7_total_len <= TCP_TSO_MAX_DEFAULT &&
+        /* check if TCP option fields are equal */
+        lpTcp->getHeaderLength() == m_lpTcp->getHeaderLength() &&
+        !memcmp(lpTcp+1, m_lpTcp+1, lpTcp->getHeaderLength() - sizeof(*lpTcp))) {
+
+        if (lpTcp->getSeqNumber() == m_lpTcp->getSeqNumber() + m_ftuple.m_l7_total_len) {
+            m_ftuple.m_l7_total_len += ftuple.m_l7_total_len;
+            rte_pktmbuf_adj(mbuf, ftuple.m_l7_offset);
+            rte_pktmbuf_chain(m_mbuf, mbuf);
+            INC_STAT(m_flow->m_pctx, m_flow->m_tg_id, tcps_rcvoffloads);
+            return true;
+        }
+        if (m_lpTcp->getSeqNumber() == lpTcp->getSeqNumber() + ftuple.m_l7_total_len) {
+            ftuple.m_l7_total_len += m_ftuple.m_l7_total_len;
+            rte_pktmbuf_adj(m_mbuf, m_ftuple.m_l7_offset);
+            rte_pktmbuf_chain(mbuf, m_mbuf);
+
+            update_mbuf(mbuf, lpTcp, ftuple);
+            INC_STAT(m_flow->m_pctx, m_flow->m_tg_id, tcps_rcvoffloads);
+            return true;
+        }
+    }
+#endif
+    return false;
+}
+
+void CTcpRxOffloadBuf::do_flush() {
+    m_flow->m_lro_buf = nullptr;
+    tcp_flow_input(m_flow, m_mbuf, m_lpTcp, m_ftuple);
+    m_mbuf = nullptr;
+    m_flow = nullptr;
+}
+
+
+void CTcpRxOffload::Create() {
+    m_size = NUM_RX_OFFLOAD;
+    m_bufs = new CTcpRxOffloadBuf[m_size];
+    m_head = 0;
+    m_tail = 0;
+}
+
+void CTcpRxOffload::Delete() {
+    for (; m_tail < m_head; ++m_tail) {
+        m_bufs[m_tail % m_size].do_clear();
+    }
+    delete[] m_bufs;
+}
+
+CTcpRxOffloadBuf* CTcpRxOffload::get_available_buf() {
+
+    assert(m_head >= m_tail);
+    if (m_head > m_tail) {
+        uint32_t index = (m_head - 1) % m_size;
+        if (!m_bufs[index].is_active()) {
+            return &m_bufs[index];
+        }
+    }
+    if (m_head - m_tail >= m_size) {
+        if (m_bufs[m_tail % m_size].is_active()) {
+            return nullptr;
+        }
+        ++m_tail;
+    }
+
+    uint32_t index = m_head % m_size;
+    ++m_head;
+    if (m_head > 2*m_size) {
+        m_head -= m_size;
+        m_tail -= m_size;
+    }
+    return &m_bufs[index];
+}
+
+CTcpRxOffloadBuf* CTcpRxOffload::find_buf(CTcpFlow* flow) {
+    return flow->m_lro_buf;
+}
+
+bool CTcpRxOffload::append_buf(CTcpFlow* flow,
+                               struct rte_mbuf* mbuf,
+                               TCPHeader* lpTcp,
+                               CFlowKeyFullTuple& ftuple) {
+
+    bool do_lro = (lpTcp->getFlags() == TCPHeader::Flag::ACK) && (ftuple.m_l7_total_len > 0);
+    CTcpRxOffloadBuf* buf = find_buf(flow);
+
+    if (buf) {
+        if (do_lro && buf->reassemble_mbuf(mbuf, lpTcp, ftuple)) {
+            return true;
+        }
+
+        buf->do_flush();
+        if (flow->is_can_close()) {
+            return false;
+        }
+    }
+
+    if (!do_lro) {
+        return false;
+    }
+
+    buf = get_available_buf();
+    if (buf) {
+        buf->set(flow, os_get_hr_tick_64());
+        buf->update_mbuf(mbuf, lpTcp, ftuple);
+        return true;
+    }
+    return false;
+}
+
+void CTcpRxOffload::flush_bufs(CTcpPerThreadCtx * ctx, uint32_t timeout_msec) {
+
+    assert(m_head >= m_tail);
+    hr_time_t current_time = os_get_hr_tick_64();
+    hr_time_t timeout_cycles = os_get_hr_freq()/1000 * timeout_msec;   // 1 msec
+
+    for (; m_tail < m_head; ++m_tail) {
+        CTcpRxOffloadBuf* buf = &m_bufs[m_tail % m_size];
+
+        if (buf->is_active()) {
+            if ((current_time - buf->get_start_time()) < timeout_cycles) {
+                break;
+            }
+
+            CTcpFlow* flow = buf->get_flow();
+
+            buf->do_flush();
+            if (flow->is_can_close()) {
+                ctx->m_ft.handle_close(ctx,flow,true);
+            }
+        }
+    }
+}
+
+
+HOT_FUNC bool CFlowTable::process_software_lro(CTcpPerThreadCtx * ctx,
+                                    CTcpFlow *  flow,
+                                    struct rte_mbuf * mbuf,
+                                    TCPHeader    * lpTcp,
+                                    CFlowKeyFullTuple &ftuple) {
+
+    return m_tcp_lro && m_tcp_lro->append_buf(flow, mbuf, lpTcp, ftuple);
+}
+
+HOT_FUNC void CFlowTable::flush_software_lro(CTcpPerThreadCtx * ctx) {
+    if (m_tcp_lro) {
+        m_tcp_lro->flush_bufs(ctx, 1);
+    }
+}
 
 HOT_FUNC bool CFlowTable::rx_handle_packet(CTcpPerThreadCtx * ctx,
                                   struct rte_mbuf * mbuf,
--- a/src/44bsd/flow_table.cpp
+++ b/src/44bsd/flow_table.cpp
@@ -95,7 +95,6 @@ bool CFlowTable::Create(uint32_t size,
 
         m_tcp_lro = new CTcpRxOffload;
         m_tcp_lro->Create(cb);
-        printf("TCP LRO is activated by software at %p\n", this);
     }
     return(true);
 }
--- a/src/44bsd/flow_table.cpp
+++ b/src/44bsd/flow_table.cpp
@@ -956,10 +956,10 @@ bool CFlowTable::rx_handle_packet_tcp_no_flow(CTcpPerThreadCtx * ctx,
 }
 
 
-void CTcpRxOffloadBuf::set(CTcpFlow* flow, hr_time_t start_time) {
+void CTcpRxOffloadBuf::set(CTcpFlow* flow) {
     m_flow = flow;
     flow->m_lro_buf = this;
-    m_start_time = start_time;
+    update_segsz();
 }
 
 void CTcpRxOffloadBuf::do_clear() {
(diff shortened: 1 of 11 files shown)


hhaim:
Nice work,

* Can you share numbers with the old and my proposal? what is the CPU% improvement and what are the  ``%`` of packets that could be merge

* There is one major problem! you can't save mbuf in the air, you must process and flush them immediately (after the burst loop),  as we don't have enough mbuf (in some cases 9K) to be in the driver 

in each iteration you can read up to 256 mbufs try to merge them and flush all after that 

```
        ctx->m_ft.flush_software_lro(ctx); << this function should no look into time, just flush all !
```

** I would reduce the size of 

```
#define NUM_RX_OFFLOAD  0x10000
==>
#define NUM_RX_OFFLOAD  256

```

and this:
```
            if (sum>NUM_RX_OFFLOAD) { << instead of 256
                ctx->m_ft.inc_rx_throttled_cnt();
                break;
            }
```

*
```
m_nsegs(struct mbuf *m)
{
    return ((struct rte_mbuf *)m)->nb_segs;
}
```

* m_nsegs is not 1 before this feature!. There are cards (almost all) that support scatter-gather and it provides number of segment (mbufs) for large packet
(e.g. for 9K long packet there are 4 2K mbufs). 


*
each of this function can return an error, need to handle it 
```
            rte_pktmbuf_adj(mbuf, ftuple.m_l7_offset);
            rte_pktmbuf_chain(m_mbuf, mbuf);
```


*
```
bool CTcpRxOffloadBuf::reassemble_mbuf(struct rte_mbuf * mbuf, TCPHeader* lpTcp, CFlowKeyFullTuple& ftuple) {
#ifndef TREX_SIM
    /* merge if the two packets are neighbors */
    if (m_ftuple.m_l7_total_len + ftuple.m_l7_total_len <= TCP_TSO_MAX_DEFAULT &&
        /* check if TCP option fields are equal */
        lpTcp->getHeaderLength() == m_lpTcp->getHeaderLength() &&


    if (m_ftuple.m_l7_total_len + ftuple.m_l7_total_len <= TCP_TSO_MAX_DEFAULT &&
        /* check if TCP option fields are equal */
        lpTcp->getHeaderLength() == m_lpTcp->getHeaderLength() &&
        !memcmp(lpTcp+1, m_lpTcp+1, lpTcp->getHeaderLength() - sizeof(*lpTcp))) {  option will never be the same ..
```

this is too strict, timestamp option could change and still it is possible to merge (take the older timestamp) but maybe we can use it because your timestamp is 1msec but this could change in the future. 

```
        !memcmp(lpTcp+1, m_lpTcp+1, lpTcp->getHeaderLength() - sizeof(*lpTcp))) {  << this is too strict
```

*


```
bool CTcpRxOffload::append_buf(CTcpFlow* flow,
                               struct rte_mbuf* mbuf,
                               TCPHeader* lpTcp,
                               CFlowKeyFullTuple& ftuple) {
```

why ```is_can_close``` should **not** be called inside, just save the packets in the ring  (in order) and flush them 

```
        if (flow->is_can_close()) {
            return false;
        }
```



the high level logic should be something like this:

```
CFlowTable::handle_tcp_mbuf(ctx, flow, mbuf, lpTcp, ftuple){

    tcp_flow_input(flow, mbuf, lpTcp, ftuple);
    if (unlikely(flow->new_template_identified())) {
        /* identifying new template has been done */
        if (!update_new_template(ctx, flow, &tcphdr, ftuple)) {
            return; /* when update is failed, flow is not available also. */
        }
    }

    if (flow->is_can_close()) {
        handle_close(ctx,flow,true);
    }

}

void HOT_FUNC CFlowTable::process_tcp_packet(CTcpPerThreadCtx * ctx,
                                    CTcpFlow *  flow,
                                    struct rte_mbuf * mbuf,
                                    TCPHeader    * lpTcp,
                                    CFlowKeyFullTuple &ftuple){
    
    if (process_software_lro(ctx, flow, mbuf, lpTcp, ftuple))
            return;
    }
    // context is free and this packet is not eligible (e.g. SYN)
    handle_tcp_mbuf(ctx, flow, mbuf, lpTcp, ftuple)
}

```

```
flush_software_lro(ctx) {
    loop_over_all_the_saved_mbufs in order and call {
        handle_tcp_mbuf(ctx, flow, mbuf, lpTcp, ftuple)
    }
}
```
for example  

```
[flow1-m1, flow1-m2, flow1-m4,  flow1-m5, flow1-m3]
```

flush_software_lro should process

```
flow1 [m1,m2] ,flow1-m4, flow1-5 flow1-m3

```

** 4 and 5 will not be merge because you gave only one ring in each flow, you can add more to solve this

hhaim:
@jsmoon 
in regard to the last bullet (```is_can_close()```)
better to add have a callback function  
to this (in the context of flowTable) a

```

    tcp_flow_input(flow, mbuf, lpTcp, ftuple);

    if (unlikely(flow->new_template_identified())) {
        /* identifying new template has been done */
        if (!update_new_template(ctx, flow, &tcphdr, ftuple)) {
            return; /* when update is failed, flow is not available also. */
        }
    }

    if (flow->is_can_close()) {
        handle_close(ctx,flow,true);
    }

```

so do_flush() will call this (add the pointer to flow_table)

jsmoon:
@hhaim, thanks for your review.

The received number of packets will be significantly decreased. This is the `http_simple.py` 1 flow result with TSO support.
```
             tcps_sndtotal |                 6 |                 4 | total packets sent         
             tcps_rcvtotal |                25 |                 6 | total packets received     
-->
             tcps_sndtotal |                 6 |                 4 | total packets sent         
             tcps_rcvtotal |                 5 |                 6 | total packets received     
          tcps_rcvoffloads |                20 |                 0 | receive offload packets by software 
```
The performance in CPU% will be better about 3~5%. (Now it is similar to the old stack result)

At first, I'll prepare the code changes for your following suggestions:
- flush immediately after the burst loop
- reduce the number of `CTcpRxOffloadBuf` and sync it to RX throttle count.
- don't use nb_segs for counting RX packets
- need to handle the error return of `rte_pktmbuf_adj` and `rte_pktmbuf_chain`
- remove comparison of options

> why is_can_close should not be called inside, just save the packets in the ring (in order) and flush them

I'm sorry I can not figure out what is your suggestion and why it is needed.
Could you give me more description of this suggestion?

Anyway, in CTcpRxOffload::append_buf(), do_flush() is called when given mbuf is not merged and it should be handled separately.
After do_flush(), there is no saved mbuf for the flow. So the given mbuf will be left for the flow.

For your example,  [flow1-m1, flow1-m2, flow1-m4,  flow1-m5, flow1-m3],
1. with the first flow1-m1, append it to flow1. flow1 has m1.
2. with the next flow1-m2, merge it to m1. flow1 has m1+m2.
3. with the next flow1-m4, flush m1+m2 instantly and then append it to flow1. flow1 has m4
4. with the next flow1-m5, append it to m4. flow1 has m4+m5.
5. with the next flow1-m3, append it to the front of m4+m5. flow1 has m3+m4+m5.

So, [flow1-[m1+m2], flow1-[m3+m4+m5]] will be handled. I think there is no issue to be solved there.

hhaim:
@jsmoon in regards to the last ```do_flush() ``` comment. 
Sorry for not being clear. I understand the code. My concern that the GRO code has a duplicate code for the ```is_close()``` handling which is **not** related to the merging of packets and I would like to remove this duplication. (the process of the return result should be at one place)

There are two options to remove this dependency: 

1) Remove the ```do_flush()``` and ```is_close()``` from GRO code 
the idea is that ```process_software_lro(ctx, flow, mbuf, lpTcp, ftuple)``` can have at most only **two results back**( (e.g. mbuf)  
the merge packets  and the new one that is not eligible and need to be flush as it is (e.g. FIN). The common case is zero (in case if it can be first eligible or merged) or one(eligible but OOO, give back the flushed). 

so change the API to be something like this 
```


class TcpResultObject {
    public:
    struct rte_mbuf * mbuf;
    TCPHeader    *    lpTcp;
};

class TcpResultObjects {
  public:
    uint8_t          size;     

    TcpResultObject m_res[2];
}

void HOT_FUNC CFlowTable::process_tcp_packet(CTcpPerThreadCtx * ctx,
                                    CTcpFlow *  flow,
                                    struct rte_mbuf * mbuf,
                                    TCPHeader    * lpTcp,
                                    CFlowKeyFullTuple &ftuple){
    TCPHeader tcphdr;
    if (unlikely(!flow->is_activated())) {
        tcphdr = *lpTcp;
    } else {
#ifndef TREX_SIM
        TcpResultObjects res;
        if (process_software_lro(ctx, flow, mbuf, lpTcp, ftuple,&res))
            return;
#endif
    int i;
    for (i=0; i<int(res.size);i++) {
        TcpResultObject *lpo = &res.m_res[i]
        tcp_flow_input(flow, lpo->mbuf, lpo->lpTcp, ftuple);

        if (unlikely(flow->new_template_identified())) {
            /* identifying new template has been done */
            if (!update_new_template(ctx, flow, &tcphdr, ftuple)) {
                return; /* when update is failed, flow is not available also. */
            }
        }

        if (flow->is_can_close()) {
            handle_close(ctx,flow,true);
        }
    }

```

2. give a callback to the GRO code, similar to what you have today but to extend it 

```

void HOT_FUNC CFlowTable::process_tcp_packet_internal(CTcpPerThreadCtx * ctx,
                                    CTcpFlow *  flow,
                                    struct rte_mbuf * mbuf,
                                    TCPHeader    * lpTcp,
                                    CFlowKeyFullTuple &ftuple){

    tcp_flow_input(flow, mbuf, lpTcp, ftuple);

    if (unlikely(flow->new_template_identified())) {
        /* identifying new template has been done */
        if (!update_new_template(ctx, flow, &tcphdr, ftuple)) {
            return; /* when update is failed, flow is not available also. */
        }
    }

    if (flow->is_can_close()) {
        handle_close(ctx,flow,true);
    }
}

```

and GRO will have a callback 

m_tcp_lro->cb = process_tcp_packet_internal


and internaly the flush will call to the callback 

```
m_cb(ctx,flow,mbuf,lpTcp,ftuple) 
```

There is no ```is_can_close()``` in this solution  

I think options 2 is better and simpler

hhaim:
@jsmoon 
```
m_lro_nsegs(struct mbuf *m, int l7_len)
{
    int segsz = ((struct rte_mbuf *)m)->tso_segsz;
    return segsz ? (l7_len+segsz-1)/segsz: 1;
}

```

```tso_segsz``` is set by the tx path and the driver is using it (before sending). You are using this field in rx side, I think it is set to zero, you should provide the MSS to the function or just keep the old code (I don't see an issue with it)

```
    return ((struct rte_mbuf *)m)->nb_segs;
```

**Comment by hhaim** [comment]
@jsmoon just checked the perf on our setup 

* better to have an indication that this feature is enabled/disable at startup 
```
    if (!CGlobalInfo::m_options.preview.get_dev_lro_support() &&
        !CGlobalInfo::m_options.preview.getLroOffloadDisable()) {
```
```
 driver capability  : TCP_UDP_OFFLOAD  TSO 
 set dpdk queues mode to RSS_DROP_QUE_FILTER 
 ```

* in case of http_simple.py with 100K active flows and TSO (X710) in 40gbps and 7 cores 
I see in both cases 14.5% CPU without improvment, however the number of packets is reduced from 24 to 8 
which is good. I don't know to explain it, will test in another setup

hhaim:
@jsmoon testing it with mlx5/CX-5 cards when LRO hardware is disabled as a hack, I see a degradation in performance with software GRO (53% without vs 56% with).

need to understand it .., how could it be that you saved packets and yet the performance is worse

hhaim:
@jsmoon 
1. Why not like this?
```

void HOT_FUNC CFlowTable::process_tcp_packet_internal(CTcpPerThreadCtx * ctx,
                                    CTcpFlow *  flow,
                                    struct rte_mbuf * mbuf,
                                    TCPHeader    * lpTcp,
                                    CFlowKeyFullTuple &ftuple){

    tcp_flow_input(flow, mbuf, lpTcp, ftuple);

    if (unlikely(flow->new_template_identified())) {
        /* identifying new template has been done */
        if (!update_new_template(ctx, flow, &tcphdr, ftuple)) {
            return; /* when update is failed, flow is not available also. */
        }
    }

    if (flow->is_can_close() && ctx) {
        handle_close(ctx,flow,true);
    }
}

void HOT_FUNC CFlowTable::process_tcp_packet(CTcpPerThreadCtx * ctx,
                                    CTcpFlow *  flow,
                                    struct rte_mbuf * mbuf,
                                    TCPHeader    * lpTcp,
                                    CFlowKeyFullTuple &ftuple){
    TCPHeader tcphdr;
    if (unlikely(!flow->is_activated())) {
        tcphdr = *lpTcp;
    } else {
#ifndef TREX_SIM
        if (process_software_lro(ctx, flow, mbuf, lpTcp, ftuple))
            return;
#endif
    }
    process_tcp_packet_internal(ctx,flow,mbuf,lpTcp,ftuple);
}

```
2. Did you look into the perf issue? I didn't find a case that this mechanism helps with performance at scale. It was surprising to see that TSO microburst is working even with a very high number of flows and the GSO merge many packets. 

I I used http_simple.py with init_wnd=10 to help with bigger window of TSO

jsmoon:
1. `if (unlikely(!flow->is_activated()))` and `if (unlikely(flow->new_template_identified()))` code blocks are for `l7_map` feature only. I think there is no need to do offload during identifying l7_map template.

2. It seems the overhead of software LRO is quite large. In my quick test, the `tcps_rcvoffloads` counter should have double of `tcps_rcvtotal`. The test result shows that about 3~4 packets should be merged to cover the overhead. In http_simple.py case, about 10 packets are merged and it gets quite a large performance gain.

@hhaim, I made this change for default action but it cannot cover all the cases. It was not a good approach to add a new feature. So, what do you think about changing it optional? (activated by --software-lro)

hhaim:
1. I think there is no relation betwean offloading the l7_map and you should **keep** the same logic for both, this is exacly why I don't think this logic should be part of the software LRO and this diversion will bring bugs in the future. Do you see a reason not to add this logic to the ```process_tcp_packet_internal``` and remove the duplicate code?

2. I agree with your observation, with http_simple.py,  iniwnd=10 this is the perfect case of software LRO, I **do** see more that 10 packets that are merged (~33% of the total packets) however do **not** see CPU% improvment.

My point is that if in the perfect case we don't see improment, there is no reason to add this feature so we need to understand why

jsmoon:
@hhaim

1. No. I'll change it as you suggested.

2. In my test with http_simple.py in a direct loopback environment, about 80% of total packets are merged.
```
             tcps_sndtotal |           8450270 |           5561130 | total packets sent         
             tcps_rcvtotal |           7205885 |           8450246 | total packets received     
          tcps_rcvoffloads |          26293754 |                 0 | receive offload packets by software 
```
In this case, CPU% is also improved.

I understood that I should try to decrease software LRO overhead.

jsmoon:
@hhaim, I'm sorry I missed a point in the process_tcp_packet.
```
    TCPHeader tcphdr;
    if (unlikely(!flow->is_activated())) {
        tcphdr = *lpTcp;
    } else {
```
and
```
    if (unlikely(flow->new_template_identified())) { 
        /* identifying new template has been done */ 
        if (!update_new_template(ctx, flow, &tcphdr, ftuple)) { 
            return; /* when update is failed, flow is not available also. */ 
        } 
    } 
```
should be located in the same function.
It is because *lpTcp cannot be accessed because tcp_flow_input will free it.

hhaim:
@jsmoon 

1. I don't understand why there is a need for this from the first place, why can't you use the original pointer?  fixing an issue with another issue is not a good solution. Even if it needed, you can solve this by doing the same trick you did before  
```
    TCPHeader tcphdr;
    if (unlikely(!flow->is_activated())) {
        tcphdr = *lpTcp;
    } else {
```
2. What is the CPU% in your case?  what is the command? what is the rate? how many cores? 
Could you show one case that it show CPU improvment, I could verify it 

>   @jsmoon testing it with mlx5/CX-5 cards when LRO hardware is disabled as a hack, I see a degradation in performance with 
>   software GRO (53% without vs 56% with).
>   need to understand it .., how could it be that you saved packets and yet the performance is worse

jsmoon:
@hhaim 

1. lpTcp refers to the contents in the mbuf. But mbuf is changed and freed in the tcp_flow_input(),
After tcp_flow_input(), lpTcp points freed memory area. But it is needed to send RST when no template is found.
Anyway, I'll change it next push.

2.
```
 set driver name net_i40e_vf 
 driver capability  : TCP_UDP_OFFLOAD  TSO 
 set dpdk queues mode to MULTI_QUE 
 Number of ports found: 2

trex>start -f astf/http_simple.py -m 10000
```
without software LRO
```
connection   : localhost, Port 4521                       total_tx_L2  : 7.65 Gbps                      
version      : ASTF @ v2.93                               total_tx_L1  : 7.79 Gbps                      
cpu_util.    : 25.47% @ 2 cores (2 per dual port)         total_rx     : 7.64 Gbps                      
rx_cpu_util. : 0.0% / 0 pps                               total_pps    : 858.91 Kpps                    
async_util.  : 0% / 17.97 bps                             drop_rate    : 0 bps                          
total_cps.   : 27.7 Kcps                                  queue_full   : 0 pkts                         
```
with software LRO
```
connection   : localhost, Port 4521                       total_tx_L2  : 7.68 Gbps                      
version      : ASTF @ v2.93                               total_tx_L1  : 7.81 Gbps                      
cpu_util.    : 22.07% @ 2 cores (2 per dual port)         total_rx     : 7.65 Gbps                      
rx_cpu_util. : 0.0% / 0 pps                               total_pps    : 869.44 Kpps                    
async_util.  : 0% / 4.35 bps                              drop_rate    : 0 bps                          
total_cps.   : 27.77 Kcps                                 queue_full   : 0 pkts                         
```

hhaim:
@jsmoon 
1. understood. good point, this is much bigger problem that the is_active(), I think it is better to defer the mbuf releasing or copy the header each time (in the internal function you copy it is stack)
```
tcp_input(struct tcpcb *tp, struct mbuf *m, struct tcphdr *th, int toff, int tlen, uint8_t iptos)
```

```
    tcp_fields_to_host(th);
``'

2. in my case with 7 cores I see 25% without and 27% with software LRO. Will check it again with latest changes 

Are you testing it with http_simple.py with initwnd=10 ?

**Comment by hhaim** [comment]
I did on XL710 setup with http_simple.py and I do see an improvment 


** with software LRO**

```
Global Statistics

connection   : csi-trex-08, Port 4501                     total_tx_L2  : 68.95 Gbps                     
version      : ASTF @ v2.95                               total_tx_L1  : 70.2 Gbps                      
cpu_util.    : 29.9% @ 14 cores (7 per dual port)         total_rx     : 68.95 Gbps                     
rx_cpu_util. : 0.0% / 0 pps                               total_pps    : 7.85 Mpps                      
async_util.  : 0% / 71 bps                                drop_rate    : 0 bps                          
total_cps.   : 249.38 Kcps                                queue_full   : 638 pkts                       

          tcps_connattempt |           8804509 |                 0 | connections initiated      
              tcps_accepts |                 0 |           8804338 | connections accepted       
             tcps_connects |           8804362 |           8804338 | connections established    
              tcps_closed |           8678795 |           8678739 | conn. closed (includes drops) 
            tcps_segstimed |          26412994 |          28717199 | segs where we tried to get rtt 
           tcps_rttupdated |          26412645 |          49069225 | times we succeeded         
               tcps_delack |             12471 |                 0 | delayed acks sent          
             tcps_sndtotal |          57873890 |          37521664 | total packets sent         
              tcps_sndpack |           8804362 |          19913105 | data packets sent          
              tcps_sndbyte |        2192322741 |      282566423772 | data bytes sent by application 
           tcps_sndbyte_ok |        2192286138 |      282564895084 | data bytes sent by tcp     
              tcps_sndctrl |           8804509 |                 0 | control (SYN|FIN|RST) packets sent 
              tcps_sndacks |          40265020 |          17608559 | ack-only packets sent      
             tcps_rcvtotal |          50242586 |          57873690 | total packets received     
              tcps_rcvpack |         138106848 |          17608432 | packets received in sequence 
              tcps_rcvbyte |      282562273552 |        2192280162 | bytes received in sequence 
           tcps_rcvackpack |          99606537 |          49069225 | rcvd ack packets           
           tcps_rcvackbyte |        2192263728 |      282561261500 | tx bytes acked by rcvd acks  
        tcps_rcvackbyte_of |           8804011 |           8803983 | tx bytes acked by rcvd acks - overflow acked 
          tcps_rcvoffloads |         169862525 |                 0 | receive offload packets by software 
              tcps_preddat |          23829941 |                 0 | times hdr predict ok for data pkts  
              tcps_predack |                 0 |          22656810 | times hdr predict ok for acks 

```

** without**

```
connection   : csi-trex-08, Port 4501                     total_tx_L2  : 68.66 Gbps                     
version      : ASTF @ v2.95                               total_tx_L1  : 69.85 Gbps                     
cpu_util.    : 32.1% @ 14 cores (7 per dual port)         total_rx     : 68.66 Gbps                     
rx_cpu_util. : 0.0% / 0 pps                               total_pps    : 7.47 Mpps                      
async_util.  : 0% / 37.05 bps                             drop_rate    : 0 bps                          
total_cps.   : 249.06 Kcps                                queue_full   : 0 pkts                         

          tcps_connattempt |          17598440 |                 0 | connections initiated      
              tcps_accepts |                 0 |          17598279 | connections accepted       
             tcps_connects |          17598314 |          17598279 | connections established    
               tcps_closed |          17448142 |          17448075 | conn. closed (includes drops) 
            tcps_segstimed |          52770238 |          52745303 | segs where we tried to get rtt 
           tcps_rttupdated |          52769857 |          70318641 | times we succeeded         
               tcps_delack |          17573590 |                 0 | delayed acks sent          
             tcps_sndtotal |          87917183 |          70343709 | total packets sent         
              tcps_sndpack |          17598314 |          35171866 | data packets sent          
              tcps_sndbyte |        4382011560 |      564799166226 | data bytes sent by application 
           tcps_sndbyte_ok |        4381980186 |      564399995354 | data bytes sent by tcp     
              tcps_sndctrl |          17598440 |                 0 | control (SYN|FIN|RST) packets sent 
              tcps_sndacks |          52720430 |          35171844 | ack-only packets sent      
             tcps_rcvtotal |         439633525 |          87917047 | total packets received     
              tcps_rcvpack |         422035211 |          35171717 | packets received in sequence 
              tcps_rcvbyte |      564396850400 |        4381971471 | bytes received in sequence 
           tcps_rcvackpack |          52769857 |          70318641 | rcvd ack packets           
           tcps_rcvackbyte |        4381948563 |      564004292444 | tx bytes acked by rcvd acks  
        tcps_rcvackbyte_of |          17573356 |          17573337 | tx bytes acked by rcvd acks - overflow acked 
              tcps_preddat |         386863668 |                 0 | times hdr predict ok for data pkts  
              tcps_predack |                 0 |          17573587 | times hdr predict ok for acks 
          err_rx_throttled |                74 |                 0 | rx thread was throttled    
```


It would be nice to add an indication at startup that we are with software LRO,as for now  the tcp counters is the only indication

**Comment by jsmoon** [comment]
@hhaim, I've added the following message before.
```
RX core uses TX queue number 65535 on all ports
 core, c-port, c-queue, s-port, s-queue, lat-queue
 ------------------------------------------
 1        0      0       1       0      0  
 -------------------------------
TCP LRO is activated by software at 0x1cf6458
TCP LRO is activated by software at 0x1de79b8
```
I thought that this is not a driver's capability.
So, I printed the message independently when the software LRO is enabled.

I've pushed another update just before. Please check it now.

**Comment by hhaim** [comment]
@jsmoon nice work.

```TCP LRO is activated by software at 0x1cf6458``` this should be reported once and not for each queue.

I think it is better to be somthing like this (*SLRO* - software LRO)

```
set driver name net_i40e_vf 
driver capability  : TCP_UDP_OFFLOAD  TSO  SLRO
set dpdk queues mode to MULTI_QUE 
Number of ports found: 2
```

hhaim:
@jsmoon thanks you. rebased

PR #802, opened 2022-02-16
GTP get/delete clients

xmaruto:
Hi Hanoch, this is Max Klyus from mobility team Cisco. Would you be so kind to take a look on a small PR from us, as we are using this small changes for our environment and it will be nice to sync it with upstream.
Thank you.


xmaruto:
sorry, one file is missed in this PR, I will create the new one

PR #803, opened 2022-02-16
GTP get/delete clients

xmaruto:
This patch allows to get GTP clients stats, and delete GTP clients

### DIFF
--- a/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
@@ -1228,7 +1228,18 @@ class ASTFClient(TRexClient):
  
         return json_attr
 
-    # execute 'method' for inserting/updateing tunnel info for clients
+    # private function to form json data for GTP tunnel
+    def _delete_gtp_tunnel(self, client_list):
+
+        json_attr = []
+
+        for value in client_list:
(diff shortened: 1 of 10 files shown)


hhaim:
@xmaruto @syaakov 

1)  I would remove the ```thread_id ``` from the delete_API, look into the update API there is no thread_id there
            json_attr.append({'client_ip': value.get('client_ip'), 'thread_id': value.get('thread_id')})

2) You can't access the tuple generator  which is private per DP, convert it to a multiple messages from CP->DP
```
trex_rpc_cmd_rc_e
TrexRpcCmdGetClientsStats::_run(const Json::Value &params, Json::Value &result) {

    std::map<string, uint32_t> msg_data;

    auto astf_db = CAstfDB::get_instance(0);
    cpu_util_full_t cpu_util_full;

    if (get_platform_api().get_cpu_util_full(cpu_util_full) != 0) {
        return TREX_RPC_CMD_INTERNAL_ERR;
    }

    Json::Value res = Json::arrayValue;

    uint32_t all_thread_active = 0;
    uint32_t all_thread_inactive = 0;
    uint32_t all_thread_gtpu = 0;

    for (int thread_id = 0; thread_id < cpu_util_full.size(); thread_id++) {

        CTupleGeneratorSmart* ctg = astf_db->get_smart_gen(thread_id); <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
```

PR #806, opened 2022-02-21
Gtpu delete clients

xmaruto:
Delete GTPU clients


hhaim:
@xmaruto looks good

hhaim:
@xmaruto why not to add this function to update_client?

hhaim:
@xmaruto there is a command to disable client and update GTPU context. topo can be used to set on-startup enable/disbale and context. I think this cover everything

PR #817, opened 2022-03-10
fix missing per template tunables on new TCP stack

jsmoon:
Hi, I found the missing per template tuneable on the recently merged new TCP stack.
They were `tcp.mss`, tcp.initwnd`, and `tcp.no_dely`.
This PR is to support them in the new TCP stack.


hhaim:
rebased

PR #820, opened 2022-03-13
Fixed typo for Latency

debanjana-a11y:
Fixed typo for Latency which was "Latecny" in some files. Please review the changes.


hhaim:
@debanjana-a11 too late .. this is an API that we can't break.

PR #821, opened 2022-03-14
fix SLRO mbuf double free issue

jsmoon:
Hi, this PR is to fix mbuf double-free issue due to SLRO.

There were two reasons:
1. when flow is closed by the merged packets in flush_bufs(), mbuf would be freed twice due to the remaining m_lro_buf info.
2. when drop_hdrlen is greater than the first mbuf's data_len, the first mbuf is freed by m_adj_fix and freed again as a handled one.

My change will fix these issues.

@hhaim, please check my fix and give your feedback.

### DIFF
--- a/src/44bsd/flow_table.cpp
+++ b/src/44bsd/flow_table.cpp
@@ -966,7 +966,7 @@ void CTcpRxOffloadBuf::set(CTcpFlow* flow, struct rte_mbuf* mbuf, TCPHeader* lpT
     flow->m_lro_buf = this;
 }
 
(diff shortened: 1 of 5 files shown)


hhaim:
@jsmoon hmm it looked very fundamental, I wonder how we didn't find it in our regression. Every mbuf_adjust should give the wrong mbuf and crash with wrong adjust
The LRO in some cases free twice the same mbuf 

How did you find it?

hhaim:
rebased

Issue #813, opened 2022-03-03
MAX_TCP_REASS_BLOCKS is too small for large OOO packets.


jsmoon:
Hi, I have been observed the following `tcps_rcvoopackdrop` counters when the traffic path has many nodes in its route.
```
  tcps_sndrexmitpack |               0                        10886             
  tcps_sndrexmitbyte |               0                      4198326             
     tcps_rcvduppack |             128                            0             
     tcps_rcvdupbyte |          107527                            0             
 tcps_rcvpartduppack |               5                            0             
 tcps_rcvpartdupbyte |          104332                            0             
  tcps_rcvoopackdrop |            2256                            0             
 tcps_rcvoobytesdrop |         3066356                            0             
      tcps_rcvoopack |            2507                            0             
      tcps_rcvoobyte |         3295510                            0             
      tcps_rcvdupack |               0                         2485             
```
Since some nodes do not guarantee packet ordering, the `tcps_rcvoopack` can be observed in general.
But, as you can see, lots of `tcps_rcvoopack` were dropped by `tcps_rcvoopackdrop`, about 90%.
I think it caused unexpected performance degradation.

As far as I know, `tcps_rcvoopackdrop` is increased when the reassembly block is full.
The maximum number of reassembly blocks is 4 which is defined by MAX_TCP_REASS_BLOCKS.
I think 4 is too small in this case.
@hhaim, what do you think about using a larger number of reassembly blocks?

hhaim:
@jsmoon in any case the ```CTcpReass``` is  allocated dynamically only when there is an out of order (OOO) issue, so enlarging the **MAX_TCP_REASS_BLOCKS** will not have an impact when there is no OOO issue . However, at scale when **all** the flows have many OOO this could be a problem (of memory and performance in server side) and I think we should maybe limit the number of this objects ```CTcpReass``` instances. 

I would capture the OOO and learn if it is an inherent problem (the node OOO all the flows and it depends in the window size) or only part of the flow has this issue. 
Another question what would be the ideal value of MAX_TCP_REASS_BLOCKS in this case?

hhaim:
@jsmoon another idea is to have a tunables for this. 
1. MAX_TCP_REASS_BLOCKS ==> max_tcp_reass_blocks_per_flow ==> by moving the object to the end of the struct and allocate the sizeof(CTcpReassBlock)* tuable
2. max_ooo_flows 


but first we need to understand the issue

[...]

PR #823, opened 2022-03-14
Improve tcp reass

jsmoon:
Hi, this PR is the change for issue #813.

It is improved by using std::map.
I added the simple histogram statistics for the maximum queue size.


hhaim:
@jsmoon did you try this on the original example you provided? did it improve the performance?

jsmoon:
@hhaim, regarding performance, I didn't check it.
I had been focused on feature improvement.

hhaim:
@jsmoon I mean this example (one flow with many drops)
https://github.com/cisco-system-traffic-generator/trex-core/issues/813

jsmoon:
@hhaim I've checked this change is working for the example by a simple simulation program.
It was similar to the unit test.

hhaim:
@jsmoon so to make sure I understand. in case of #813 example, after this fix the issue was solved and the DUT just happen to create many out of order that the bigger queue size (100) solves (without the need to transmission or at least very few one) ? 
There could be another option that there were a real drop in the window that this bigger queue only **mitigate** by saving the holes

jsmoon:
@hhaim, my user did not test with this yet.
With this fix, I expect that I can distinguish the packet drops from out of ordered packets.
I'll share the result after the test is done.

Issue #807, opened 2022-02-22
portattr -a doesn't print all ports


bdollma:
```
trex>portattr -h
usage: portattr [-h] [-p PORTS [PORTS ...] | -a] [--prom {on,off}]
                [--link {up,down}] [--led {on,off}] [--fc {none,tx,rx,full}]
                [--vxlan-fs [VXLAN_FS [VXLAN_FS ...]]] [--supp]
                [--mult {on,off}]

Sets port attributes

optional arguments:
  -h, --help            show this help message and exit
  -p PORTS [PORTS ...], --port PORTS [PORTS ...]
                        A list of ports on which to apply the command
  -a                    Set this flag to apply the command on all available
                        ports
  --prom {on,off}       Set port promiscuous on/off
  --link {up,down}      Set link status up/down
  --led {on,off}        Set LED status on/off
  --fc {none,tx,rx,full}
                        Set Flow Control type
  --vxlan-fs [VXLAN_FS [VXLAN_FS ...]]
                        UDP ports for which HW flow stats will be read from
                        layers after VXLAN
  --supp                Show which attributes are supported by current NICs
  --mult {on,off}       Set port multicast on/off
```

The `-a` specifies that it will show all ports, however:

```
trex>portattr -a
Port Status

     port       |          0           |          1           |          2           |          3
----------------+----------------------+----------------------+----------------------+---------------------
driver          |    net_e1000_igb     |    net_e1000_igb     |    net_e1000_igb     |    net_e1000_igb
description     |  I350 Gigabit Netwo  |  I350 Gigabit Netwo  |  I350 Gigabit Netwo  |  I350 Gigabit Netwo
link status     |          UP          |          UP          |          UP          |          UP
link speed      |        1 Gb/s        |        1 Gb/s        |        1 Gb/s        |        1 Gb/s
port status     |         IDLE         |         IDLE         |         IDLE         |         IDLE
promiscuous     |         off          |         off          |         off          |         off
multicast       |         off          |         off          |         off          |         off
flow ctrl       |         none         |         none         |         none         |         none
vxlan fs        |          -           |          -           |          -           |          -
--              |                      |                      |                      |
layer mode      |         IPv4         |         IPv4         |         IPv4         |         IPv4
src IPv4        |       1.1.1.2        |       1.1.2.2        |       1.1.3.2        |       1.1.4.2
IPv6            |         off          |         off          |         off          |         off
[...]

PR #824, opened 2022-03-18
Solution for "portattr -a doesn't print all ports #807"

debanjana-a11y:
Hi, Please review the code changes. I have tested it with trex server running in astf mode and trex-console.
Any feedback is welcomed.


bdollma:
Hi @debanjana-a11y,
See comment in the original issue.

Issue #829, opened 2022-03-21
ASTF capture packet twice with software RSS.


jsmoon:
Hi, I think the following code in `CFlowTable::parse_packet()` could make capturing packets twice.

https://github.com/cisco-system-traffic-generator/trex-core/blob/42c368a6af101a587e7c224f96fa281c5057eee7/src/44bsd/flow_table.cpp#L163

The `parse_packet()` had been called at `CFlowTable::rx_handle_packet()`.
But, after software RSS is merged, it is also called at `CFlowGenListPerThread::handle_astf_software_rss()`.
When the capturing packet is enabled, a received packet could be copied to the buffer by software RSS once and later copied once more by rx_handle_packet.

I think this code block should be moved from `parse_packet()` to the `rx_handle_packet()`.
@hhaim what do you think about this issue?

hhaim:
@jsmoon do you mean moving  ```handle_pkt_rx_dp``` up?

jsmoon:
@hhaim yes, like the following code in the `CFlowTable::rx_handle_packet()`.
```
    parse_packet(mbuf,
                parser,
                tuple,
                ftuple,
                ctx->get_rx_checksum_check(),
                action,
                port_id);

    if ( action != tPROCESS ) {
        rx_non_process_packet(action, ctx, mbuf);
        return false;
    }

    TrexCaptureMngr::getInstance().handle_pkt_rx_dp(mbuf, port_id); 

    if ( is_idle ) {
        rte_pktmbuf_free(mbuf);
        return false;
    }
```

hhaim:
@jsmoon   it will work, but there is this action that is not used   ```action=tDROP;``` that won't work in the future

PR #830, opened 2022-03-22
fix packet capture twice with software RSS

jsmoon:
Hi, this PR is the change for #829 

In `rx_handle_packet()`, all the packets except tREDIRECT_RX_CORE will be captured.
It is the same as before.


hhaim:
@jsmoon rebased.

PR #831, opened 2022-03-23
fix TCP connection early drop by tcps_persisttimeo

jsmoon:
Hi, I recently met an early drop issue and this PR will fix it.

When the first PERSIST timeout happens, it sends a probe segment.
After that, unexpected tcp_drop() is followed immediately.
It causes an early drop without waiting for the probing result.

I'm sorry for this bug left during the new stack integration.


hhaim:
rebased

Issue #822, opened 2022-03-14
[Improvement suggestion] High performance packet capturing in TRex


egwakim:
Current performance of packet capturing is limited because it's done by remote JSON RPC processing.
High performance packet capturing is quite useful to troubleshoot issues or tuning.
There could be alternative ways to capture packets faster,
Altertive#1. Add option to save .pcap file by TRex server's master_thread instead of sending JSON RPC to client
Others : There could be other ways

Hi @hhaim,
How about your opinion about it?

hhaim:
@egwakim there is a feature to capture directly from server using ZMQ channel, have a look into the manual and example in the regression. 
There was an issue with watchdog (when closing the ZMQ channel that stuck the Rx thread but I think it was fixed)

hhaim:
@egwakim see this
https://trex-tgn.cisco.com/trex/doc/trex_stateless.html#_using_capture_port_for_faster_packet_capture_packet_injection

egwakim:
Hi @hhaim 
It's for STL with service enabled, would it works in ASTF mode?

egwakim:
Hi @hhaim 
Can we apply it in ASTF mode?

hhaim:
@egwakim yes, the capture is a common code for interactive (STL and ASTF)

jsmoon:
@hhaim "start_capture_port" won't capture traffic packets of DP cores.
It only works with packets of RX core including redirected packets from DP cores. (e.g. latency packets)
I think it was not designed for packet capturing.

hhaim:
@jsmoon you are right, already discussed it in the other thread. 
I think the best option is to let the Rx core to pull the software ring and send it throw the zmq channel

jsmoon:
@hhaim I checked the performance of the zmq channel in a simple implementation.
The implementation is ...
- based on the TrexCaptureMngr and TrexCapture feature in CRxCore. (by JSON RPC "capture")
- added an "endpoint" parameter to establish the zmq channel.
- used TrexPktBuffer as the software ring
- TrexCapture pulls the packet in the TrexPktBuffer and writes to the zmq channel in the CRxCore::work_tick().

The result shows about **120Kpps**. But I need about **1Mpps**.
To get the performance requirement, I think I need to use **shared memory** to deliver the matched packets.

The first option to use the shared memory is the capture client offering the shared memory as the "endpoint".
[...]

PR #839, opened 2022-04-15
Packet capture by endpoint

jsmoon:
Hi, this PR is the change for issue #822.

There are lots of changes but the original capture method is still functional.
This feature will be activated only by an 'endpoint' parameter in the 'capture' 'start' request.
I have met a traffic performance issue (about 2/3 degradation during capturing) with the original packet buffer.
So I introduce a new 'endpoint buffer' for bulk operation and get the performance improvement. (about 1/8 degradation during capturing with a small snaplen value)
Because the endpoint can be a file, the buffer will keep the packets in the PCAP/PCAPNG binary format.

@hhaim please check my code and give your feedback

### DIFF
--- a/linux_dpdk/ws_main.py
+++ b/linux_dpdk/ws_main.py
@@ -2056,6 +2056,7 @@ dpdk_includes_path =''' ../src/
                         ../src/dpdk/lib/librte_ring/
                         ../src/dpdk/lib/librte_timer/
                         ../src/dpdk/lib/librte_ip_frag/
+                        ../src/dpdk/lib/librte_pcapng/
                         ../src/dpdk/
                         
                         ../src/dpdk/lib/librte_security/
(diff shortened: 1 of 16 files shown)


jsmoon:
@hhaim I've prepared the whole changes I've suggested.

It contains FILE and ZMQ endpoint like the followings. (changed from #822 suggestion)
image
image

This change supports PCAP and PCAPNG with minimal information.
I think we can add more information later to the PCAPNG file.
PCAP is the default. PCAPNG can be selected by adding "?pcapng" to the endpoint.

I also added the trex-console capture command parameters to use this feature.

The FILE endpoint use case:
```capture record start --tx 0 --rx 0 --endpoint file:////tmp/?pcapng --snaplen 120```
This makes /tmp/endpoint_<number> in the server host temporarily.
```capture record stop --id 1 -o xxx.pcapng```
The file will be saved to the `xxx.pcapng` file and removed from the server host.

The ZMQ endpoint use case:
```capture monitor start -o yyy.pcap```
The CaptureMonitor thread opens the ZMQ channel, receives captured packets, and writes them to `yyy.pcap`.

egwakim:
Hi @hhaim 
Can you check the PR?

bdollma:
Hi @egwakim , @jsmoon 
It is the holidays season in Israel, we will be back at work next week.
Thanks,

hhaim:
@jsmoon looks good. rebased

Issue #834, opened 2022-04-06
[Improvement suggestion] Get RTT, CWND info


egwakim:
Hi,
User want to migrate existing verification tool(iperf) to TRex.
TCP throughput was not the issue with TRex, but, TRex lacks information about rtt(round trip time) and cwnd(congestion window).
In classic OS like Linux and FreeBSD,
iperf using getsockopt() API to get rtt and cwnd information per socket.
` tcp_info info = {};`
`  socklen_t len = sizeof(tcp_info);`
`  getsockopt(socket_fd, SOL_TCP, TCP_INFO, &info, &len);`
`  printf("RTT=%d, CWND=%d\n", info.tcpi_rtt, info.tcpi_snd_cwnd)`

But, there is no mapping API in TRex.
I assume that following 2 approaches would be possible, 
Hi @hhaim, how about your opinion?
Alternative#1. Adding new "TCP info" counter type and add them(rtt, cwnd) in "TCP info" counter. Currently,  4 types of counters('General','TCP','UDP','Flow table') available(see https://trex-tgn.cisco.com/trex/doc/trex_astf.html#_counters_reference). Adding them in 5th 'TCP info' type would be possible.
Alternative#2. Simply adding them(rtt,cwnd) in TCP counters(https://trex-tgn.cisco.com/trex/doc/trex_astf.html#_counters_reference) can be considered,

hhaim:
@egwakim I think adding them to TCP section is better. The point about RTT/CWND that those counters should use the same infra of latency histogram (accurate). They are not a standard counters. 
I suggest to add this feature with a knob (disable by default) and only to sample of flows (e.g. 1-5%) because it could affect performance if you read the **accurate** high res counter. Another alternative is to use the 1msec resolution (the tick) with higher sample.

egwakim:
Hi @hhaim 
I assume that you prefer following, is it correct?
Accuracy vs performance need to be discussed more in the implementation phase.
![image](https://user-images.githubusercontent.com/4927607/162149932-52a3706e-e4b7-4e8f-b0bd-33198d2696fb.png)

hhaim:
@egwakim 
each "counter" should be an object with average/min/max and **histogram** (HDR) 
similar to latency object in STL/ASTF  
https://trex-tgn.cisco.com/trex/doc/cp_stl_docs/_modules/trex/stl/trex_stl_client.html#STLClient.get_pgid_stats

egwakim:
[...]

PR #841, opened 2022-04-20
get TCP flow information per profile

jsmoon:
Hi, this is the change for my suggestion in #834.

The user can get the information about a TCP flow per given tg_id.
It includes almost all the FreeBSD `struct tcp_info` information.

@hhaim please review my change and give your feedback.

### DIFF
--- a/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
@@ -689,7 +689,7 @@ class ASTFClient(TRexClient):
                 self.logger.info(format_text("Cannot remove a profile: %s is not state IDLE and state LOADED.\n" % profile_id, "bold", "magenta"))
 
     @client_api('command', True)
-    def start(self, mult = 1, duration = -1, nc = False, block = True, latency_pps = 0, ipv6 = False, pid_input = DEFAULT_PROFILE_ID, client_mask = 0xffffffff, e_duration = 0, t_duration = 0):
+    def start(self, mult = 1, duration = -1, nc = False, block = True, latency_pps = 0, ipv6 = False, pid_input = DEFAULT_PROFILE_ID, client_mask = 0xffffffff, e_duration = 0, t_duration = 0, dump_interval = 0):
         """
             Start the traffic on loaded profile. Procedure is async.
 
@@ -729,6 +729,10 @@ class ASTFClient(TRexClient):
                     Stop immediately (overrides nc pararmeter) when this time is over.
                     Disabled by default. Enabled by non-zero values.
 
+                dump_interval: float
+                    Interval time for periodic dump of TCP flow information: RTT, CWND, etc.
+                    TCP flow dump enabled by non-zero values.
+
             :raises:
                 + :exc:`TRexError`
         """
@@ -744,6 +748,7 @@ class ASTFClient(TRexClient):
             'client_mask': client_mask,
             'e_duration': e_duration,
             't_duration': t_duration,
+            'dump_interval': dump_interval,
             }
 
         self.ctx.logger.pre_cmd('Starting traffic.')
@@ -861,6 +866,51 @@ class ASTFClient(TRexClient):
         self.ctx.logger.post_cmd(rc)
         if not rc:
             raise TRexError(rc.err())
+
+    @client_api('getter', True)
+    def get_flow_info(self, profile_id = DEFAULT_PROFILE_ID, duration = 0):
+        """
+            Get TCP flow information
+
+            :parameters:
+                profile_id: string
+                    Input profile ID
+
+                duration: float
+                    Requests stacked TCP flow information during duration time
+                    Default value is 0 which means one time request.
+
+            :raises:
+                + :exc:`TRexError`
+
+        """
+        flows = {}
+        index = 0
+
+        timer = PassiveTimer(duration)
+        while True:
+            params = {
+                'profile_id': profile_id,
+                'index': index,
+                }
+            rc = self._transmit('get_flow_info', params = params)
+            if rc:
+                raise TRexError(rc.err())
+
+            for rc_flows in rc.data():
+                for flow_id in rc_flows:
+                    if flow_id == 'index':
+                        index = rc_flows[flow_id] + 1
+                    else:
+                        flows[flow_id] = flows.get(flow_id, [])
+                        flows[flow_id].append(rc_flows[flow_id])
+
+            if not duration or timer.has_expired():
+                break
+
+            time.sleep(min(1.0, duration))
+
+        return flows
 
     @client_api('command', True)
     def wait_on_traffic(self, timeout = None, profile_id = None):
@@ -1569,6 +1619,7 @@ class ASTFClient(TRexClient):
                                          parsing_opts.DURATION,
                                          parsing_opts.ESTABLISH_DURATION,
                                          parsing_opts.TERMINATE_DURATION,
+                                         parsing_opts.DUMP_INTERVAL,
                                          parsing_opts.ARGPARSE_TUNABLES,
                                          parsing_opts.ASTF_NC,
                                          parsing_opts.ASTF_LATENCY,
@@ -1612,7 +1663,11 @@ class ASTFClient(TRexClient):
             elif opts.servers_only:
                 kw['client_mask'] = 0
 
-            self.start(opts.mult, opts.duration, opts.nc, False, opts.latency_pps, opts.ipv6, pid_input = profile_id, e_duration = opts.e_duration, t_duration = opts.t_duration, **kw)
+            kw['e_duration'] = opts.e_duration
+            kw['t_duration'] = opts.t_duration
+            kw['dump_interval'] = opts.dump_interval
+
+            self.start(opts.mult, opts.duration, opts.nc, False, opts.latency_pps, opts.ipv6, pid_input = profile_id, **kw)
 
         return True
 
--- a/scripts/automation/trex_control_plane/interactive/trex/utils/parsing_opts.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/utils/parsing_opts.py
@@ -773,6 +773,15 @@ class OPTIONS_DB_ARGS:
          'default': 0.0,
          'help': "Set time limit waiting for all the flow to terminate gracefully."})
 
+    DUMP_INTERVAL = ArgumentPack(
+        ['--dump_interval'],
+        {'action': "store",
+         'metavar': 'TIME',
+         'dest': 'dump_interval',
+         'type': match_time_unit,
+         'default': 0.0,
+         'help': "Set interval time for dumping TCP flow information"})
+
     TIMEOUT = ArgumentPack(
         ['-t'],
         {'action': "store",
--- a/src/44bsd/flow_table.cpp
+++ b/src/44bsd/flow_table.cpp
@@ -397,8 +397,9 @@ bool CFlowTable::update_new_template(CTcpPerThreadCtx * ctx,
     assert(!flow->is_activated());
     if (flow->m_template_info) {
         flow->update_new_template_assoc_info();
-        flow->m_app.start(false);
         assert(flow->is_activated()); /* flow is now activated */
+        flow->m_pctx->set_flow_info(flow);
+        flow->m_app.start(false);
     } else {    /* no matched template found, now mbuf should be freed */
         CPerProfileCtx* pctx = FALLBACK_PROFILE_CTX(ctx);
         uint8_t mask = 3;
@@ -559,6 +560,7 @@ HOT_FUNC void CFlowTable::free_flow(CFlowBase * flow){
         delete udp_flow;
     }else{
         CTcpFlow * tcp_flow=(CTcpFlow *)flow;
+        flow->m_pctx->clear_flow_info(tcp_flow);
         tcp_flow->Delete();
         delete tcp_flow;
     }
@@ -993,6 +995,7 @@ bool CFlowTable::rx_handle_packet_tcp_no_flow(CTcpPerThreadCtx * ctx,
     if (unlikely(temp->has_payload_params())) {
         lptflow->start_identifying_template(pctx, payload_info);
     } else {
+        lptflow->m_pctx->set_flow_info(lptflow);
         app->start(true); /* start the application */
     }
 
--- a/src/44bsd/netinet/tcp_var.h
+++ b/src/44bsd/netinet/tcp_var.h
@@ -217,6 +217,7 @@ struct tcpcb {
     struct sackhint	sackhint;	/* SACK scoreboard hint */
     int	t_rttlow;		/* smallest observerved RTT */
     int	t_sndrexmitpack;	/* retransmit packets sent */
+    int	t_rcvoopack;		/* out-of-order packets received */
     struct cc_algo	*cc_algo;	/* congestion control algorithm */
     struct cc_var	*ccv;		/* congestion control specific vars */
     int	t_bytes_acked;		/* # bytes acked during current RTT */
--- a/src/44bsd/tcp_input.cpp
+++ b/src/44bsd/tcp_input.cpp
@@ -248,6 +248,7 @@ int CTcpReass::pre_tcp_reass(CPerProfileCtx * pctx,
     uint16_t tg_id = tp->m_flow->m_tg_id;
     INC_STAT(pctx, tg_id, tcps_rcvoopack);
     INC_STAT_CNT(pctx, tg_id, tcps_rcvoobyte, ti->ti_len);
+    tp->t_rcvoopack++;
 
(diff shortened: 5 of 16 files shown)


hhaim:
@jsmoon, I expect a message from ```TrexRpcCmdAstfGetFlowInfo``` CP->DP to collect the the flow info (as json). the flow could be closed by the time the CP is reading the data.

jsmoon:
@hhaim, I missed that point. I think I can add a sanity check if the flow is closed during the JSON dump.
When it is closed, the JSON dump will be discarded because it could be garbage. I'll update it tomorrow.

hhaim:
@jsmoon it is not advised to do this. Better to add CP->DP message as the flow memory is not protected with memory barrier, so you could have a race of free and read by CP

jsmoon:
@hhaim if I understand correctly, you are concerned CP might access the freed memory and cause a segmentation fault.
But I'd like to avoid the CP<->DP messages for this feature due to the DP performance impact.
So I think I can get the information fast enough to avoid the situation before the flow is closed.
Currently, while the flow is closing, DP clears the flow pointer and calls Delete(), and then deletes the flow.
If I copy the flow memory before the JSON dump, I guess that the memory access can be finished before the flow is deleted.

If you think this way still does not guarantee safer access, I can consider adding a memory barrier in CTcpFlow.

What do you think about this?

hhaim:
@jsmoon 

> avoid the CP<->DP messages for this feature due to the DP performance impact.

I don't see why there would be a DP performance issue? dumping the flow  stats in DP context **only** when RPC is called (rare not every packet) should not have any performance effect. 

When CP reads share information from DP, usually this is a memory that was allocated once and there is a **no** need for memory barrier (e.g. counters)

jsmoon:
@hhaim I mentioned the "DP performance impact" as the DP core interference by CP<->DP message handling. Please ignore the "performance" wording.
And I think I misused the term "memory barrier". I meant it to the protection mechanism preventing freed memory access.

Frankly speaking, I followed the "get_tg_id_stats" implementation. It gathers the statistics counters from DP memory directly in the CP core. I thought that this way minimizes the DP core interference. So I thought that solving the issue of illegal memory access was the key.
What is the difference between "get_tg_id_stats" and "get_flow_info" from your point of view?

hhaim:
@jsmoon the memory for tpgids (```get_tg_id_stats``` ) is allocated once per profile and there is a memory barrier (state-machine) when the profile is removed. If you remember there were a number of issues with illegal memory access into the counters memory. When you share data betwean two cores it is dangerous and there should be a good reason for it. 

In case of flow, you don't want to create a new protection mechanism (e.g. State-machine) for each new flow creation --> this will have a performance issue. 
The simplest way is to have CP->DP messaging as this is not a hot spot.

jsmoon:
@hhaim there is no issue when CP gets the flow pointer in the profile.
The only problem is the possibility that DP can free the flow memory while CP is reading it, as you mentioned at first.
I agree that CP<->DP message is the simplest way.
But, in my opinion, if we can handle the shared data carefully and definitely, it will be worth trying it.
If I could solve the issue with shared data without performance issues, could I try it anyway?

jsmoon:
@hhaim, when I revisited your suggestion, CP->DP message, I realized that there is no simple way to get the response.
AFAIK, in ASTF, all the CP->DP messages do not wait for responses.
CP master thread collects the reply messages asynchronous way and reports it through async_port.
It is rather complex to me.
I think the CP->DP message is not appropriate for the RPC request which requires JSON results in the response message.

jsmoon:
@hhaim, I updated my change to protect against illegal memory access using spinlock.
Do you still want me to add the CP->DP message?
Then, could you give me any ideas on how the client receives the result of "get_flow_info"?

hhaim:
Your code does not solve the problem and could create a crash see snipets,  and I don't want maintainer to keep thinking about this corner case. 
The current concept is simple: flows are created, free and accessed **only by DP** this has the advantage that you don't have false sharing on flows memory and can't have access to a flow that have been freed. 
Let's keep this concept and don't add corner case because we have a minor feature to read only flow statistic. 
the flow-stats should  be sent async to CP and CP can handle it


**The issue**:
 
```TrexRpcCmdAstfGetFlowInfo

  ->DumpTGFlows()
     ->dump_tg_flows()
       for (int i = 0; i < m_profile_ctx.size(); i++) {
	   pctx->get_tcp_info(tcp_info, tg_id); <<
       }

	
	CP->(time 0)
================

        if (flow && !flow->is_can_close()) { <<<(1) this is done without protection
		
			
	
            /* flow memory should be protected from DP free during CP access */
            CSpinLock slock(&m_tcpstat.m_tg_locks[flow->m_tg_id]);
            tcp_info.update_info(flow);


	from DP->(time 1)
================
(2) call free_flow() and finish this function (lock/unlock) the pointer in the vector is null 

HOT_FUNC void CFlowTable::free_flow(CFlowBase * flow){
    assert(flow);
    CPerProfileCtx* pctx = flow->m_pctx;
    flow->m_pctx->m_flow_cnt--;
    flow->m_pctx->on_flow_close();

    if ( flow->is_udp() ){
        CUdpFlow * udp_flow=(CUdpFlow *)flow;
        udp_flow->Delete();
        delete udp_flow;
    }else{
        CTcpFlow * tcp_flow=(CTcpFlow *)flow;
        tcp_flow->m_pctx->clear_flow_info(tcp_flow);
        tcp_flow->Delete();
        delete tcp_flow;  <<<<<<<<<<<<< this is called (lock unlock was already done)
    }



	CP->(time 2)
================
        CTcpFlow* flow = m_tcpstat.m_tg_flows[tg_id];  << this flow is not valid anymore, was removed and freed!!!!


                 if (flow && !flow->is_can_close()) { 
			
	
            CSpinLock slock(&m_tcpstat.m_tg_locks[flow->m_tg_id]);   <<<3 lock pass and flow was freed already 
            tcp_info.update_info(flow);  << update crash the flow is already freed 

```

jsmoon:
@hhaim thank you for your quick review. I updated my change.

> Let's keep this concept and don't add corner case because we have a minor feature to read only flow statistic.
the flow-stats should be sent async to CP and CP can handle it

Although CP can handle it, the client doesn't wait for it under the current mechanism.
In my experience, the DP->CP async message handling was quite complex.

hhaim:
@jsmoon it should be simple in this counters case. 
```CP->DP``` ask to send all the counters vector 
```DP->CP``` (each DP sends the vector ) async
CP < add the counters to a vector and turn **bit** for each core
once all the cores bitmask is on, you can read the vector of counters (and convert them to json)

**Comment by jsmoon** [comment]
@hhaim I know the mechanism, but mainly I want to know the client get the data.
As far as I know, there is no way currently.
I don't want to get the data asynchronousely. It forces the client's additional handling. I don't want it.

**Comment by hhaim** [comment]
@jsmoon the client should have the **same** blocking API (```get_tcp_flow_counters()```)
inside this client API you should have two RPC API to the server 

``` rpc_start_counter_reading()
    while True:
      if rpc_pool_counter_reading() == finish: (this API pool the core bitmask once they all up, read the data)
          break;
```
     
as the CP can't wait infinitely. 

You can simplify this by having one API and  let the CP to wait for a short time for all the results (in most cases it will work), but in rare cases return EAGAIN so the client should return the call

egwakim:
Hi all,
I see that there are different opinion about method to refer data between CP and DP.
"Messaging vs Direct access"

According to the @h's opinion
1) Messaging can solve illegal access issue in smooth way
According to @jsmoon 's opinion
It's heavier than direct access and it requires quite big change because it's differ from current way of accessing TCP statitics, we need to invent new way of accessing DP.

According to @hhaim 's opinion
It's light and simple enough to change

2) Direct access would requires some locking to prevent illegal access
According to @jsmoon 's opinion,
It's simple, lighter and align with current way of accessing TCP statistics

There is different opinion for the amount of required changes and weight in messaging method.
There will be no issue if we have simple and light solution for it.

Hi @hhaim 
Would it possible to suggest more detailed way of messaging method(Light and simple)?
Maybe code level suggestion would be very welcome.

egwakim:
Hi @hhaim 
How about your opinion?
Do you have detailed insight about it?

hhaim:
@egwakim the solution suggested by @jsmoon  ( spinlock) does not solve the problem and can **crash** the server. 

>> @hhaim this is the new way which is different to the current mechanism. Do you really want me to add this way for this feature?
Since the client has timeout for the response waiting, I should stop waiting for the DP response before timeout. Rarely the response can be mixed to the next request. And I should add the response to the DP idle state.
To my thought, it's not simple as you said. Your suggestion requires more considerations due to communication timing issue.
If I must do it this time as you suggested, I cannot help asking this question. Do you have plan to.improve all the statistics implementation to the right way also? To me, DP profile statistics has no different if I dedicated the selected flow memory to the profile by separated memory manager or locking.
If you have plan to do it. I agree that the new mechanism is worth doing it.

@jsmoon to answer your questions, My suggestion is the standard way to solve the problem. Dynamic counters  already use this type of solution. 

One more suggestion, I think that you can use the dynamic counter mechanism (that GTPU uses) could you have a look into this (add/remove set of counters in DP with names that map to tpgid) 

sharing flow memory betwean CP-DP is **not** an option

jsmoon:
@hhaim according to your suggestion, I think I can implement based on the following diagram.
image

jsmoon:
@hhaim,
> @jsmoon to answer your questions, My suggestion is the standard way to solve the problem. Dynamic counters already use this type of solution.
> 
> One more suggestion, I think that you can use the dynamic counter mechanism (that GTPU uses) could you have a look into this (add/remove set of counters in DP with names that map to tpgid)

This PR is not implemented to provide the **counters**. It is to provide the **TCP status** including TCP state at a certain time.
I think you are still seeing this feature as **statistics**.
I thought you agreed that the statistics feature is another story in your comment, "understood" [https://github.com/cisco-system-traffic-generator/trex-core/issues/834#issuecomment-1108276284] for my comment [https://github.com/cisco-system-traffic-generator/trex-core/issues/834#issuecomment-1099879711].
So, this PR is implemented based on thoughts of TCP flow inspection by the user.

If you want to get your statistics feature, please comment that on the original thread #834 and reject this PR, or suggest it in another new thread.
For the statistics feature implementation, I'll check your suggestion later.

> sharing flow memory betwean CP-DP is not an option

sharing flow memory between CP-DP should not be allowed, but sharing profile memory has no problem.
@hhaim, can I think of your answer like this? please confirm sharing profile memory.

> the solution suggested by @jsmoon ( spinlock) does not solve the problem and can crash the server.

My last commit is using mutex lock and I think I removed the crash possibility that you pointed out. I missed the other threads are running in the CP core at that time.
What makes you think that it still causes a crash?

hhaim:
@egwakim I understand your need (status per flow), I am not talking about my aggregated statistics ( e.g. all  flows latency syn-syn/ack). 

>> If you want to get your statistics feature, please comment that on the original thread https://github.com/cisco-system-traffic-generator/trex-core/issues/834 and reject this PR, or suggest it in another new thread.
For the statistics feature implementation, I'll check your suggestion later.

[hh] no, I'm not talking about this


What I've suggested is to use the dynamic-counters infra for the flow-tpgid-status (the **meaning** of the counter would be status). 
This is one option.  Another option is to implement async messaging to get the same API as you have today (from python point of view)


>> My last commit is using mutex lock and I think I removed the crash possibility that you pointed out. I missed the other threads are running in the CP core at that time.
What makes you think that it still causes a crash?

Read the example I gave, there would be a crash with suggested code and even if you fix this, I don't think it wise to use spinlock only for this feature, as this is a dangerous path and contradict the basic way trex works today (almost no sharing of information, only global counters are shared, dynamic **flow** **context** is a different story)

jsmoon:
@hhaim 

> What I've suggested is to use the dynamic-counters infra for the flow-tpgid-status (the meaning of the counter would be status).
This is one option.

Frankly speaking, I don't know about the dynamic-counters infra what you said already used for GTPU counters.
Could you elaborate on how the dynamic-counters infra can cover the `struct tcp_info`?

> Another option is to implement async messaging to get the same API as you have today (from python point of view)

Does this mean that **you can agree with my latest suggestion** depicted in the https://github.com/cisco-system-traffic-generator/trex-core/pull/841#issuecomment-1115856129 diagram?

> Read the example I gave, there would be a crash with suggested code

I think I've fixed it already as I have mentioned at https://github.com/cisco-system-traffic-generator/trex-core/pull/841#issuecomment-1111891153. With this change, the flow pointer in get_tcp_info() will be null during the lock is acquired after clear_flow() has been done. Please check [5b38c87](https://github.com/cisco-system-traffic-generator/trex-core/commit/5b38c876069ad1ea4b7ad46e21365fe2b47ade37).

> and even if you fix this, I don't think it wise to use spinlock only for this feature, as this is a dangerous path 

What do you see as the dangerous path? spinlock usage? sharing dynamic flow memory?

> and contradict the basic way trex works today (almost no sharing of information, only global counters are shared, dynamic flow context is a different story)

As far as I know, DP profile statistics counters are not global counters. But they are shared between CP and DP. So, I can hardly agree with you on this point of view.

I don't think my change is the right way, but I don't think it is the wrong way. I think it is a possible way can be considered. If I understand correctly, your concern is not on the memory sharing itself, but you are concerned about the sharing of the dynamic memory. right?

Finally, I've proposed a new way to change as you suggested I think. Please confirm if you agree with that or not.

jsmoon:
@hhaim I looked into the dynamic counter shortly. I found the following code added a few months ago.
```
1095     uint8_t dp_core_count = astf->get_dp_core_count();
1096     DpsDynSts dp_sts;
1097     for (uint8_t core_id = 0; core_id < dp_core_count; core_id++) {
1098         static MsgReply<dp_sts_t> reply;
1099         reply.reset();
1100         TrexCpToDpMsgBase *msg = new TrexAstfDpInitTunnelHandler(activate, tunnel_type, loopback, reply);
1101         astf->send_msg_to_dp(core_id, msg);
1102         dp_sts.add_dp_dyn_sts(reply.wait_for_reply());
1103     }
```
The MsgReply<T> had been used for the request and reply with RX core. Your intention was to use this way for this feature?
I have thought that it is not designed to communicate with DP cores, so I gave up to use it this way.
If you don't mind this way, I'll change this PR with this right away.

hhaim:
@jsmoon, good point, actually I was wrong with this example, the example of ```dynamic_counters``` is not the right example. In this case we assume there is **no traffic** and no profiles so we can **wait** for messages response (no profile at all). We started by having async activation but we simplify it as there was no request for multiple profile for GTPU (we will need to fix this in the future, make the activation async if requested) 

see this:
```
    TrexAstf *astf = get_astf_object();
    bool is_idle = (astf->get_state() == AstfProfileState::STATE_IDLE) ? true : false;

    if (!is_idle) {
        generate_execute_err(result, "The server state should be IDLE");
    }

```

in **your** case you can't asume it.

The right example (that actually similar to your need) is this:

```
TREX_RPC_CMD(TrexRpcCmdGetClientsInfoStart, "get_clients_info_start");
TREX_RPC_CMD(TrexRpcCmdGetClientsInfoPull, "get_clients_info_pull");

```

(ClientsInfoStart) ask the data (async)
ClientsInfoPull waits for it. 

BTW, before this implementation, there was similar implementation to what you did

in ASTF we can't wait for message reply if there is traffic or if we load a profile. 
```reply.wait_for_reply()```

let me know if you want me to investigate your old fixed code (with spinlock) or better change the code to make the response in async way (like client status)

**Comment by jsmoon** [comment]
@hhaim, according to your last comment, I think https://github.com/cisco-system-traffic-generator/trex-core/pull/841#issuecomment-1115856129 is better. I'll prepare for it.

**Comment by hhaim** [comment]
@jsmoon looks good!


Minor comments:
1. because the DP->CP is async, you should handle the case that there is no profile (it was removed) 
with the current code has an unhandled exception, better to just drop the message 

```
void TrexAstf::add_flows_info(profile_id_t dp_profile_id, Json::Value& flows) {
    get_profile(get_profile_id(dp_profile_id))->add_flow_info(flows);
}
```

2. There is no need to add another node to the low level scheduler only for this feature. Just use timer object that work on top of it, 

```
        dump_node->m_type = CGenNode::TCP_FLOW_DUMP;
```

See example how to use the timer_wheel in DP 


```
CAstfFifRampup::CAstfFifRampup(CPerProfileCtx * pctx,
                   uint16_t           rampup_sec,
                   double             cps){
        /* total ticks */
        m_pctx = pctx;
        m_ctx = pctx->m_ctx;
        m_ticks = (uint32_t)rampup_sec*1000/TICK_MSEC;
        m_rampup_sec =rampup_sec;
        assert(m_ticks>1);
        m_cur_tick = 1;
        m_cps = cps;
        m_tmr = new CAstfTimerFunctorObj();
        assert(m_tmr);
        m_tmr->m_cb = std::bind(&CAstfFifRampup::on_timer_update,this,m_tmr);
        on_timer_update(m_tmr);
}

```

hhaim:
rebased

PR #843, opened 2022-04-28
capture monitor can save pcapng file by snaplen

jsmoon:
Hi, this small change is to save the PCAPNG format file by the 'capture monitor' command.

`capture monitor start --tx 0 --rx 0 -o xxx.pcapng -s 100` will save the captured packet in PCAPNG format by snaplen 100 bytes.

Please review the change and give your feedback.


hhaim:
rebased

PR #845, opened 2022-05-07
scripts: add adaptive connection rate test tool

rjarry:
```
Add a standalone script that connects to a TRex server and sends TCP/UDP
traffic at varying rates of connections per second following a binary
search algorithm to find a "non-drop rate".

The traffic is only stopped in between rate changes when there were
transmission errors to avoid existing flows from piling up due to
retransmissions. When there was no error, the multiplier is increased
without interrupting traffic generation.

The main focus here is "connections per second" and therefore the
traffic profile cannot be dissociated from TRex's multiplier value. This
prevents from allowing arbitrary traffic profiles. Instead the script
generates a traffic profile based on the provided options: TCP/UDP
ratio, data message size, number of data messages per connection, delay
between request and reply.

By default, the binary search runs until interrupted by the user. The
maximum number of iterations can be specified for use in automatic
tests.

The raw numbers of the best iteration can be written in the JSON format
to a file at the end of the test.

The --adaptive-window option allows to diverge from the standard RFC
2544 non drop rate binary search algorithm. When this option is
specified, after three consecutive error/success iterations, the
lower/upper binary search bound will be decreased/increased by
--adaptive-window percent.

Here are some example runs:

10 iterations, 5% UDP connections, CPS range 10k-1m:

  ./cps_ndr.py -o cps.json -i10 -u5 -m10k -M1m

200 requests/replies, 1200 bytes per message, CPS range 1k-200k:

  ./cps_ndr.py -o cps.json -n200 -s1200 -m10k -M1m

Same profile with server replies delayed 100ms:

  ./cps_ndr.py -o cps.json -w100 -n200 -s1200 -m10k -M1m
```

### DIFF
--- a/scripts/cps_ndr.py
+++ b/scripts/cps_ndr.py
@@ -1,0 +1,528 @@
+#!/usr/bin/env python3
+# vim: tw=88 et sts=4 ts=4 sw=4
+
+"""
+Connect to a TRex server and send TCP/UDP traffic at varying rates of connections per
+second following a binary search algorithm to find a "non-drop rate". The traffic is
+only stopped in between rate changes when there were transmission errors to avoid
+existing flows from piling up.
+"""
+
+import argparse
+import json
+import pathlib
+import sys
+import time
+
+# ugly
+here = pathlib.Path(__file__).parent
+sys.path.insert(0, str(here / "external_libs/texttable-0.8.4"))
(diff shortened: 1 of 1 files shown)


hhaim:
rebased.

rjarry:
Thanks!

Issue #850, opened 2022-05-17
clear_profile() may get stuck


bdollma:
The ASTF function:

https://github.com/cisco-system-traffic-generator/trex-core/blob/8a2002b6a0c5a4a8db997aa1da1b7bdbc2fd5aa7/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py#L658

may (need to run multiple times) get stuck in a loop.

Minimal reproducible example:

```python

    try:
        # take all the ports
        c.reset()

        # load ASTF profile
        if not profile_path:
            profile_path = os.path.join(astf_path.get_profiles_path(), 'http_simple.py')

        c.load_profile(profile_path)

        print(c.get_profiles())

        c.clear_profile()

        print(c.get_profiles())
```

Just by starting at the function, the following code does not make sense:

```python
                if block:
                    rc = self._transmit_async('profile_clear', params = params, ok_states = self.STATE_IDLE)
                else:
                    rc = self._transmit('profile_clear', params = params)
```

The async should be the non blocking one.

Originally described here: https://groups.google.com/g/trex-tgn/c/1Ivfqn3Bfao

jsmoon:
@bdollma, I think the following code could solve this issue.

```python
                if block and profile_state is self.STATE_ASTF_LOADED:
                    rc = self._transmit_async('profile_clear', params = params, ok_states = self.STATE_IDLE)
                else:
                    rc = self._transmit('profile_clear', params = params)
```

bdollma:
Hi @jsmoon ,
This doesn't solve the problem, `clear_profile` still gets stuck.
It doesn't make sense that if the `block` parameter is set to `True`, we send an async command. It should be the other way around.

jsmoon:
@bdollma
> This doesn't solve the problem, clear_profile still gets stuck.

I have failed to reproduce the stuck in clear_profile with your example.
So I checked the original thread, in that case, clear_profile is called without a profile loaded.
The suck is reproduced with it well. My suggestion works well for this stuck case.

> It doesn't make sense that if the block parameter is set to True, we send an async command. It should be the other way around.

I cannot agree with this. The block parameter is meaningful only when we need to get the result asynchronous way.
When the profile_state is self.STATE_IDLE, there is no result reported asynchronously.

bdollma:
@jsmoon,
Ok, this is very confusing but I think I am starting to understand what happens here.

[...]

PR #852, opened 2022-05-23
fix clear_profile() may get stuck

jsmoon:
Hi, this change is to fix issue #850.

Since _transmit_async expects that the command would change the profile to the transient state,
when there is no transient state expected for the profile, it should not be called.

Please review my change and give your feedback.

### DIFF
--- a/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
@@ -341,7 +341,7 @@ class ASTFClient(TRexClient):
         else:
             return self.state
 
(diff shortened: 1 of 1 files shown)


jsmoon:
@hhaim could you look into the failure reason?
I just changed a python file but failed to compile the C++ source file.

hhaim:
@jsmoon yes, there was an issue with the build due to the new dpdk

hhaim:
recheck

jsmoon:
@hhaim I'm sorry I deleted the branch abruptly.

bdollma (approved):
Looks good, thanks :)

hhaim:
rebased

Issue #853, opened 2022-05-23
Huge memory consumption from a large amount of event log messages


jsmoon:
Hi, my colleague observed a huge memory usage and high CPU load after massive profiles are loaded/started/stopped.
He observed it after he ran millions of profiles.

He found that the EventHandlers saves the event log messages **implicitly** after handling events from 'get_async_events'.
Since the size of event log messages is unlimited, the memory size of log messages may reach giga-bytes from a few million profiles.
And it takes time to append new log messages because list type is used for saving.

To avoid this situation, the TRex client user should clear the event log messages **explicitly** from time to time.
But the user cannot recognize it easily because it is unknown behavior.
So I think the number of the event log messages should be limited to a certain number, e.g. 1000.

@hhaim, @bdollma, what do you think about this issue?

hhaim:
@jsmoon it sounds like a bug. It you point to the unlimited queue? is it python side or server side?

hhaim:
```
    def on_event (self, event_name, *args, **kwargs):
        """
            trigger an event

            if a handler is registered for 'event_name' it will be called
            and be passed with *args and **kwargs
        """
        if event_name not in self.events_handlers:
            raise TRexError("TRex Events: unhandled event '{0}'".format(event_name))


        event = self.events_handlers[event_name](*args, **kwargs)
        if event:  
            self.events.append(event)   <<<<<

```

are you talking about this?

jsmoon:
@hhaim yes, right.

bdollma:
@jsmoon , I agree that we should limit it. I can see in the code that there are calls to `pop_event` or `get_events` expect for in the client Api. Actually, the optimal data structure for this use case would be a `Queue` (FIFO) so you could `pop` elements very easily. (If we reach the limit, we will have to `pop` for each `append` and that would be nasty on a list).

jsmoon:
@bdollma, `Queue` is a good suggestion. I'll prepare a PR for this issue.

PR #855, opened 2022-05-24
limit the number of saved event log messages

jsmoon:
Hi, this PR is to solve issue #853.

The number of saved event log messages is limited to 1000.
`Queue` is not appropriate to save the event log messages. It does not provide the contents in the middle of the queue.
Instead, `deque` is suitable for it and better than `list`.

@hhaim, @bdollma, please review my change and give your feedback.

### DIFF
--- a/scripts/automation/trex_control_plane/interactive/trex/common/trex_events.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/common/trex_events.py
@@ -8,6 +8,8 @@ from .trex_types import listify
 from .trex_exceptions import TRexError
 from ..utils.text_opts import format_text
 
+from collections import deque
+
 
 # an event
 class Event(object):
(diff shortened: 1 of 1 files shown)


bdollma:
Hi @jsmoon ,
I am sure you have noticed that the code needs to be thread safe.

https://github.com/cisco-system-traffic-generator/trex-core/blob/8a2002b6a0c5a4a8db997aa1da1b7bdbc2fd5aa7/scripts/automation/trex_control_plane/interactive/trex/common/trex_subscriber.py#L231

Even though `pop` and `append` in `deque` are thread safe, other operations might not be. `deque` gets its multi thread safety from GIL, not some syncing mechanism.

`Queue` was created exactly for this reason, a multi producer - multi consumer queue.

I agree that `Queue` can't be indexed, but it can be iterated.

I think `deque` is also very nice (`Queue` uses a `deque` under the hood), but we need to make sure that the code is thread safe and at least add some comment that warns the user about using `self.events`

hhaim:
@jsmoon looks good. @bdollma there is no issue of thread safe.

hhaim:
rebased

PR #857, opened 2022-05-24
ndr: allow writing raw json to a file

rjarry:
The current `--output json` format is not usable. It prints a python string on stdout after debugging information. This makes it impossible to be reused by external tools.

Change the `-o/--output` option to take a file path as argument. If specified, the raw JSON results are dumped into this file.

The results are always printed on stdout in a human readable format.


hhaim:
recheck

hhaim:
rebased

PR #859, opened 2022-05-26
fix client_api get_flow_info does not return flows

jsmoon:
Hi, recently I have added the "get TCP flow information per profile", #841.
But the `get_flow_info` Python client API won't return any flows.
In this PR, I fix it.
In addition, when no additional information is expected, the collected flows will be returned early even if the duration is left.


hhaim:
rebased

PR #864, opened 2022-06-09
silence maybe-uninitialized warning in get_plugin_id()

GabrielGanne:
Hi,

This is just to silence the warning when building with distrib default on debian-11.
gcc is `gcc (Debian 10.2.1-6) 10.2.1 20210110`

Best regards

### DIFF
--- a/src/bp_sim.cpp
+++ b/src/bp_sim.cpp
@@ -1962,20 +1962,6 @@ bool CCapFileFlowInfo::Create(){
 }
 
 
-void CCapFileFlowInfo::dump_pkt_sizes(void){
-    int i;
-    for (i=0; i<(int)Size(); i++) {
-        flow_pkt_info_t lp=GetPacket((uint32_t)i);
-        CGenNode node;
-        node.m_dest_ip  = 0x10000110;
-        node.m_src_ip   = 0x20000110;
-        node.m_src_port = 12;
-        rte_mbuf_t * buf=lp->generate_new_mbuf(&node);
-        //rte_pktmbuf_dump(buf, buf->pkt_len);
-        rte_pktmbuf_free(buf);
-    }
-}
-
 void CCapFileFlowInfo::RemoveAll(){
     int i;
     clear();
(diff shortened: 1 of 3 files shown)


bdollma (commented):
Is this because `m_template_info_` is uninitialized?

If so how does this
```c++
inline bool CGenNode::is_repeat_flow()
```

not provide the same warning?

Maybe just adding an `assert(m_template_info)` might solve the problem instead of code that makes the function unreadable.

In case it is the `plugin_id` that provides the warning, you can simply do `uint8_t         m_plugin_id = 0`;

GabrielGanne:
The warning comes from here which looks like a dead function:
``` C
// src/bp_sim.cpp
void CCapFileFlowInfo::dump_pkt_sizes(void){
    int i;
    for (i=0; i<(int)Size(); i++) {
        flow_pkt_info_t lp=GetPacket((uint32_t)i);
        CGenNode node;
        node.m_dest_ip  = 0x10000110;
        node.m_src_ip   = 0x20000110;
        node.m_src_port = 12;
        rte_mbuf_t * buf=lp->generate_new_mbuf(&node);
        //rte_pktmbuf_dump(buf, buf->pkt_len);
        rte_pktmbuf_free(buf);
    }
}
```
I don't know why I had not seen this first, sorry.

It's been like this since the beginning. So I guess it's not needed anyway ...
Can I just remove that function and its callers instead ?

hhaim:
@GabrielGanne removing code (that is used) is not a solution for maybe-uninitialized issued.

GabrielGanne:
@hhaim as a general statement I agree of course.
However the code I propose to remove here does not do anything, and that also has to count for something, right ?

hhaim:
@GabrielGanne looking again into this, you are right. 
The code is a dead code and as you said it has to be counted for something .

hhaim:
rebased. @GabrielGanne thanks for looking into this

GabrielGanne:
Great, thanks for the review

PR #868, opened 2022-06-14
fix tx queue number config

Li-Xiaoyun:
Use total_tx_queues() as txq number instead of "core number + 2"
since some devices don't support uneven txq and rxq number.

Signed-off-by: Li, Xiaoyun <xiaoyun.li@intel.com>


hhaim:
rebased

PR #880, opened 2022-06-30
update missing trex patch for i40e_ethdev_vf.c

jsmoon:
Hi, this change is to update the missing TREX_PATCH from https://github.com/cisco-system-traffic-generator/trex-core/commit/9e3b836bd10b7367cb770cd8f2e980315f7da4e4

I observed that the `get_pgid_stats` takes a longer time than before.
And I found that there was TREX_PATCH in i40e_ethdev_vf.c which reduces the retry counter.

@hhaim @syaakov please review my change.


egwakim:
Hi @hhaim 
Can you check this patch?

PR #887, opened 2022-07-11
high resolution timestamp in rpc log

jsmoon:
Hi, this change is to get the RPC response time more precisely from the saved RPC log file.


hhaim:
rebased

PR #888, opened 2022-07-13
Fixed typos of "does not exits"

sinaebrahimi:
I just saw an error running one of my scripts in BIRD Stateless API. So I searched through the hole repo to find out if there are any more typos like that. 

Wherever I saw "does not exits", I modified it to "does not exist"!


hhaim:
rebased

Issue #885, opened 2022-07-08
STL get_pgid_stats takes long response time


jsmoon:
Hi, recently I found that STL `get_pgid_stats`'s response time is relatively high; 20 msec when the i40evf driver is used.
It updates port stats counters by calling `g_trex.sync_threads_stats()`.
```C++
COLD_FUNC int TrexDpdkPlatformApi::get_pgid_stats(Json::Value &json, std::vector<uint32_t> pgids) const {
    g_trex.sync_threads_stats();  // <-- why this has been called?
    CFlowStatRuleMgr::instance()->dump_json(json, pgids);
    return 0;
}
```
To get the port stats counters, the i40evf driver asks them to the PF driver, which takes most of the response time.
But, AFAIK, the PGID stats do not need to collect the port stats counters.
Why this has been called?

I think it can be removed without side effects and the response time can be improved.
@hhaim, what do you think of it?

hhaim:
@jsmoon sorry of the late response, I think that in both cases ( software mode and hardware mode) this is not needed. 
as in hardware mode there is a direct API to read the counters and sync them and in software this sync is not needed.

PR #892, opened 2022-07-19
remove irrelevant sync_threads_stats() call

jsmoon:
Hi, this is the change from the suggestion at #885.

I think this change will improve the response time of the `get_pgid_stats` request.


hhaim:
@jsmoon looks good

hhaim:
rebased

PR #902, opened 2022-08-12
use uio_pci_generic.ko to support virtio-net nic

ELWIN-MAO:
support trex running in qemu-kvm vms which use virtio-net nic

Signed-off-by: Mao YingMing <maoyingming@baidu.com>


hhaim:
rebased

PR #904, opened 2022-08-18
fix active flow generation from inactive templates

jsmoon:
Hi, this PR is to remove the flow generation possibility from inactive templates in generate_flow().
When an inactive template is placed before active templates, generate_flow() will generate new flows from the inactive template.
And last active template won't generate new flows. This change will fix this issue.


hhaim:
rebased

PR #906, opened 2022-08-25
improve STL flow latency error counters

jsmoon:
Hi, this change is to count accurately duplicated packets with the STL flow latency feature.
Until now, when the packet drops and duplications happen in the same flow, the error counters cannot show correct values.
It's because the interleaved duplicated packet reduces the drop counter.
With this change, the counters will show accurate values and the users will know the issue more precisely, drop or duplication problem.
But this feature is quite complicated and consumes more CPU and memory even though there is no duplicated packet.
So, I make it optional which can be activated by `--latency-diag` command line argument.
Users can use this feature when some duplicated packets are suspected on DUT.

@hhaim please review this change and give me your feedback.

### DIFF
--- a/src/main_dpdk.cpp
+++ b/src/main_dpdk.cpp
@@ -224,6 +224,7 @@ enum {
        OPT_HDRH,
        OPT_BNXT_SO,
        OPT_DISABLE_IEEE_1588,
+       OPT_LATENCY_DIAG,
 
        /* no more pass this */
        OPT_MAX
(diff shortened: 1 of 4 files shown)


hhaim:
@jsmoon looks good, I would protect the map not to save too many packets in flight (with OOO). Limit to 10k.

jsmoon:
@hhaim, do you mean that `m_seqblks.size() <= 10000`?

hhaim:
@jsmoon yes

hhaim:
rebased

Issue #908, opened 2022-08-29
CPU load does not include some working time.


jsmoon:
Hi, I found that DP and RX CPU utilization report does not include some working time:

DP Core: in `idle_state_loop()`
1. handling time of received packets during the IDLE state is not included: `rx_for_idle()`
2. handling time of RX core messages during the IDLE state is not included.
3. handling time of CP core messages during the IDLE state is not included.

RX Core: in `work_tick()`
1. handling time of DP core messages (LATENCY_PKT) is not included in ASTF mode: `CRxAstfCore::handle_msg_packets()`
2. handling time of CP core messages is not included.
3. handling time of the TPG thread is not included.
4. handling time of the TX queue work tick is not included.
5. handling time of the capture manager work tick is not included.

I'm not sure which time should be included or not.
@hhaim, please give me your opinion.

hhaim:
@jsmoon all the work should be included. It is a bug

PR #912, opened 2022-08-30
Update missing cpu load

jsmoon:
Hi, this change is to fix the issue defined at #908.

I did not add the time for checking if there are some works to do.
So, DP and RX core will report 0 % without work.


hhaim:
rebased

PR #913, opened 2022-08-31
get pgid stats with the update of tx counters.

jsmoon:
Hi, after #892 had merged, my colleague observed more received packets than sent packets during transmitting.
I found the reason is that the TX counters are not updated during the get_pgid_stats command response, unlike the RX counters.
This change will update the TX counters during the response.

In addition, I changed the order of RX and TX counters among ports.
`port 0 (RX --> TX) --> port 1 (RX --> TX) --> ...`
is changed to
`RX (port 0 --> port 1 --> ...) --> TX (port 0 --> port 1 --> ...)`
This will guarantee that TX counters >= RX counters.

@hhaim please review my changes and give your feedback.

### DIFF
--- a/src/main_dpdk.cpp
+++ b/src/main_dpdk.cpp
@@ -4652,6 +4652,20 @@ COLD_FUNC void CGlobalTRex::update_stats(){
 }
 
 COLD_FUNC tx_per_flow_t CGlobalTRex::get_flow_tx_stats(uint8_t port, uint16_t index) {
+    uint8_t port0;
+    CFlowGenListPerThread * lpt;
+
+    m_stats.m_port[port].m_tx_per_flow[index].clear();
+
+    for (int i=0; i < get_cores_tx(); i++) {
+        lpt = m_fl.m_threads_info[i];
+        port0 = lpt->getDualPortId() * 2;
+        if ((port == port0) || (port == port0 + 1)) {
+            m_stats.m_port[port].m_tx_per_flow[index] +=
+                lpt->m_node_gen.m_v_if->get_stats()[port - port0].m_tx_per_flow[index];
+        }
+    }
+
     return m_stats.m_port[port].m_tx_per_flow[index] - m_stats.m_port[port].m_prev_tx_per_flow[index];
 }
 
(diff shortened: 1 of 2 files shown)


src/main_dpdk.cpp
```diff
@@ -4652,6 +4652,20 @@ COLD_FUNC void CGlobalTRex::update_stats(){
 }
 
 COLD_FUNC tx_per_flow_t CGlobalTRex::get_flow_tx_stats(uint8_t port, uint16_t index) {
+    uint8_t port0;
+    CFlowGenListPerThread * lpt;
+
+    m_stats.m_port[port].m_tx_per_flow[index].clear();
```
hhaim, line 4658: You can't clear the counters while the traffic is running ..

jsmoon: No, it must be cleared because it is used to accumulate the saved counters from DP threads.

hhaim: @jsmoon the code work in diff. m_tx_per_flow should go up and accumulate diff from the per port counter. 
the code that you did you did a "clear" to the counters and it does not work as it worked before ..

```
    return m_stats.m_port[port].m_tx_per_flow[index] - m_stats.m_port[port].m_prev_tx_per_flow[index];
```

the ```prev_tx_per_flow``` is updated in clear() 

The interator over the DP cores is done one every 500 msec. You can add an explicit update in this function but you can't clean the counters before

jsmoon: @hhaim, frankly speaking, I had missed the clear during my implementation., and I observed the tx counter fluctuation at TUI. After I added this `clear` routine, the fluctuation disappeared.

`CGlobalTRex::get_stats` updates the counters [here](https://github.com/cisco-system-traffic-generator/trex-core/blob/8916531cc1a3a829dd821449d76061e713eea0fe/src/main_dpdk.cpp#L4822) every 500 msec and it also calls the clear [here](https://github.com/cisco-system-traffic-generator/trex-core/blob/8916531cc1a3a829dd821449d76061e713eea0fe/src/main_dpdk.cpp#L4749) for the update.

hhaim:
@jsmoon you are right, our old implementation is a bit strange and is not aligned to other counters logics we have (all the counters are going up and the cache is done by the python client)

hhaim:
rebased

PR #914, opened 2022-09-06
DPDK bonding device inherits max_mtu from slave devices

jsmoon:
Hi, this PR will fix the following issue when the DPDK bonding device is used.
```
MTU (9710) > device max MTU (9702) for port_id 0
slave_configure(1757) - Cannot configure slave device: port 0, err (-22)
```
Intel NIC's overhead length is 8 bytes larger than others.
So, the calculated MTU size from slave's max_rx_pktlen will be larger than slave's max_mtu and cause a failure like above during slave device configuration.
I thought that the bonding device's max_mtu also should inherit from the slave devices as max_rx_pktlen does.

@hhaim please review my change and give your opinion.


hhaim:
@jsmoon sorry for the delay. Could you just change the advertise max MTU to be max_mtu -8 ? do you mind to have 9702-8 bytes as MTU?
I think it would be a simpler patch to support over time

hhaim:
rebased

PR #916, opened 2022-09-15
ena: use highest link speed

Li-Xiaoyun:
TRex needs valid speed which is 50G in ENA. However, the highest capability of ENA devices is 100G while the generating speed is limited to 50G. So change the link speed to 100G.

Signed-off-by: Xiaoyun Li <xiaoyun.li@intel.com>


hhaim:
rebased

PR #923, opened 2022-10-07
Induce latency in ASTF mode

amith-shetty:


rashikrishna:
So, our use case required sending real world traffic, for which we needed some packets to have more latency than others.

The basic idea is to add a ring buffer before sending the packets to queue.
We added a ring buffer for each port which stores the pointer to the packet mbuf and a timestamp. The timestamp is based on the latency provided by the user. We keep adding to this ring buffer with a timestamp and send the packet mbuf to the queue only when the timestamp expires:
image

So with this, user can give both the latency duration and the percent of packets which will have that latency.
--induce-server-latency <duration_in_microsec> <percent_packet>
duration: the duration of latency for each packet
percent_packet: randomly choose packets, based on this number to induce latency.

We have modified the code in a way that produces minimum effect on the performance.
We have also added `utl_random.h`,  which can later be to expanded to write different types of probability distribution function to choose packets. Users can then be able to give a random function as argument as well!

[latency.pdf](https://github.com/cisco-system-traffic-generator/trex-core/files/9733230/latency.pdf)

PR #925, opened 2022-10-09
IPv6Header: fix FlowLabel field size

raslandarawsheh:
following the RFC6437 [1] flow lable from IPv6 is a 20 bits field, this is setting the correct number of bits for both setter and getter in the ipv6 header

[1] https://www.rfc-editor.org/rfc/rfc6437

Signed-off-by: Raslan Darawsheh <rasland@nvidia.com>

### DIFF
--- a/src/common/Network/Packet/IPv6Header.h
+++ b/src/common/Network/Packet/IPv6Header.h
@@ -205,13 +205,13 @@ IPv6Header::setTrafficClass(uint16_t argTrafficClass)
 //--------------------------------
 inline uint32_t IPv6Header::getFlowLabel()
 {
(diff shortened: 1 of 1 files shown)


raslandarawsheh:
Closing in favor of #1217.

After deeper investigation this change is incorrect. `IPv6Header::getFlowLabel/setFlowLabel` use `getMaskBit32(..., 12, 31)`, which already selects the full 20-bit flow label (width = stopbit - startbit + 1 = 20). Changing the start bit to 20 would *reduce* the field to 12 bits, so this PR would have made things worse, not fixed the reported truncation.

The real cause of the wrong IPv6 flow label is in the flow-stat reserved-IP-ID handling, which applies a 16-bit IPv4-style mask (`& 0x0000efff`) to the 20-bit IPv6 flow label. That is fixed in #1217.

Issue #924, opened 2022-10-07
TCP checksum by software not working


jcpvdm:
Hello, 
I understand the TCP/UDP checksums should be done by software for interfaces that don't support offload since long time ago (#189 ). 
My scenario:
- using STL,
- linux bridge interfaces
- observe the checksums are not fixed at all, they are just copied from the original packet
- Already tried version 3.00 and 2.99. same behavior.
- HTTP traffic from avl/delay_10_http_browsing_0.pcap
- command ./t-rex-64 -f cap2/http_very_long.yaml -m 1 --software (default .yaml)

trex_cfg.yaml
```
- port_limit      : 2
  version         : 2
#List of interfaces. Change to suit your setup. Use ./dpdk_setup_ports.py -s to see available options
  interfaces    : ["p1","p2"]
  low_end: true
  port_info       :  # Port IPs. Change to suit your needs. In case of loopback, you can leave as is.
          - ip         : 1.1.1.1
            default_gw : 2.2.2.2
          - ip         : 2.2.2.2
            default_gw : 1.1.1.1
```

Then played with GDB to investigate.
./t-rex-64-debug-gdb -f cap2/http_very_long.yaml -m 1 --software

Noticed that **tx_offload_csum** was not being called because m->ol_flags=0, therefore not hitting the condition.
`inline void HOT_FUNC tx_burst_offload_csum(struct rte_mbuf **tx_pkts, uint16_t nb_pkts, uint64_t tx_offload) {
        uint16_t nb_tx = 0;

        for (nb_tx = 0; nb_tx < nb_pkts; nb_tx++) {
            struct rte_mbuf *m = tx_pkts[nb_tx];

            if (((m->ol_flags & RTE_MBUF_F_TX_L4_MASK) == RTE_MBUF_F_TX_TCP_CKSUM) ||
                ((m->ol_flags & RTE_MBUF_F_TX_L4_MASK) == RTE_MBUF_F_TX_UDP_CKSUM)) {
                tx_offload_csum(m, tx_offload);
            }
        }
    }`

As desperate attempt, added a breakpoint and modified m->ol_flags in memory using GDB, so it matched RTE_MBUF_F_TX_TCP_CKSUM.
This time tx_offload_csum was called, and the checksums were modified, however checksum was calculated wrong.

Then had a brief look on the code tx_offload_csum. It's unclear to me how the pseudo header is being included in the calculation and if the original packet TCP checksum field is being set to zero initially.

[...]

PR #927, opened 2022-10-09
calculate tcp/udp checksum by software when offload is disabled.

jcpvdm:
Linked to #924.
I didn't touch the original code pointed in the issue as i wasn't sure what was doing.
The location where I introduced the code might not cover all t-rex use cases. Leaving for someone more experienced to review.


hhaim:
@jcpvdm why you are not using ASTF?

PR #931, opened 2022-10-17
Fix Scapy ctypes usage with Python 3.9

vincentbernat:
The embedded version of Scapy is using `find_library("libc")` which is incorrect. This does not work anymore with Python 3.9. This should use `find_library("c")` instead (which also works for older versions). See https://github.com/secdev/scapy/pull/3000 for more details.

Anoher option would be to upgrade Scapy to 2.4.5, where this bug is fixed.

Signed-off-by: Vincent Bernat <vincent@bernat.ch>


hhaim:
rebased

PR #935, opened 2022-10-26
ASTFProgram separates tick_var id from var id.

jsmoon:
Hi, this change is to make the tick vars and vars separate.

In the server, they are separated.
```
    uint64_t                m_vars[apVAR_NUM_SIZE];
    uint64_t                m_tick_vars[apVAR_NUM_SIZE];
```
But they are mixed in the client.
That limits the total number of variables. If one tick var is used, one var can not be used.
This PR will make tick var can be used independently.

@hhaim, please review my change and give your opinion as soon as possible.


egwakim:
Hi @hhaim 
It's urgent issue for us.
Can you please prioritize your review on it?

PR #941, opened 2022-11-07
fix reass_maxqlen range limited by uint8_t.

jsmoon:
Hi, this PR is to fix the `reass_maxqlen` range limited by uint8_t, 255.
It should be supported to 1000 as documented.
So, uint8_t is changed to uint16_t to cover the value of 1000.


hhaim:
rebased

PR #946, opened 2022-11-17
apply XL710 workaround for #508 to bonding

jsmoon:
Hi, this PR is to apply the old XL710 workaround for the #508 issue to the bonding device driver.

Recently, my user observed performance degradation when the bonding device uses Intel 25G NIC with F/W 6.8 as slaves.
I found that a similar issue #508 was already solved by adding an extra queue.
But the workaround was not applied when the device is used for bonding device slaves.

To apply the workaround, I added `m_slave_drv` to the bonding driver and made a proper call for `extra_tx_queues_requires`.

@hhaim please review my change and give your feedback. Any suggestions are welcome.

### DIFF
--- a/src/drivers/trex_driver_base.h
+++ b/src/drivers/trex_driver_base.h
@@ -219,6 +219,8 @@ public:
 
     void create_dummy();
 
+    CTRexExtendedDriverBase * create_driver(std::string name);
+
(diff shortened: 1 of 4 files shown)


hhaim:
rebased

PR #949, opened 2022-11-23
append index parameter to get_flow_info

jsmoon:
Hi, this change is to improve ASTFClient's get_flow_info API.

Since non-blocking requests with duration 0 may contain duplicated entries, the user should compare the duplicated ones and remove them. It's quite an annoying job. So, I added the index parameter for the users of non-blocking requests.

@hhaim please review my changes and give your opinion.


hhaim:
rebased

PR #951, opened 2022-11-25
simulate ftp download speed throttling

jsmoon:
Hi, this change is to control the FTP download speed in `ftp_multiple_flows_sim.py`.

When I used it with multiple flows, each flow tried to get maximum throughput. It prunes to cause some packet drops and bad characteristics in the result.
So, I want to test multiple flows with less throughput to see the improvement of the characteristics.
I added the throughput throttling by speed and I can get good results in some cases; no packet drops and more total throughput.

@hhaim please review my changes and give your feedback. Any further suggestions are welcome.


hhaim:
rebased

PR #936, opened 2022-10-27
ASTF extends STL feature

jsmoon:
Hi, this change is to run a typical stateless profile on ASTF mode.

On this change, there are some restrictions.
- TCP packet is not supported.
- UDP packets should not be handled by ASTF.
- TPG is not supported yet.
- Performance will be low.

```
trex>start_stl -f stl/flow_stats.py

Removing all streams from port(s) [0._, 1._]:                [SUCCESS]


Attaching 1 streams to port(s) [0._]:                        [SUCCESS]


Attaching 1 streams to port(s) [1._]:                        [SUCCESS]


Starting traffic on port(s) [0._, 1._]:                      [SUCCESS]

21.64 [ms]

trex>stats -g
Global Statistics

connection   : localhost, Port 4521                       total_tx_L2  : 1.52 Kbps                      
version      : ASTF @ v3.00                               total_tx_L1  : 1.82 Kbps                      
cpu_util.    : 0.45% @ 1 cores (1 per dual port)          total_rx     : 1.46 Kbps                      
rx_cpu_util. : 0.0% / 0 pps                               total_pps    : 1.88 pps                       
async_util.  : 0% / 1.84 bps                              drop_rate    : 0 bps                          
total_cps.   : 0 cps                                      queue_full   : 0 pkts                         

trex>stats -s
Streams Statistics

  PG ID    |         1         
-----------+------------------
Tx pps     |             0 pps 
Tx bps L2  |             0 bps 
Tx bps L1  |             0 bps 
---        |                   
Rx pps     |             0 pps 
Rx bps     |             0 bps 
----       |                   
opackets   |                48 
ipackets   |                48 
obytes     |              4848 
ibytes     |              4848 
-----      |                   
opackets   |           48 pkts 
ipackets   |           48 pkts 
obytes     |           4.85 KB 
ibytes     |           4.85 KB 
```

@hhaim please review my change and give your opinion.

PR #953, opened 2022-12-07
change get_profile_list response for trex-java-sdk compatibility

jsmoon:
Hi, this PR is to keep backward compatibility with trex-java-sdk behavior.

https://github.com/cisco-system-traffic-generator/trex-java-sdk/blob/c8032462c4be86720bc01f4e1c359c3d803d88e5/src/main/java/com/cisco/trex/stateful/TRexAstfClient.java#L383-L387

In this code, the result from the response should be an array of strings.
So, I added an optional parameter "profiles_in_ports" for backward compatibility.

@hhaim please review my changes and give your feedback quickly.

### DIFF
--- a/src/stx/astf/trex_astf_rpc_cmds.cpp
+++ b/src/stx/astf/trex_astf_rpc_cmds.cpp
@@ -332,29 +332,32 @@ trex_rpc_cmd_rc_e
 TrexRpcCmdAstfProfileList::_run(const Json::Value &params, Json::Value &result) {
     vector<string> profile_list = get_astf_object()->get_profile_id_list();
     Json::Value json_profile_list = Json::arrayValue;
+    bool profiles_in_ports = parse_bool(params, "profiles_in_ports", result, false);
 
     for (auto &profile_id : profile_list) {
         json_profile_list.append(profile_id);
     }
 
-    uint8_t port_cnt = get_astf_object()->get_port_count();
-    for (uint8_t port_id = 0; port_id < port_cnt; port_id++) {
-        TrexStatelessPort *port = get_stateless_obj()->get_port_by_id(port_id);
-        profile_list.clear();
+    if (profiles_in_ports) {
+        Json::Value json_ports = Json::objectValue;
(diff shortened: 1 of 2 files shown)


egwakim:
Hi @hhaim 
It's fix for critical issue.
All trex-java-sdk users cannot use TRex without this fix.
Can you prioritize review on this PR?

src/stx/astf/trex_astf_rpc_cmds.cpp
```diff
@@ -332,29 +332,30 @@ trex_rpc_cmd_rc_e
...
 
     for (auto &profile_id : profile_list) {
         json_profile_list.append(profile_id);
     }
 
-    uint8_t port_cnt = get_astf_object()->get_port_count();
-    for (uint8_t port_id = 0; port_id < port_cnt; port_id++) {
-        TrexStatelessPort *port = get_stateless_obj()->get_port_by_id(port_id);
-        profile_list.clear();
+    if (profiles_in_ports) {
```
hhaim, line 341: I don't see the code in case that profiles_in_ports==False (backward compatible)

jsmoon: @hhaim this change is for trex-java-sdk's backward compatible.
ASTF extends STL feature #936 added the STL profiles in ports by default.
It result of the command changed from [ "_" ] to [ "_", { "\u0000": [ "_" ] }, { "\u0001": [ "_" ] } ].
The Python client doesn't care about the result.
But, trex-java-sdk client treats the result as an array of strings. It causes an exception in the client.
By this change, the result is back to [ "_" ] and the exception will be suppressed.

egwakim:
Hi,
Is there any progress on this commit?

src/stx/astf/trex_astf_rpc_cmds.cpp
hhaim: @jsmoon understood

hhaim:
rebased

Issue #954, opened 2022-12-07
sync_barrier timeout with multiple ASTF profiles


jsmoon:
Hi, recently I've got the following assert error message while multiple ASTF profiles were starting in dual DP cores.
```
assert: ../../../../trex-core/src/utils/utl_sync_barrier.cpp:57 int CSyncBarrier::sync_barrier(uint16_t) Assertion 'm_atomic.cnt!=m_max_ids' failed.
```
According to my investigation, DP core 1 could not enter the TRANSMITTING state by __sync_barrier timeout__.
The reason is that DP core 0 is parsing the following profiles whose size is quite large while DP core 1 is waiting.

I think some solutions for this issue.
1. enable process_at_cp by default when multiple profiles are used.
2. DP core does not continue handling CP messages when the state is changed to a transient state: STARTING, STOPPING.
3. After CP sent the first profile's start_transmit message, CP holds the following CP messages until all DP cores are TRANSMITTING.

@hhaim what do you think about my solutions? which one do you prefer?

hhaim:
@jsmoon as a guideline I would like to keep the single profile as today. (DP is doing the parse to unlimited time and CP is free to get commands) 
in case of multiple profile (after loading the second) I would go with option (3) as long as it won't stuck (async)

PR #955, opened 2022-12-08
ASTF suspends core 0 messages during DP cores starting

jsmoon:
Hi, this PR is to prevent sync_barrier timeout possibility as discussed in issue #954.

During DP cores starting, all DP core 0 messages will be saved in m_suspended_core0_msgs.
They will be sent again later when all DP cores are STATE_TRANSMITTING state.

@hhaim please review my change and give your feedback.


hhaim:
rebased

PR #969, opened 2023-01-23
doc: update OFED installation link

f0rget-the-sad:
Looks like old [OFED installation link](http://www.mellanox.com/page/products_dyn?product_family=26&mtag=linux_sw_drivers) is down and was moved to [new location](https://network.nvidia.com/products/infiniband-drivers/linux/mlnx_ofed). Updated the link.


hhaim:
rebased

Issue #962, opened 2023-01-06
ASTF too many ACKs with TSO


jsmoon:
Hi, I found an issue that TSO does not work as I expected.
I've tested in my loopback environment by `start -f astf/ftp_multiple_flows_sim.py -t fsize=10000`.
```
Cpu Util(%)

  Thread   | Avg | Latest | -1  | -2  | -3  | -4  | -5  | -6  | -7  | -8  | -9  | -10 | -11 | -12 | -13 | -14
-----------+-----+--------+-----+-----+-----+-----+-----+-----+-----+-----+-----+-----+-----+-----+-----+----
 0 (IDLE)  |   4 |      0 |   0 |   0 |   0 |  20 | 100 | 100 | 100 | 100 | 100 |  97 |   0 |   0 |   0 |   0

Traffic stats of Profile ID : _. Number of template groups = 0

                           |      client       |      server       |
---------------------------+-------------------+-------------------+---------------------------
            m_active_flows |                 0 |                 0 | active open flows
               m_est_flows |                 0 |                 0 | active established flows
              m_tx_bw_l7_r |             0 bps |       469.93 Mbps | tx L7 bw acked
        m_tx_bw_l7_total_r |             0 bps |       469.69 Mbps | tx L7 bw total
              m_rx_bw_l7_r |       469.86 Mbps |             0 bps | rx L7 bw acked
                m_tx_pps_r |        40.57 Kpps |        40.53 Kpps | tx pps
                m_rx_pps_r |        40.61 Kpps |        40.58 Kpps | rx pps
                m_avg_size |          723.46 B |          724.21 B | average pkt size
                m_tx_ratio |               0 % |          100.05 % | Tx acked/sent ratio
                         - |                   |                   |
        m_traffic_duration |          7.12 sec |          6.62 sec | measured traffic duration
                         - |                   |                   |
                       TCP |                   |                   |
                         - |                   |                   |
          tcps_connattempt |                 1 |                 0 | connections initiated
              tcps_accepts |                 0 |                 1 | connections accepted
             tcps_connects |                 1 |                 1 | connections established
               tcps_closed |                 1 |                 1 | conn. closed (includes drops)
            tcps_segstimed |                 3 |           7222290 | segs where we tried to get rtt
           tcps_rttupdated |                 3 |           7227987 | times we succeeded
             tcps_sndtotal |           7227988 |           7222444 | total packets sent
[...]

PR #970, opened 2023-01-25
Prevent the send sockbuf from being emptied by TSO

jsmoon:
Hi, as discussed at #962, this change prevents TSO packets from adding the PUSH flag.
Since the PUSH flag is added when the send sockbuf is emptied, with this change, TSO left some data always and the PUSH flag will not be added to the packets.

@hhaim please review my change and give your feedback if any.


hhaim:
rebased

PR #972, opened 2023-01-30
SLRO merges consecutive ACK-only packets

jsmoon:
Hi, this PR is the deferring ACK implementation.

The most significant ACK number in a loop will be delivered to tcp_input() routine.
As discussed in #962, this may expand the TSO packet size.


jsmoon:
@hhaim, this PR has the side effects of performance degradation.
image
I think it is due to the reduced number of ACK packets.
Since this PR is to cover a special use case (tcp.no_delay=2), I think it is better not to merge this PR.

Issue #975, opened 2023-02-09
Captured TSO packet is not dissected by Wireshark.


jsmoon:
Recently I captured TSO packets but Wireshark does not dissect the packet properly.
![image](https://user-images.githubusercontent.com/4134762/217780734-1f5e3a21-457f-4d94-9f76-4f78b5d73558.png)
It is under the "Support packet-capture from IP TSO-enabled hardware" is enabled.
On the web, I found a comment the IP total length should be 0 to dissect the TSO packet properly.
The following source code in trex-core set it as 20.
https://github.com/cisco-system-traffic-generator/trex-core/blob/8a2002b6a0c5a4a8db997aa1da1b7bdbc2fd5aa7/src/44bsd/tcp_output.cpp#L112-L114
When I changed it to 0, I can see the TCP contents properly.
![image](https://user-images.githubusercontent.com/4134762/217784204-cb82d1c5-b4fd-455d-83bb-b738db4017fe.png)

@hhaim do you know why the 20 (maybe IPv4 header length) is used?

hhaim:
@jsmoon this is how the hardware except the packet. ( the ip.cs should be calculated and hardware is adding the difference).
I would capture the traffic after the TSO ( in the other size, rx side)

hhaim:
@jsmoon in the rx side, the ip.totla_len should be the right one (not zero).

jsmoon:
@hhaim in the rx side, the received packet's ip.total_len is not zero and the checksum is correct.
This is captured packets in the rx side.
![image](https://user-images.githubusercontent.com/4134762/217952828-d231ffd2-3b2f-48b5-8131-a2bc8d59e98b.png)
Surely, the statistics counters are normal.
I found no issues.

hhaim:
@jsmoon so there is no issue.

PR #976, opened 2023-02-14
enable TSO packet dissection of tools

jsmoon:
Hi, as discussed at #975, I prepared this PR which enables TSO packet dissection of tools.
This change allows Wireshark and tcpdump to show TCP header information in the captured TSO packet.

@hhaim please review my change and give your opinion.

### DIFF
--- a/src/44bsd/tcp_output.cpp
+++ b/src/44bsd/tcp_output.cpp
@@ -110,7 +110,7 @@ static inline void tcp_pkt_update_len(CFlowTemplate *ftp,
             }
             if (tso_done){
                 /* in case of TSO the len is auto calculated */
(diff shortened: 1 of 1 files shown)


jsmoon:
@hhaim, already 3 weeks passed. How long should I wait for more to get your comment?

hhaim:
@jsmoon sorry for the delay. Did you test this on a setup that has TSO ( XL710)? the pseudo checksum calculation should be fixed too with the same header. I see fix only in one place instead of two.

hhaim:
@jsmoon looking again into the code, the PR is OK.

hhaim:
rebased

PR #977, opened 2023-02-14
SLRO merges PSH pkt to saved mbuf and pushes it

jsmoon:
Hi, this change will merge a PSH packet to the previously merged packet by SLRO.
With this change, more packets can be merged and fewer ACK packets can be shown.
```
ACK ACK ACK PSH ACK PSH --(SLRO)--> ACK PSH ACK PSH
=>
ACK ACK ACK PSH ACK PSH --(SLRO)--> PSH PSH
```
PSH means the packet has PSH and ACK flags, and ACK means the packet has only the ACK flag.


hhaim:
rebased

Issue #979, opened 2023-02-22
I cannot access https://trex-tgn.cisco.com/youtrack/issue/trex-537


jsmoon:
Hi, the comment in commit 48a4cc736a tells me to see https://trex-tgn.cisco.com/youtrack/issue/trex-537, but I cannot access the page. How could I get the content?

I'm trying to use the mbuf-const to reduce the usage of mbuf. But I get an error when the mbuf-const is freed and then used again.
If I don't free the mbuf-const, can I use it safely? Why the mbuf-const is disabled for ASTF interactive mode?

jsmoon:
@hhaim, thank you for your quick response.
If I understand it correctly, I can use the const-mbuf for my change.

hhaim:
@jsmoon you **can't** use the const-mbuf if you are not syncing with the driver tx- queue (there is no simple way to do this). 
I looked into the PR, not sure I understand how it solves the issue or what is the issue?
In the old ASTF **none-interactive** there was no need to sync with a new profile (all the mbuf would return after program exist).

The performance penalty is low so I still do not understand what you are trying to solve?

jsmoon:
@hhaim from your e-mail,
> To implement that we can add a cache for const mbuf that won't return to the main pool.

I implemented a cache `mbufs_cached` and the const-mbuf in `mbufs_cached` will never return to the main pool. 

As you know that my colleague runs lots of profiles, and lots of 2K mbufs are consumed by redundant ones.
I would like to reduce 2K mbuf consumption by minimizing the redundant mbufs.

hhaim:
@jsmoon I've answered this in the PR

PR #981, opened 2023-02-23
cache filled mbuf to reduce 2K mbuf usage

jsmoon:
Hi, this change reduces 2K mbuf usage when lots of profiles and buffers are loaded.
Until now, a buffer in a profile shares a mbuf filled by pattern. The shared mbuf is dedicated to one buffer only.
From now on, the shared mbuf will be cached globally and can be used by other buffers even though in other profiles.

I implemented the cache using `const-mbuf` due to the possibility of refcnt overflow.
As discussed at #979, `const-mbuf` will not be freed from the cache.
I think there is no issue because the number of cached mbufs will be small.

@hhaim, please review this change and give your feedback.

### DIFF
--- a/src/44bsd/tcp_socket.cpp
+++ b/src/44bsd/tcp_socket.cpp
@@ -1100,19 +1100,54 @@ int utl_mbuf_buffer_create_and_copy(uint8_t socket,
 }
 
 
+/*
+ * The mbuf filled with patterns can be cached for other buffers.
+ * It will reduce 2K mbuf usage when lots of buffers are loaded.
+ * To avoid the 16bit refcnt overflow, const-mbuf is used.
+ * The const-mbuf could not be freed because it will stay in the driver's TX descriptor.
+ * In general, the number of cached mbufs is small and has no issue.
+ */
+#define MAX_MBUF_CACHE  32
+static std::map<std::string,rte_mbuf_t*> mbufs_cached[MAX_SOCKETS_SUPPORTED];
+static std::mutex mbufs_cached_mutex[MAX_SOCKETS_SUPPORTED];
+
+static rte_mbuf_t* get_mbuf_cached(uint8_t socket, std::string& fstring) {
+    if (CGlobalInfo::m_do_mbuf_cache) {
+        std::unique_lock<std::mutex> my_lock(mbufs_cached_mutex[socket]);
+
+        auto it = mbufs_cached[socket].find(fstring);
+        if (it != mbufs_cached[socket].end()) {
+            return it->second;
+        }
+    }
+    return nullptr;
+}
+
+static void set_mbuf_cached(uint8_t socket, std::string& fstring, rte_mbuf_t* mbuf) {
+    if (CGlobalInfo::m_do_mbuf_cache && (mbufs_cached[socket].size() < MAX_MBUF_CACHE)) {
+        std::unique_lock<std::mutex> my_lock(mbufs_cached_mutex[socket]);
+
+        if (mbufs_cached[socket].find(fstring) == mbufs_cached[socket].end()) {
+            mbufs_cached[socket][fstring] = mbuf;
+            rte_mbuf_set_as_core_const(mbuf);
+        }
+    }
+}
+
+
 int utl_mbuf_buffer_create_and_copy(uint8_t socket,
                                     CMbufBuffer * buf,
                                     uint32_t blk_size,
                                     uint8_t *d,
                                     uint32_t d_size,
                                     uint32_t size,
-                                    uint8_t *f,
-                                    uint32_t f_size,
+                                    std::string& fill_string,
                                     bool mbuf_const){
 
     buf->Create(blk_size);
     uint8_t cnt = 0;
-    rte_mbuf_t* mbuf_fill = nullptr;
+    rte_mbuf_t* mbuf_fill = get_mbuf_cached(socket, fill_string);
+
     while (size>0) {
         uint32_t alloc_size = bsd_umin(blk_size,size);
         auto copy_size = bsd_umin(d_size, alloc_size);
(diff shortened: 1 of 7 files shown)


hhaim:
@jsmoon could you give the numbers of 2K mbufs with a big profiles with and without this feature?
I would be happy if there would be a tunable that enable this optimization(cache) explicitly.

jsmoon:
@hhaim I added "do_mbuf_cache" tunable to global_cfg trex-console command.

212,845 mbufs used for running 13,053 profiles ==> 37,173 mbufs used for running 20,445 profiles
The 2K mbufs are saved about 89% with this change.

src/44bsd/tcp_socket.cpp
```diff
@@ -1100,19 +1100,48 @@ int utl_mbuf_buffer_create_and_copy(uint8_t socket,
...
                                     uint32_t size,
-                                    uint8_t *f,
-                                    uint32_t f_size,
+                                    std::string& fill_string,
                                     bool mbuf_const){
 
     buf->Create(blk_size);
     uint8_t cnt = 0;
-    rte_mbuf_t* mbuf_fill = nullptr;
+    rte_mbuf_t* mbuf_fill = get_mbuf_cached(socket, fill_string);
```
hhaim, line 1143: if it was found, it is the same string in the mbuf and you can skip the rest of the function, you could return here

src/44bsd/tcp_socket.cpp
```diff
@@ -1100,19 +1100,48 @@ int utl_mbuf_buffer_create_and_copy(uint8_t socket,
...
 
 
+/*
+ * The mbuf filled with patterns can be cached for other buffers.
+ * It will reduce 2K mbuf usage when lots of buffers are loaded.
+ * To avoid the 16bit refcnt overflow, const-mbuf is used.
+ * The const-mbuf could not be freed because it will stay in the driver's TX descriptor.
+ * In general, the number of cached mbufs is small and has no issue.
+ */
+static std::map<std::string,rte_mbuf_t*> mbufs_cached[MAX_SOCKETS_SUPPORTED];
```
hhaim, line 1110: There is a leakage of mbufs in case you are using the cache. Where this map is clean and return the mbufs to the pools? 
more other in case that you move from enable -> disable there is no process that free the mbufs

src/astf/astf_proto.h
```diff
@@ -0,0 +1,56 @@
+#ifndef ASTF_PROTO_H
```
hhaim, line 1: why do you need this file?

hhaim (commented):
The saving is significant, but there is a need to clean the cache when there is no traffic and no profiles (explict) -- after the driver finished with the mbufs

src/astf/astf_proto.h
jsmoon: I'm sorry it is added abruptly. I'll remove it.

src/44bsd/tcp_socket.cpp
jsmoon: All the mbufs in mbufs_cached are const-mbuf. It should not be freed. It can be done safely at shutdown time.

hhaim: @jsmoon so if you are using interactive mode, you don't have a way to clean the cache?  will it stuck there forever? 
I think we need to give a way to clean the cache or disable this mode in interactive.

hhaim:
@jsmoon another thing is to document this feature

src/44bsd/tcp_socket.cpp
hhaim: @jsmoon right, you are correct. The return object is of another type.

jsmoon:
> @jsmoon another thing is to document this feature

About global_cfg tunable?

hhaim:
@jsmoon yes.

jsmoon:
@hhaim O.K. I'll prepare for it. But I think it takes time.

Anyway, you want to clean up the const-mbuf in mbuf_cached, right?

jsmoon:
But as I mentioned in the first comment, I think there is no issue.

> As discussed at #979, `const-mbuf` will not be freed from the cache.
I think there is no issue because the number of cached mbufs will be small.

hhaim:
> @hhaim O.K. I'll prepare for it. But I think it takes time.
> 
> Anyway, you want to clean up the const-mbuf in mbuf_cached, right?

In case of ASTF interactive mode, the server would be up forever, so this mode will create a leakage without a workaround way to clean it (when all the conditions are met). Are you working in interactive on none-interactive mode?

jsmoon:
@hhaim I'm working in interactive mode.

> this mode will create a leakage without a workaround way to clean it

But I don't think it a leakage. As far as I know, the leakage means the mbuf cannot be used any more.
The cached mbuf, although it is not used by current running profiles, can be used anytime by any profile.
It can hardly be called leakage.

In general, the entries in mbuf_cached will be { "", "*" } with the current typical usage.
If you are using random fill_string, definitely the number of entries are increased and consume more mbufs as const-mbuf.
I think this is a kind of special use case. In this case, the user should know it and do it with do_mbuf_cache disabled.

You are concerning too many const-mbuf can be used; I can limit the total number of entries in mbuf_cached.

As you know that, the const-mbuf cannot be returned to the main pool.
So, a separated const-mbuf pool is needed to manage them and the allocation code should be changed to know  in advance whether const-mbuf is needed or not.

@hhaim, which one do you want me to do in this PR?
1. current change is enough
2. limit the total number of cache entries
3. manage a separate const-mbuf pool

hhaim:
@jsmoon looks good. I would go with limiting the number of cache (2) in case this feature is used. this way you won't exhaust all the 2K mbuf of the system

jsmoon:
@hhaim I've limited the number of cached mbuf up to 32.

hhaim:
rebased

PR #983, opened 2023-02-27
Increase get_link_speed maximum returned value

HassanKhadour:
Increase get_link_speed maximum returned value
to be 200G


HassanKhadour:
@hhaim can you please review the failure?

Thanks,
Hassan

hhaim:
@HassanKhadour thanks for the PR. the PR does not compile have a look into the build output to fix this

Issue #987, opened 2023-03-07
tui Exception in thread Thread-5


digger-yu:
```
Exiting TUI...

trex>tui
Exception in thread Thread-5:
Traceback (most recent call last):
  File "/usr/lib/python3.9/threading.py", line 980, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.9/threading.py", line 917, in run
    self._target(*self._args, **self._kwargs)
  File "/trex-core/scripts/automation/trex_control_plane/interactive/trex/console/trex_tui.py", line 646, in __handler
    self.__redraw()
  File "/trex-core/scripts/automation/trex_control_plane/interactive/trex/console/trex_tui.py", line 654, in __redraw
    self.redraw_cb(buffer)
  File "/trex-core/scripts/automation/trex_control_plane/interactive/trex/console/trex_tui.py", line 725, in redraw_handler
    self.pm.show(show_legend = self.async_keys.is_legend_mode(), buffer = buffer)
  File "/trex-core/scripts/automation/trex_control_plane/interactive/trex/console/trex_tui.py", line 524, in show
    self.main_panel.show(buffer)
  File "/trex-core/scripts/automation/trex_control_plane/interactive/trex/console/trex_tui.py", line 251, in show
    self.client._show_latency_stats(buffer = buffer)
AttributeError: 'STLClient' object has no attribute '_show_latency_stats'

```
```
root@server:/trex-core/scripts# cat ../VERSION
v3.02
```
when i in trex-console ,enter  tui , 'l' - latency,
Error encountered

jsmoon:
Hi, it was changed abruptly when I added STLClient to ASTFClient.
I'll prepare a PR for this.

PR #991, opened 2023-03-08
Fix TUI exception from missing _show_stats_latency in STLClient

jsmoon:
Hi, this PR is to fix the issue reported at #987.

And I added STL streams and the latency panel to the ASTF TUI.


hhaim:
rebased

Issue #1003, opened 2023-03-09
link error in trex_vm_manual.asciidoc


digger-yu:
https://trex-tgn.cisco.com/trex/doc/trex_vm_manual.html

![image](https://user-images.githubusercontent.com/55081697/223970086-68c48790-0cf2-4813-909a-314cfd64b8e3.png)
The link cannot be accessed

```
[NOTE]
See http://trex-tgn.cisco.com/trex/doc/trex_book.pdf[TRex full manual] for a complete understading of the tool features and options.
```

i see the document source code ,If I know the correct link, it should be easy to fix

is this link right  https://trex-tgn.cisco.com/trex/doc/trex_manual.html  ?

hhaim:
@digger-yu yes, you can use relative path instead of a full path

PR #1005, opened 2023-03-09
fix bug #1003

digger-yu:
fix trex_faq.asciidoc digital errors  and bug #1003


hhaim:
this is not the right way, this is how to do it in relative way 

```see http://trex-tgn.cisco.com/trex/doc/trex_book.pdf[TRex full manual]```

==> 

``` see link:trex_book.pdf[TRex full manual]```

digger-yu:
Fix the format error. It should be OK this time. Thank you for your correction

PR #1011, opened 2023-03-22
clear stats available in trex-console read-only mode

jsmoon:
Hi, this PR is to improve trex-console behavior.

Until now, statistics cannot be cleared in trex-console read-only mode.
It's because the acquired ports are required for the default action.
But, since the clearing statistics does not change the T-Rex server, there is no need to acquire ports to do it.
So, I changed the acquired ports to all ports for clearing stats.

Also, I have found a bug that calls clear_dynamic_stats with an invalid argument and fixed it.

@hhaim please review my change.

### DIFF
--- a/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
@@ -1194,7 +1194,7 @@ class ASTFClient(STLClient):
                     clear all the profiles sts
         """
         if clear_all:
-            self.traffic_stats.clear_dynamic_stats(clear_all)
+            self.traffic_stats.clear_dynamic_stats(clear_all = True)
         if pid_input:
             self.traffic_stats.clear_dynamic_stats(pid_input)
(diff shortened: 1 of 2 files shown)


scripts/automation/trex_control_plane/interactive/trex/astf/trex_astf_client.py
```diff
@@ -1194,7 +1194,7 @@ def clear_dynamic_traffic_stats(self, pid_input = None, clear_all = False):
                     clear all the profiles sts
         """
         if clear_all:
-            self.traffic_stats.clear_dynamic_stats(clear_all)
+            self.traffic_stats.clear_dynamic_stats(clear_all = True)
```
hhaim, line 1197: @jsmoon why do you need this change? clear_all is True

jsmoon: @hhaim I've changed the positional argument to the keyword argument. The first positional argument should be `pid_input`, not `clear_all`.

hhaim:
rebased

PR #1015, opened 2023-03-27
set device mtu by port_mtu in trex_cfg.yaml

jsmoon:
Hi, this change overrides the device's MTU size by port_mtu in trex_cfg.yaml.

My colleague found the bonding driver had failed to set the MTU size when the slave device is the MLX5.
Since the MLX5 DPDK driver returns 65536 for max_rx_pktlen and T-Rex calculates the RX MTU size from the max_rx_pktlen, the bonding driver tries to set the large RX MTU size and failed.
https://github.com/cisco-system-traffic-generator/trex-core/blob/7216782dc312333a7c66fd10644f07eed32edb17/src/dpdk/drivers/net/mlx5/mlx5_ethdev.c#L321-L324
I tried to set RTE_ETHER_MAX_JUMBO_FRAME_LEN but it was failed also. So, I decided to set it from the already defined configuration value, port_mtu.

@hhaim, please review my change and give your opinion.


hhaim:
rebased

Issue #608, opened 2021-01-08
ASTF udp latency feature


jsmoon:
Hi, I need to measure UDP packet delivery latency in ASTF.

I know the latency feature exists already for the network via ICMP.
But I need the latency feature for the UDP packets. (DUT may show different characteristics to the different flow.)
It requires the statistics of min/max/average latency and counter of delayed packets over given latency time.

In my thought, there are two discussion points:
1. How to trigger the latency measurement operation
2. How to save the statistics and provide them to user

@hhaim, do you have any idea or preference regarding this feature?

hhaim:
@jsmoon what about TCP latency statistic and TCP latency per application?. 
For UDP the idea is to create another layer of application on top of the socket instructions. So the user would be able to create RTP applications with specific characteristic  (video,audio, rate). the logic how RTP works will be implemented in the server (both side of the app, client and server)./
The server side will be able to collect the statistic (jitter,latency) to a general infra. The statistic would be global per application (RTP/tpid).

jsmoon:
@hhaim thank you for your reply.
For TCP, I don't need it currently. I think it can be different from UDP. So I'd like to talk about it after UDP latency has been done.
For UDP, your idea will be a more general feature for the user. I agreed.
Could you elaborate on your thought about the layer?

hhaim:
@jsmoon this layer should be written as an application layer on top of the layer today. 
Each application (e.g. RTP) would this:

* Plug a client/server counters, pool, that will be accumulated from all cores
* Add DP application instructions to the instruction set for example for RTP there would be an instruction start_rtp(type="audio", rate=100kpbs), in the server start_rtp_collect(counter=true)
* Add RPC layer to parse the new json instructions

For HTTP for example, there would be http_parse () to parse the headers and make sure the length is OK, in the client side there would be a way to check the response (200 OK)

jsmoon:
@hhaim I think I missed one point yesterday.
Actually, I need to measure latency without the RTP header also.
Could it be covered by your idea also?

hhaim:
@jsmoon the latency information should be somehow be in the payload, better to use a well known header of a real application than inventing a new one. 
In my model you can define a new dummy application with a new header.

jsmoon:
@hhaim, good news.
But I think I still don't figure out your model exactly. What did you mean by `application layer`?

hhaim:
@jsmoon it is just a model that wasn't implemented yet. 
[...]

PR #1021, opened 2023-04-19
ASTF UDP latency with protocol add-on feature

jsmoon:
Hi, this PR is an implementation of issue #608 suggestion.

I added a protocol add-on feature on top of the UDP flow and implemented a simple UDP latency addon.

```
trex>start -f astf/udp_latency.py -d 10
trex>template_group stats --name latency
Profile ID : _. Template Group Name: latency. Number of template groups = 1

                           |      client       |      server       |                            
---------------------------+-------------------+-------------------+---------------------------
            m_active_flows |                 0 |                 0 | active open flows          
               m_est_flows |                 0 |                 0 | active established flows   
              m_tx_bw_l7_r |         5.82 Kbps |         3.12 Kbps | tx L7 bw acked             
        m_tx_bw_l7_total_r |         5.82 Kbps |         3.12 Kbps | tx L7 bw total             
              m_rx_bw_l7_r |         3.12 Kbps |         5.82 Kbps | rx L7 bw acked             
                m_tx_pps_r |          2.79 pps |          2.79 pps | tx pps                     
                m_rx_pps_r |          2.79 pps |          2.79 pps | rx pps                     
                m_avg_size |           200.5 B |           200.5 B | average pkt size           
                m_tx_ratio |             100 % |             100 % | Tx acked/sent ratio        
        m_traffic_duration |            10 sec |           9.9 sec | measured traffic duration  
              udps_accepts |                 0 |               100 | connections accepted       
             udps_connects |                10 |                 0 | connections established    
               udps_closed |                10 |               100 | conn. closed (includes drops) 
              udps_sndbyte |             26100 |             14000 | data bytes transmitted     
               udps_sndpkt |               100 |               100 | data packets transmitted   
              udps_rcvbyte |             14000 |             26100 | data bytes received        
               udps_rcvpkt |               100 |               100 | data packets received      
               Application |                   |                   |                            
                   latency |                   |                   |                            
                  min_time |                31 |                14 | minimum latency time (usec) 
                  max_time |                53 |                22 | maximum latency time (usec) 
               measure_cnt |                 1 |                 1 | number of accumulated min/max time 
                total_time |              3987 |              2001 | total sum of latency time (usec) 
                total_pkts |               100 |               100 | total received latency packets 
```

@hhaim please review my changes and give your opinion.

### DIFF
--- a/doc/trex_astf.asciidoc
+++ b/doc/trex_astf.asciidoc
@@ -2160,6 +2160,34 @@ ASTFAssociationRule(port=80, ip_start="48.0.0.0", ip_end="48.0.0.255",
                     l7_map=[0,1,2,3])                                   # for template 2
 ----
 
+==== Tutorial: UDP latency emulation with protcol add-on 
+
+You can measure latency time in micro-second unit on top of the UDP flow.
+
+*File*:: link:{github_astf_path}/udp_latency.py[astf/udp_latency.py]
+
+It can be simply activated by specifying `'latency'` to the UDP 'ASTFProgram'.
+
+[source,python]
+----
+ASTFProgram(stream=False, addon='latency')
+----
+
+The UDP latency addon will insert a latency header between the UDP header and UDP payload given by send_msg. So, udps_sendbyte and udps_rcvbyte will be larger than the payload size.
+Since the latency statistics is reported per template group only, you should specify tg_name to get the statistics counters.
+
+[source,python]
+----
+ASTFTemplate(client_template=temp_c, server_template=temp_s, tg_name='latency')
+----
+
+You can use template group API to get the counters. In trex-console, following command will show the latency counters.
+
+[source,bash]
+----
+[trex]>template_group stats --name latency
+----
+
 ==== Black list of MAC and IPv4 addresses
 
 Until version v2.97, in case the received packets are not destined for TRex server, TRex response depends on the tunable tcp.blackhole value. +
--- a/linux_dpdk/ws_main.py
+++ b/linux_dpdk/ws_main.py
@@ -960,6 +960,8 @@ main_src = SrcGroup(dir='src',
              '44bsd/tick_cmd_clock.cpp',
              '44bsd/udp.cpp',
 
+             '44bsd/udp_latency.cpp',
+
              'tunnels/gtp_man.cpp',
              'tunnels/tunnel_db.cpp',
 
--- a/scripts/astf/udp_latency.py
+++ b/scripts/astf/udp_latency.py
@@ -1,0 +1,58 @@
+# Example for creating your program by specifying buffers to send, without relaying on pcap file
+
+from trex.astf.api import *
+import argparse
+
+
+# we can send either Python bytes type as below:
+http_req = b'GET /3384 HTTP/1.1\r\nHost: 22.0.0.3\r\nConnection: Keep-Alive\r\nUser-Agent: Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 5.1; SV1; .NET CLR 1.1.4322; .NET CLR 2.0.50727)\r\nAccept: */*\r\nAccept-Language: en-us\r\nAccept-Encoding: gzip, deflate, compress\r\n\r\n'
+# or we can send Python string containing ascii chars, as below:
+http_response = 'HTTP/1.1 200 OK\r\nServer: Microsoft-IIS/6.0\r\nContent-Type: text/html\r\nContent-Length: 32000\r\n\r\n<html><pre>**********</pre></html>'
+
+class Prof1():
+    def __init__(self):
+        pass  # tunables
+
+    def create_profile(self, interval, attempts):
+        # client commands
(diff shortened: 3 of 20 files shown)


jsmoon:
Since the changes for the protocol add-on feature are quite large, I'll write the overview of the changes for your review.

### `CEmulAddon` class is introduced to handle registered protocol add-ons.
  - `CEmulAddon::m_addon_list` keeps the registered protocol add-ons.
    - The protocol add-on can be registered with the operations defined by `struct addon_op`.
  - `CEmulApp` in `CFlowBase` keeps a reference to the selected protocol add-on, a `CEmulAddon` instance.
    - It supports some helper methods for the protocol add-on operations.

### `CAddonStats` class is introduced to keep the statistics counters for the protocol add-ons.
  - The add-on statistics counters are allocated per template group.
  - `CEmulAddon::get_stats_desc()` returns the description of the counters.
  - The statistics counters will be reported in the result of `"get_tg_id_stats"` request.
    - `CSTTCpPerTGIDPerDir::m_dyn_sts` per template group is used for this. It was unused when it is for the template group.
    - The description will be reported in the result of `"get_tg_names"` request, `"tg_addon_desc"`.
    - The client concatenates the global desc and tg_addon_desc to get counter names. 

### The `connection_id` of the protocol add-on is considered.
  - The protocol add-on also should distinguish connections in the flow. The connection ID can be used to identify the server template like `l7_map` feature.
    - The protocol add-on and `l7_map` cannot be used for the same port and IP range.
  - The connection ID for the server template will be assigned and it will be propagated to the same template in the client.
    - `CEmulApp::m_peer_id` contains the connection ID and the protocol add-on operations can refer to it.
  - `assoc_id` is introduced to `ASTFAssociationRule` for the server-only usage case.

### Changes for the protocol add-on (UDP latency) implementation
  - `CFlowTemplate` provides its tail room for the protocol add-on header.
  - `CEmulApp` provides `m_priv_data` to save the flow-dependent data pointer.
  - `CUdpFlow` provides `flush_tx()` method to remove waiting time.

### Changes for future works
  - `CEmulAppProgram::m_stream` does not represent UDP or TCP flow when the `addon` is specified. It will be used for the program itself; datagram or stream.
    - When the `addon` is a stream protocol, it can be `true` and the stream command can be used but the flow can be `CUdpFlow`.

@hhaim, if you are still unclear, please let me know it. I'll prepare a detailed description of it.

hhaim:
@jsmoon l've started to look into this and it looks good and in the right direction. It would be very easy with this code to add more emu-plugins logic 

One question: 

1. Why do you need the plugin to work on **all** the template data? Would it be enogth to add **sample** option? (e.g. 1/1000 flows to add the timestamp)

jsmoon:
@hhaim Since I've made it work with the shared port mechanism, it can be accomplished by adding another template.

hhaim:
@jsmoon I've reviewed the code, it was much bigger than I had imagined. You even used the dynamic counters mechanism which is nice. 

a few minor comments 


1. 
```
struct addon_op {
    uint64_t (*new_connection_id)(uint64_t ref_value);
    bool (*get_connection_id)(uint8_t* data, uint16_t len, uint64_t* conn_id);
    void (*setup)(CEmulApp* app);
    void (*close)(CEmulApp* app);
    void (*send_data)(CEmulApp* app, CMbufBuffer* buf);
    void (*recv_data)(CEmulApp* app, struct rte_mbuf* mbuf);
    sts_desc_t* (*get_stats_desc)();
    uint64_t* (*get_stats_data)(CEmulApp* app);
}; 
``` 
this model is for kernel driver in C why not to use C++ virtual interface instead?

2. There is an assumtion that the first 128 bytes of the packet (that is RW) in all cases (e.g. ip6+tunnel+plugin_addon_data <128) is enogth for all the "emu_plugins",  I would add a virtual function to plugin that return the maximum add_on_data and check it before you load the plugin 


```
    return flow->m_template.reserve_addon(sizeof(struct latency_header));
```

else you would have a crash

jsmoon:
@hhaim 
1. I agree that the C++ virtual function could be better. I've considered the dynamic plugin mechanism at the beginning. Since I was not sure it can be possible with the C++ virtual function, I implemented it as a C-style interface.
2. I agree that it can be crashed. I will prevent the case.

hhaim:
@jsmoon regarding 1, just register a factory function that returns an object that implement the virtual function 
(factory pattern)

jsmoon:
@hhaim 
1. I changed it using by C++ virtual functions. But the factory pattern is not used because there is no need to return additional objects for the registered addon.
2. I made it work with the allocated memory. Now, the `reserve_addon` function won't return nullptr. FYI, the tunnel data does not locate in the m_template_pkt[128].

jsmoon:
@hhaim, is anything else required?

hhaim:
@jsmoon looked good. I'm OOO this week, will do it next week

jsmoon:
@hhaim one more week passed now. Are you still OOO?

hhaim:
@jsmoon 
rebased. sorry for the delay

PR #1023, opened 2023-04-28
enable ECN packet handling

jsmoon:
Hi, this PR is to enable ECN packet handling in the integrated TCP stack.

When I integrated FreeBSD 13.0 TCP stack, the ECN packet handling was left over. Recently it needs to be enabled. This change delivers ECN bits in the received IP packet header to the TCP stack.

@hhaim please review and give your feedback.

### DIFF
--- a/doc/trex_astf.asciidoc
+++ b/doc/trex_astf.asciidoc
@@ -3881,6 +3881,7 @@ link:cp_astf_docs/index.html[python index]
 |   tcp.do_rfc1323  | bool   |1|-|+| enable timestamp rfc 1323. 
 |   tcp.do_sack     | bool   |1|-|+| enable SACK.
 |   tcp.cc_algo     | 0,1    |0|-|+| select congestion control algorithm (0 - newreno, 1 - cubic).
+|   tcp.do_ecn      | 0,1,2  |0|-|+| enable ECN (1 - active, 2 - passive). limitation can't be used with hardware filter on some NICs.
 |   tcp.reass_maxqlen | 0,1000 |100|-|+| maximum number of TCP segments per reassembly queue.
 |   tcp.keepinit    |2-65533|5|-|+| value in second for TCP keepalive.
 |   tcp.keepidle    |2-65533|5|-|+| value in second for TCP keepidle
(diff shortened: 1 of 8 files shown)


hhaim:
@jsmoon the tos ecs bit is used for latency packets redirect in case of hardware mode and in case of software mode too ( but can be solved by different method of redirect)

Adding a mode to enable this feature in software mode and redirect any none tcp-udp will solve this

jsmoon:
@hhaim I did not find any relationship between tos ECN bits and ASTF behavior. Could you elaborate your concerns about it?
According to my investigation, the TOS_GO_TO_CPU is not used for ASTF mode. Since all the packets will go to CPU in ASTF mode and my change affects the ASTF mode only, I think it will have no effect.
If my thought is wrong, please let me know the points.

hhaim:
@jsmoon 

see this
```
const uint8_t icmp_pkt[]={
    0x00,0x04,0x96,0x08,0xe0,0x40,
    0x00,0x0e,0x2e,0x24,0x37,0x5f,
    0x08,0x00,

    0x45,0x03,0x00,0x30,
    0x00,0x00,0x40,0x00,
    0xff,0x01,0xbd,0x04,
    0x9b,0xe6,0x18,0x9b, //SIP
    0xcb,0xff,0xfc,0xc2, //DIP


```
the TOS is 0x3, the rules of hardware to forward latency packets is here :

this is the code for hardware filter configuration 
```
    int configure_rx_filter_rules() {
        if (m_drop_all){
            set_mode(mfDROP_ALL_PASS_TOS);
        }else{
            set_mode(mfPASS_TOS);
        }
        return(0);
    }

```

jsmoon:
@hhaim, thank you for your explanation. Now I understand the case that ECN CE packets cannot be delivered to the TCP stack.
I can see ip.tos tunable mentioned the limitation. What do you think about making it tunable (disabled by default)?

hhaim:
@jsmoon you still can add this feature, but need more coed changes and doc changes 
1. add a knob for this feature
2. support only in software-mode and warning in case of an hardware mode 
3. do software redirect in a diff way if this feature is enabled

jsmoon:
@hhaim, I added tcp.do_ecn tunable and it can be enabled only when the hardware filter is not used.

jsmoon:
@hhaim I think there is something wrong with trex-bot. It was not triggered by my latest commit and I cannot access http://81.218.86.50:8080/job/trex_build/ also.
Could you check what's wrong with it?

hhaim:
@jsmoon sorry for the late response, been out of office. we will look into this

hhaim:
@jsmoon looks good. let's wait for our bot server.

hhaim:
@jsmoon regarding the server, there was a power cycle in the lab and we try to load it again

hhaim:
rebased. sorry for the late response

PR #1027, opened 2023-05-15
Avoid exception when all packets are empty

pt-maros:
I'm not sure if this change makes sense in other scenarios. I wanted to make traffic with a TCP 3-way handshake pcap file, so the packets without payload resulted in an exception.

----

If a pcap file contains only packets that have no payload the loop would not assign any packet to `combined_data`. That would result in a thrown exception at `new_dirs.append(combined_data.direction)` because string `combined_data` has no attribute `direction`.


hhaim:
@pt-maros can you check why the tests fails

hhaim:
recheck

hhaim:
@pt-maros you can re-trigger by "recheck" comment

Issue #861, opened 2022-05-27
Cannot Rx after DPDK 22.03 version with iavf(XL710 VF)


egwakim:
Hi,
With the latest TRex version.
Sometimes XL710 VF cannot Rx packets when startup.
It happens randomly. 
No issues found for several years before applying the latest TRex version.

1. Used TRex version : The latest git snapshot version of TRex ( master of 22.05.12 version )

2. Test environment
1) Used NIC : XL710 ( 40G NIC ) SRIOV-VF NIC
2) Used driver and F/W version
root:~# ethtool -i eth0
driver: i40e
version: 2.1.14-k
firmware-version: 7.10 0x800075e6 19.5.12
expansion-rom-version:
bus-info: 0000:3b:00.0
supports-statistics: yes
supports-test: yes
supports-eeprom-access: yes
supports-register-dump: yes
supports-priv-flags: yes
3) Reproduction
- Start TRex : Sometimes TRex start fails because TRex cannot get ARP from peer because Rx stopped
- Broadcasted packets are exist in same subnet
- It happens randomly : occurs with 10% ~ 80% probability

3. Logs
1) Kernel syslog : Fault found in IOMMU operation with iavf driver, 3b:02.4 stopped working(Could not get ARP)
May 23 03:55:32  xx[31168]: EAL: VFIO support initialized
May 23 03:55:32  xx[31168]: EAL: Using IOMMU type 1 (Type 1)
May 23 03:55:32  kernel: vfio-pci 0000:3b:02.3: enabling device (0000 -> 0002)
May 23 03:55:33  xx[31168]: EAL: Probe PCI driver: net_iavf (8086:154c) device: 0000:3b:02.3 (socket 0)
May 23 03:55:33  kernel: vfio-pci 0000:3b:02.4: enabling device (0000 -> 0002)
May 23 03:55:33  xx[31168]: EAL: Probe PCI driver: net_iavf (8086:154c) device: 0000:3b:02.4 (socket 0)
May 23 03:55:33  xx[31168]: TELEMETRY: No legacy callbacks, legacy socket not created
May 23 03:55:33  xx[31168]: iavf_configure_queues(): RXDID[22] is not supported, request default RXDID[1] in Queue[0]
May 23 03:55:34  kernel: i40e 0000:3b:00.0: Unprivileged VF 3 is attempting to configure promiscuous mode
[...]

PR #1030, opened 2023-05-17
can separate queue start from device start

jsmoon:
Hi, this change is to add the capability of deferring start queues.

Recently, my colleague experienced an issue similar to #861 and he resolved it by deferring queue start.
I think it can be another possible workaround for #861 also.
And I observed a similar issue with i40evf now, I need to apply it also.

I thought to make it default for i40evf and iavf drivers for the first time. But, since the behavior (queue started during device start) has been used, finally I thought making it optional could be better.
To my thought, after this PR has been merged, i40evf can be removed also.

@hhaim please check my changes and give me your opinion.

### DIFF
--- a/src/main_dpdk.cpp
+++ b/src/main_dpdk.cpp
@@ -225,6 +225,7 @@ enum {
        OPT_BNXT_SO,
        OPT_DISABLE_IEEE_1588,
        OPT_LATENCY_DIAG,
+       OPT_DEFER_START_QUEUES,
 
        /* no more pass this */
        OPT_MAX
@@ -324,6 +325,7 @@ static CSimpleOpt::SOption parser_options[] =
         { OPT_BNXT_SO,                "--bnxt-so",         SO_NONE},
         { OPT_DISABLE_IEEE_1588,      "--disable-ieee-1588", SO_NONE},
         { OPT_LATENCY_DIAG,           "--latency-diag", SO_NONE},
+        { OPT_DEFER_START_QUEUES,     "--defer-start-queues", SO_NONE},
 
         SO_END_OF_OPTIONS
     };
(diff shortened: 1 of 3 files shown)


hhaim:
@jsmoon I think that i40evf does not work for us (in our lab) so we are still using the old i40e driver. 
Adding @syaakov. what would happen to you without this change? 

BTW I don't see an issue to add this PR

syaakov:
@hhaim the i40evf is not supported anymore by dpdk, and iavf should be used instead.
But we found some issues with the new iavf so that's why we have added back the old i40evf (9e3b836).
@jsmoon In order to completely remove the i40evf we first need rolleback the changes from 9e3b836 and test if our regression passes with iavf driver and your patch.

jsmoon:
@syaakov do you want to include the rollback of 9e3b836 in this PR for your regression?
In my thought, it can be done separately after this feature is merged. What do you think about it?

hhaim:
@jsmoon what @syaakov is suggesting that the driver that you are using didn't work for us, 
maybe due to the issue you solve here. We need to check this and decide if to roolback. 
I don't see an issue with the PR

egwakim:
Hi @hhaim 
This issue is reproducible during iavf or i40evf driver initialization phase with broadcast packets are flooding condition.
Same patch applied on 2 different other DPDK applications and resolved issue.
Since DPDK already migrated to IAVF driver form i40evf, using iavf driver would be better approach in long-term.
It's urgent issue for us because it sometimes happening in our STP.
Can you review the code?

hhaim:
@egwakim The PR looks good and I agree with what you are saying. We need to revisit our usage of i40evf and maybe to remove it in the next DPDK upgrade.

hhaim:
rebased

PR #1033, opened 2023-05-25
build and cleanup per-thread database at CP

jsmoon:
Hi, this PR will do more processing at CP when `process_at_cp` is enabled.

There was missed processing per-thread RW database when the `process_at_cp` has been added.
According to my investigation, all AstfDB instances can be initialized without the contexts from the DP cores.
So, parsing, building, and cleanup of AstfDB instance can be done safely and it will reduce DP core CPU consumption when multiple profiles are used dynamically.


hhaim:
@jsmoon there is only one minor, the memory of the DP is allocated in CP which could be a remote memory (in the old version it was the DP memory which is a local memory) 
I'm not sure what would be the performance impact of that, will commit and check

hhaim:
rebased

PR #1043, opened 2023-06-20
fix build bug during process_at_cp

jsmoon:
Hi, this PR is to fix the build bug during process_at_cp.

My previous PR #1033 made a bug that calls get_db_ro with dual_port_id.
It causes a crash when NUMA 1 memory is used. It should be called with socket_id.
I'm sorry for my previous fault.

@hhaim, please take a look at this change and give your feedback as soon as possible.


hhaim:
rebased

PR #1047, opened 2023-06-27
fix SyntaxError: Lambda expression parameters cannot be parenthesized

digger-yu:
fix SyntaxError: Lambda expression parameters cannot be parenthesized
modified:   doc/ELKConnect.py

please check


hhaim:
rebased

PR #1048, opened 2023-07-04
Some debian bookworm compilation fixes

GabrielGanne:
Hi,

These are some small fixes for warnings that I encountered after upgrading to debian bookworm which uses gcc/g++ (Debian 12.2.0-14) 12.2.0.

They are mostly trivial and I do not expect they will cause trouble in any other env.
It's not all of them: I still see some dangling-pointer or maybe-uninitialized warnings that I will try to address when I can.
EDIT: removed https://github.com/GabrielGanne/trex-core/commit/ff6c225dae0f6e2a1353e78771f889f9cabb4f40 from the initial PR after CI failure: it seems there are some assertions I missed.

Best regards

### DIFF
--- a/src/bp_gtest.cpp
+++ b/src/bp_gtest.cpp
@@ -2223,7 +2223,7 @@ class CRxCheck1 : public CRxCheckCallbackBase {
 public:
 
     virtual void handle_packet(rte_mbuf_t  * m){
-        rte_pktmbuf_mtod(m, char*);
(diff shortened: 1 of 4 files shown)


hhaim:
@GabrielGanne why did you remove the stdlibc++?

GabrielGanne:
That's my bad, I shouldn't have. Sorry.

hhaim:
@GabrielGanne thanks for the first commit. 
rebased.

PR #1052, opened 2023-08-03
Enable_iavf_driver

egwakim:
Hi, This patch enables iavf driver. Known hang issue can be resolved with --defer-queue-start option.


egwakim:
Hi @hhaim 
Last year, new iavf driver reverted to old i40e driver due to https://github.com/cisco-system-traffic-generator/trex-core/issues/861 issue
But, it can be resolved with --defer-queue-start option with https://github.com/cisco-system-traffic-generator/trex-core/pull/1030 patch. There is no reason to keep old driver. It would be good for long-term maintenance because Intel already migrated to iavf driver.

hhaim:
@syaakov can you check that this works on our setup

syaakov:
@egwakim we ran our regression testing on the iavf driver, but some ASTF tests failed.
We're going to try this driver with dpdk version 23.07 to see if the problem has been fixed already.

hhaim:
@egwakim 
solved by @syaakov by upgrading to dpdk 23.03

PR #1054, opened 2023-08-22
trex_astf_client: use proper to_json_str call

bzalkilani:
The profile object returned from ASTFProfile.load
can be of type ASTFProfileLight if the passed
"profile" is a path to a json profile file.

ASTFProfileLight has a different to_json_str
signature, and calling the method assuming
the returned profile object is of type
ASTFProfile will lead to failures in this
case.


bzalkilani:
@hhaim can you please help me figure out why the CI fails so I can fix it?

hhaim:
@bzalkilani it seems as our issue.

ybrustin:
recheck

PR #1058, opened 2023-09-13
dpdk: Add support for compressed firmware

daosman:
DPDK has support for loading compressed firmware through the RTE_HAS_LIBARCHIVE config.  This commit enables that config if libarchive is found on the build host.


hhaim:
rebased

Issue #1061, opened 2023-10-06
All packets drops if broadcast received during TRex start-up


egwakim:
Hi @hhaim 
I commented this issue on https://github.com/cisco-system-traffic-generator/trex-core/issues/861
This issue is re-happening from DPDK 23.03.

Issue
All iavf ports hang if broadcast packets received during TRex initialization.

Reproduction : 100% reproducible if broadcast packets flooded

Run flood ping with broadcast in same subnet
# sudo ping -f -b -q 1.1.1.255 -I enp59s13f3 
Note : enp59s13f3 is one of IAVF which shares same PF with TRex ports
It's happening whatever the latest version of iavf kernel driver used or i40evf driver used.

Start TRex
Failed to get MAC, all packets are dropped and marked as ierror and rx_missed error
trex>stats -x
Xstats

     Name:              |     Port 0:     |     Port 1:     
-------------------------------+-----------------+----------------
rx_good_packets | 522 | 521
tx_good_packets | 11 | 11
rx_good_bytes | 84596 | 79896
tx_good_bytes | 660 | 660
rx_missed_errors | 373 | 326
rx_bytes | 86684 | 81980
rx_broadcast_packets | 895 | 847
rx_dropped_packets | 373 | 326
tx_bytes | 660 | 660
tx_broadcast_packets | 11 | 11

Found kernel driver error
kernel: DMAR: [DMA Write] Request device [3b:0d.1] fault addr 0 [fault reason 05] PTE Write access is not set

It happens when packets received during iavf_dev_rx_queue_start() call.

PR #1068, opened 2023-10-31
fix_iavf_queue_drop_issue

egwakim:
Fix IAVF queue drop issue(#1061) when broadcast packets received during TRex startup. To avoid IO access conflict between queue 0 Rx and startup of other queues, start queue 0 in last.

### DIFF
--- a/src/main_dpdk.cpp
+++ b/src/main_dpdk.cpp
@@ -6037,7 +6037,7 @@ COLD_FUNC void CPhyEthIF::start_queues(void) {
     struct rte_eth_dev_info dev_info;
     rte_eth_dev_info_get(m_repid, &dev_info);
 
-    for (int queue_id = 0; queue_id < dev_info.nb_rx_queues; queue_id++) {
+    for (int queue_id = dev_info.nb_rx_queues-1; queue_id >= 0; queue_id--) {
         struct rte_eth_rxq_info rxq_info;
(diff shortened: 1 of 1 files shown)


egwakim:
Hi @hhaim 
I applied the fix in '--defer-start' case only to minimize impact on existing legacy usage.
If same issue found in teh default case(no defer-start case), same change can be applied.
It seems that TRex verification job environment have some issue.
Job failed in non-related part.

hhaim:
@egwakim you are right, job failure is not related. 
Why making the queue zero last help in IAVF flood issue? Why this issue is specific to IAVF? I see #1061 but can't see how it helps.

egwakim:
If broadcast packets flooded during TRex startup, if TRex using multiple queues.
Since default queue(0) used by TRex and Rx function working after queue 0 started.
But, iavf driver does not designed to isolate IO access between Rx and queue start.
So if queue 0 Rx and queue 1 start operation overlaps, it'll cause IOMMU fault as attached in #1061. 
To avoid such situation for iavf, enabling queue 0 last need to be enabled last.
This issue can be reproduced easily by using flood ping to sibling SRIOVF interface during TRex startup with multiple queues enabled.

egwakim:
Hi @hhaim 
Is it possible to re-run failed verification job?

egwakim:
Hi @hhaim 
Is there any updates on verification job?

hhaim:
@egwakim rebased and fixed the job perf issue.

PR #1069, opened 2023-11-02
Fix compiler warnings with gcc 13.2.1

ususdei:
Hi!

gcc 13.2.1 complains about a few minor issues.
I addressed some of them here.

Still, I need to build with `-Wno-maybe-uninitialized` and a workaround for `-Wdangling-pointer`, but from a first glance, those seem to be false positives, so I did not include them in this PR.

### DIFF
--- a/external_libs/bpf/portability.h
+++ b/external_libs/bpf/portability.h
@@ -39,6 +39,7 @@
  * flavors of UN*X.
  */
 
+#include <string.h>
 #include "funcattrs.h"
 
 #ifdef __cplusplus
--- a/src/pkt_gen.h
+++ b/src/pkt_gen.h
@@ -22,6 +22,8 @@
 #ifndef __PKT_GEN_H__
 #define __PKT_GEN_H__
 
+#include <stdint.h>
+
 enum {
     D_PKT_TYPE_ICMP = 1,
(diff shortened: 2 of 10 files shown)


ususdei:
Hmmm, this probably looks like an issue in your CI:

```
+ python /jenkins/scripts/xmls2html.py --input-xml /tmp/jenkins/reports/build.xml --output-dir /tmp/jenkins/reports --start-time=1698933848 '--scenario=TRex DMZ build - github PR' --build-url=http://81.218.86.50:8080/job/trex_build/807/ --build-id=807 --repo-path=/tmp/jenkins/repo
Processing report: /tmp/jenkins/reports/build.xml
Removing common prefix (build) from classnames.
  File "/jenkins/scripts/xmls2html.py", line 311, in <module>
    html_output += pad_tag(add_tests(tests, add_output = True, add_show_toggle = True), 'div')
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1487: ordinal not in range(128)
+ STATUS=1
```

hhaim:
rebased

hhaim:
@ususdei I found the issue, it is a special chars in comments of the commit, I've reverted the code. 
I will try to fix it

PR #1074, opened 2023-11-10
[v2] Fix compiler warnings with gcc 13.2.1

ususdei:
I grepped for all invalid-ascii chars in the codebase and couln't find one included in my recent PR.
So I figured the failed CI might have been due to the umlaut in my surname.

I rebased the branch with an altered name and just want so see what happens now.


hhaim:
@ususdei now it is ok, thanks for fixing it

hhaim:
rebased

PR #1076, opened 2023-11-21
Use ie40evf instead of iavf driver

ehhkjjn:
Hi, this PR fixes IOMMU fault when the IAVF driver is used together with other DPDK applications. This is done by using the i40evf driver instead of IAVF if TREX_PATCH is set.

This is related to the following issue:
Cannot Rx after DPDK 22.03 version with iavf(XL710 VF) #861

### DIFF
--- a/src/dpdk/drivers/net/iavf/iavf_ethdev.c
+++ b/src/dpdk/drivers/net/iavf/iavf_ethdev.c
@@ -142,10 +142,12 @@ static int iavf_tm_ops_get(struct rte_eth_dev *dev __rte_unused, void *arg);
 
 static const struct rte_pci_id pci_id_iavf_map[] = {
 	{ RTE_PCI_DEVICE(IAVF_INTEL_VENDOR_ID, IAVF_DEV_ID_ADAPTIVE_VF) },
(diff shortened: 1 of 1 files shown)


egwakim:
The issue with standalone TRex startup can be resolved by by https://github.com/cisco-system-traffic-generator/trex-core/pull/1068
But, if different multiple DPDK applications access same PF, it still have issue with IAVF.
We need to use i40evf.
I reported iavf issue to DPDK bugzilla.
https://bugs.dpdk.org/show_bug.cgi?id=1019

egwakim:
Hi @hhaim 
i40evf is still maintained by DPDK and TRex.
See the latest version of DPDK : https://github.com/DPDK/dpdk/tree/main/drivers/net/i40e
See code in TRex : https://github.com/cisco-system-traffic-generator/trex-core/tree/master/src/dpdk/drivers/net/i40e
Above code tested in my lab and worked well with i40evf driver.

hhaim:
@egwakim understood, let me check on our setup

hhaim:
Works on our setup. 
rebased.

PR #1096, opened 2024-01-12
Fix sanitizer errors in testcases

bjosv:
- Fixed the heap-use-after-free issues in TPGRxStatsTest.
The tests uses `tpg_header` when validating the statistics while its memory `m1` has been freed.
`tmp_buf` is not used in tests but is now correct in size.

- Fixed direct and indirect leaks in tests
Fixed a couple of testcases that triggered errors by the leak sanitizer.

All issues are due to missing actions in the testcases.

There are still 2 testcase left that are expected to leak for some reason:
* gt_astf_inter.astf_negative_2
* gt_astf_inter.astf_negative_3

while negative_4 and negative_5 are commented away in legacy. Maybe we should remove all of them?
The flows are already covered by other testcases, except they perform proper cleanup.
    
Currently we need to filter out these negative testruns when using memory analysis tools:
 `./bp-sim-64 --ut --gtest_filter='-*.astf_negative*'`

### DIFF
--- a/src/bp_gtest.cpp
+++ b/src/bp_gtest.cpp
@@ -2884,6 +2884,7 @@ TEST_F(gt_ring, ring1) {
         uint32_t *p;
         assert(my.Dequeue(p)==0);
         EXPECT_EQ_UINT32(*p, i);
+        delete p;
     }
     uint32_t *p;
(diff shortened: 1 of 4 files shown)


hhaim:
recheck

PR #1102, opened 2024-02-08
dpdk: use 1gb hugepages in priority

rjarry:
Even when 1GB hugepages are allocated on boot and available, trex will try to allocate 2MB pages and ignore the 1GB pages. Change the logic and see if 1GB hugepages can be used in priority. If there are not enough 1GB hugepages, fallback on 2MB.

### DIFF
--- a/scripts/dpdk_setup_ports.py
+++ b/scripts/dpdk_setup_ports.py
@@ -729,7 +729,7 @@ Other network devices
             pass # should we fail here?
 
     def is_hugepage_file_exits(self,socket_id):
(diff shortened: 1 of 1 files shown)


aharivel:
I've tested this patch and the result is as discribed.
This prevent Warning like:

# ./t-rex-64 -i ./stl/udp_1pkt_mac.py
WARNING: tried to configure 2048 hugepages for socket 0, but result is: 1136
...

Tested-by: Anthony Harivel <aharivel@redhat.com>

hhaim:
rebased

PR #1103, opened 2024-02-08
Revert "Use ie40evf instead of iavf driver"

ehhkjjn:
This reverts commit 7825d05cfbf2b8da3eaa3c6d9714a98e46086f6b - "Use ie40evf instead of iavf driver".

The commit "Use ie40evf instead of iavf driver" did not solve any issue. Instead it introduced other problems. After discussion with Gwangmoon Kim we came to the conclusion that we need to revert the previous commit.


hhaim:
rebased

PR #1104, opened 2024-02-15
fix doc typo for bidirectional traffic

joetats:
As someone who was using the cookbook as a replacement for the documentation (whoops) this was a little gotcha that caught me, hoping this helps!


hhaim:
rebased

PR #1106, opened 2024-02-25
Fix packet counters for mlx5 driver in trex

Civil:
That should fix wrong stats on ConnectX6 network cards once and for all

That is a minimized version of the patch (that retains counters diff).

Detailed changes:
 * Instead of keeping enums for counters, query them and store mapping in unordered_map
 * Remove xstats_array as it is no longer needed if we have mapping
 * Reference counters by their DPDK name
 * To make it future proof - make sure if counters are missing trex will crash with out_of_bound exception (asserts are there to verify counters still exists)
 * stats are now uint64_t instead of uint32_t as that is how they are reported from DPDK and down-conversion seems unnecessary

Screenshot of the change before the changes (current v3.04):
v3.04

Screenshot of the state of counters after this change:
This PR


hhaim:
@Civil pass our regression. rebased, thanks for the digging into this

PR #1112, opened 2024-02-27
Fix console `stats -x` not working on mlx5 cards

### LINKED ISSUE #1111: trex-console doesn't report xstats when len(values) doesn't match len(names)
In DPDK 23.03 (and later), it is possible that Length of xstats_name doesn't match lenth of xstats_values:

https://doc.dpdk.org/api-23.03/rte__ethdev_8h.html#a300d75b583c1f5acfe5b162a5d8c0ac1

```
There is an assumption that 'xstat_names' and 'xstats' arrays are matched by array index: xstats_names[i].name => xstats[i].value

```

This is exactly what happens in my case:
```
trex>stats -x

stats - Length of get_xstats_names: 79 and get_port_xstats_values: 63
```

https://github.com/cisco-system-traffic-generator/trex-core/blob/master/scripts/automation/trex_control_plane/interactive/trex/common/stats/trex_port_stats.py#L226-L227

Simple workaround would be to do:
```
        names = self.names[:len(values)]
        if len(values) != len(names):
            raise TRexError('Length of get_xstats_names: %s and get_port_xstats_values: %s' % (len(self.names), len(values)))

```

And that would make `stats -x` working again even if you combine multiple generations of connectx cards:
CleanShot 2024-02-27 at 19 35 46

Civil:
At least in DPDK 23.03 (and later) xstats_names might return more names than xstats_values and that is considered normal.

See https://doc.dpdk.org/api-23.03/rte__ethdev_8h.html#a300d75b583c1f5acfe5b162a5d8c0ac1

In this case it seems appropriate to force names to be same length as values

Fixes #1111

### DIFF
--- a/scripts/automation/trex_control_plane/interactive/trex/common/stats/trex_port_stats.py
+++ b/scripts/automation/trex_control_plane/interactive/trex/common/stats/trex_port_stats.py
@@ -223,10 +223,11 @@ class PortXStats(AbstractStats):
     def _pre_update(self, snapshot):
         values = snapshot['xstats_values']
 
-        if len(values) != len(self.names):
-            raise TRexError('Length of get_xstats_names: %s and get_port_xstats_values: %s' % (len(self.names), len(values)))
+        # DPDK might report more available names than values
+        # See https://doc.dpdk.org/api-23.03/rte__ethdev_8h.html#a300d75b583c1f5acfe5b162a5d8c0ac1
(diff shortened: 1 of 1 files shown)


Civil:
@hhaim would it be possible for you to look at this PR? As to be honest, I don't want to maintain my own fork but without that patch TUI doesn't work properly on ConnectX-6 and 7.

Issue #1109, opened 2024-02-26
trex-console TUI and trex-server port stat should display more than 4 ports


Civil:
I understand that there is a consideration about smaller terminals. However, I don't see any reason why that shouldn't be configurable.

E.x. running on 4k monitor displaying stats from more than 4 ports is desirable and running multiple consoles is less convenient.

In my environment I did something like this:
```
--- v3.04/automation/trex_control_plane/interactive/trex/common/trex_client.py	2023-08-29 14:13:59.000000000 +0200
+++ v3.04-patched/automation/trex_control_plane/interactive/trex/common/trex_client.py	2024-02-26 23:48:34.872232247 +0100
@@ -3622,7 +3622,7 @@
             self.logger.warning(format_text('Empty set of ports\n', 'bold'))
             return

-        port_stats = [self.ports[port_id].get_port_stats() for port_id in ports[:4]]
+        port_stats = [self.ports[port_id].get_port_stats() for port_id in ports]

         # update in a batch
         StatsBatch.update(port_stats, self.conn.rpc)
@@ -3649,7 +3649,7 @@
             self.logger.warning(format_text('Empty set of ports\n', 'bold'))
             return

-        port_xstats = [self.ports[port_id].get_port_xstats() for port_id in ports[:4]]
+        port_xstats = [self.ports[port_id].get_port_xstats() for port_id in ports]

         # update in a batch
         StatsBatch.update(port_xstats, self.conn.rpc)
```

Here is how my UI looks with 8 ports:
![CleanShot 2024-02-27 at 00 06 56](https://github.com/cisco-system-traffic-generator/trex-core/assets/844380/c2e98318-c673-4897-b038-62acf8385447)

And on a big screen, it looks good even with 10 ports:
![CleanShot 2024-02-27 at 00 08 59](https://github.com/cisco-system-traffic-generator/trex-core/assets/844380/ad4f5b50-0373-4b81-9d91-7cee0277a44f)

Similarly on trex server side:
![CleanShot 2024-02-27 at 00 13 18](https://github.com/cisco-system-traffic-generator/trex-core/assets/844380/d47ddc58-533a-4647-89b1-d617d971860c)

I haven't fixed the table drawing code, just commented out artificial 4 port limit.

To make it cleaner it still compatible with old behavior that can be made a CLI flag for those who want to remove the limit.

PR #1113, opened 2024-02-27
Make amount of shown ports configurable

Civil:
Currently it is hardcoded to '4' which is suboptimal for bigger screens. Add a new flag to trex-console, `-m` that allows to specify amount of ports user want to see.

Keep default value as 4 for compatibility reasons.

Related to #1109


hhaim:
@Civil seems as a good idea, I will rebase it

Civil:
Any updates on that? And also on #1112 ?

PR #1114, opened 2024-02-29
docs: wget skip certificate check

CorruptedPixl:
The cisco website doesnt have a valid certificate so the wget commands need --no-check-certificate added.
Pure documentation change

### DIFF
--- a/doc/trex_book.asciidoc
+++ b/doc/trex_book.asciidoc
@@ -335,20 +335,20 @@ Latest (stable) release:
 ----
 [bash]>mkdir -p /opt/trex
 [bash]>cd /opt/trex
-[bash]>wget --no-cache https://trex-tgn.cisco.com/trex/release/latest
+[bash]>wget --no-cache --no-check-certificate https://trex-tgn.cisco.com/trex/release/latest
 [bash]>tar -xzvf latest
 ----
 
 Latest (bleeding edge) version:
 [source,bash]
 ----
-[bash]>wget --no-cache https://trex-tgn.cisco.com/trex/release/be_latest
+[bash]>wget --no-cache --no-check-certificate https://trex-tgn.cisco.com/trex/release/be_latest
 ----
 
 To obtain a specific version:
 [source,bash]
 ----
-[bash]>wget --no-cache https://trex-tgn.cisco.com/trex/release/vX.XX.tar.gz #<1>
+[bash]>wget --no-cache --no-check-certificate https://trex-tgn.cisco.com/trex/release/vX.XX.tar.gz #<1>
 ----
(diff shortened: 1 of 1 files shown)


hhaim:
@CorruptedPixl the certificate is ok, your linux machine is not up to date, you can check it using a browser

CorruptedPixl:
Thanks for the heads up. It seems the CA just can't be verified the newest Ubuntu LTS release yet. Works fine in a browser.
Is it worth adding a note in the docs until this is fixed?

Civil:
>  the certificate is ok, your linux machine is not up to date, you can check it using a browser

I'm on OSX latest (and I have linux machines) that all reports `NET::ERR_CERT_AUTHORITY_INVALID`

And log from `openssl s_client`: https://gist.github.com/Civil/1ae889e7b3729f51244e34de4528c312

Also, report from ssllabs, most important extracts:
```
CONNECTED(00000005)
depth=0 C=US, ST=California, L=San Jose, O=Cisco Systems Inc., CN=trex-tgn.cisco.com
verify error:num=20:unable to get local issuer certificate
verify return:1
depth=0 C=US, ST=California, L=San Jose, O=Cisco Systems Inc., CN=trex-tgn.cisco.com
verify error:num=21:unable to verify the first certificate
verify return:1
depth=0 C=US, ST=California, L=San Jose, O=Cisco Systems Inc., CN=trex-tgn.cisco.com
verify return:1
---
Certificate chain
 0 s:C=US, ST=California, L=San Jose, O=Cisco Systems Inc., CN=trex-tgn.cisco.com
   i:C=US, O=IdenTrust, OU=HydrantID Trusted Certificate Service, CN=HydrantID Server CA O1
   a:PKEY: rsaEncryption, 2048 (bit); sigalg: RSA-SHA256
   v:NotBefore: Dec  5 12:00:57 2023 GMT; NotAfter: Dec  4 11:59:57 2024 GMT
---
```

and:
```
SSL handshake has read 2362 bytes and written 452 bytes
Verification error: unable to verify the first certificate
```

Also, here is ssllabs report: https://www.ssllabs.com/ssltest/analyze.html?d=trex-tgn.cisco.com

It also mentions that the certificate chain is incomplete (alongside other potential issues, like allowing RC4 as a cipher).

Civil:
What worries me is what is reported in Server header:
```
Server: Apache/2.4.6 (CentOS) OpenSSL/1.0.2k-fip
```

If that is true and not masked, that also means that you should upgrade before CentOS 7 is phased out (June 30, 2024). Also a smaller problem that OpenSSL 1.0.2 doesn't support TLS 1.3 as far as I remember and RedHat never backported that part from 1.1.0+.

PR #1121, opened 2024-03-31
Update bundled DPDK to 24.03

Civil:
* Update bundled DPDK to 24.03

I've tested that it works on my mellanox cards, but I don't have neither intel nor broadcoms so I can't actually verify that it is working well with them.


Civil:
Also, for simplicity, I haven't fixed a small problem with how jsoncpp is used (also, it is extremely outdated and that might be a problem on its own, I'd strongly suggest to pull in latest jsoncpp and probably use a submodule instead...)

If you are interested in that small fix:
```
--- a/src/stx/stl/trex_stl_rpc_cmds.cpp
+++ b/src/stx/stl/trex_stl_rpc_cmds.cpp
@@ -1896,7 +1896,7 @@ TrexRpcCmdUpdateTaggedPktGroupTags::_run(const Json::Value &params, Json::Value

         std::string type = parse_string(itr, "type", result, "");
         uint16_t tag_id = parse_uint16(itr, "tag_id", result);
-        const Json::Value& value = parse_object(itr, "value", result, Json::nullValue);
+        static const Json::Value& value = parse_object(itr, "value", result, Json::nullValue);

         if (tag_id >= num_tags) {
             err_msg = "Tag Id " + std::to_string(tag_id) + " greater than num of tags " + std::to_string(num_tags);
```

Civil:
Honestly I don't know how to fix the current problem in your build system with "This machine doesn't support RDRAND" because I don't see any place where it would be forced at this moment.

PR #1141, opened 2024-08-12
Adding BFD to Filter modes

haklat43:
This pull request is about addition of the BFD protocol to TRex filter modes.


haklat43:
Think trex-bot failed on some undeclared Mellanox C variable, 
RDMA_NLDEV_ATTR_NDEV_INDEX

Cannot see how that could be caused by the PR code change.

haklat43:
Sorry think I messed up the original PR with some two other issue suggestions (1144 and 1145).

PR #1193, opened 2025-09-08
dpdk: fix ice tx

jeffreybshaw:
The tx packets were dropped when creating drop any rule for ingress direction only. Added a direction lookup flag to fix the issue. Added accounting for Rx drop packets in ipackets and ibytes.

The add_dir_lkup flag we set is considered in
ice_tun_type_match_word(), called by ice_add_special_words() When the flag is set, a "special" lookup is added to the field vector list to tell hardware to in fact lookup the direction.

The function that interprets/translates the lookups is ice_get_sw_fv_list(). When ice_get_sw_fv_list() is called *before* ice_add_special_words(), the "special" words don't exist yet, so they're not translated. The fix is to call ice_get_sw_fv_list() *after* ice_add_special_words().


jeffreybshaw:
These changes are made in v3.0.7

PR #1200, opened 2025-09-12
Initialize values in jsoncpp.

bluhm:
Due do compiling json with -Werrro, g++ (Debian 14.2.0-19) 14.2.0 fails with "warning: '*(Json::Value::ValueHolder*)this' is used uninitialized".  The default constructor uses nullValue to initialize the Value without a value.  When calling method swapPayload() with such an object, the uninitialized internal value gets accessed.  As a simple workaround set new Value objects to integer 0.

### DIFF
--- a/external_libs/json/jsoncpp.cpp
+++ b/external_libs/json/jsoncpp.cpp
@@ -425,7 +425,7 @@ bool Reader::readValue() {
     break;
   case tokenNull:
     {
-    Value v;
+    Value v(0);
     currentValue().swapPayload(v);
     currentValue().setOffsetStart(token.start_ - begin_);
     currentValue().setOffsetLimit(token.end_ - begin_);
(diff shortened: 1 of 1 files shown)


external_libs/json/jsoncpp.cpp
```diff
@@ -425,7 +425,7 @@ bool Reader::readValue() {
     break;
   case tokenNull:
     {
-    Value v;
+    Value v(0);
```
emmanuel-deloget, line 428: Since Value exists with a default ctor (one that takes nullValue by default) this is not, IMHO, the right change. The default ctor shall be changed so that it correctly initialize all the Value properties when called (this is not done if it's called without any parameter or if it's called with vtype = nullValue).

The same goes for the other occurences of the same change.

PR #1209, opened 2026-01-28
mlx5 : correct ibv_import_device autoconf check from 'type' to 'func'

ErezF:
The autoconf check for ibv_import_device was incorrectly using 'type' instead of 'func', causing the compilation test to fails and as a result fails to probe mlx5 devices :

Starting  TRex v3.08 please wait  ...
mlx5_common: Failed to import IB device "mlx5_1": Operation not supported mlx5_net: probe of PCI device 0000:81:00.0 aborted after encountering an error: Operation not supported mlx5_common: Failed to load driver mlx5_eth

The correct parameter should be 'func', which creates a function pointer test that succeeds when ibv_import_device is available.

This issue was introduced in commit d24324189 (Dec 2024) when the autoconf checks were added during DPDK 25.07 integration.


vincentbernat:
This patch is really useful and works for me. Is it possible to merge it?

PR #1212, opened 2026-02-24
ndr: fix rate measurement accuracy and add ramp-up support

rjarry:
The NDR benchmark can produce inaccurate results due to two independent measurement issues.

First, traffic jumps instantly from zero to the target rate at each iteration. This sudden burst can overwhelm the DUT with flow table misses, scheduling lag or buffer overflows that would not occur under steady-state conditions. The benchmark interprets these transient drops as real packet loss and lowers the NDR estimate. An optional exponential ramp-up is added (disabled by default) that starts at 0.1% of the target rate and geometrically increases to 100%. Users who experience artificially low NDR results due to burst-induced drops can enable it with --ramp-up-time. The ramp uses the existing update_traffic RPC path so no C++ changes are needed.

Second, get_stats() was called while traffic was still running. Per-port counters are not sampled atomically, so TX and RX stats from different ports are read at slightly different times. This causes discrepancies in opackets/ipackets totals, producing spurious phantom loss or negative drop rates that confuse the binary search. Traffic is now stopped before reading stats so that all counters reflect final settled values.


aharivel:
This patch resolves inconsistent NDR test results. Much needed commit.

Tested-by: Anthony Harivel <aharivel@redhat.com>

